Issue #115, opened 2023-03-05
Gateway and HTTPRoute unable to be deleted -- stuck finalizing


ellistarn:
### Reproduction steps:
1. Create gateway and gateway class
2. Delete gateway class
3. Delete gateway (stuck deleting)
4. Delete HTTPRoute (stuck deleting)
5. Observe finalizer still on resource.

This log line looks like the culprit.
```
gateway-api-controller-6d499b898c-nq4sr manager 2023-03-05T01:56:33.540Z	INFO	controller.gateway	Ignore it since not link to any gatewayclass	{"reconciler group": "gateway.networking.k8s.io", "reconciler kind": "Gateway", "name": "carverspring-2-v1bayfdckt", "namespace": "default"}
```

```
- apiVersion: gateway.networking.k8s.io/v1alpha2
  kind: HTTPRoute
  metadata:
    creationTimestamp: "2023-03-03T23:47:55Z"
    deletionGracePeriodSeconds: 0
    deletionTimestamp: "2023-03-04T02:28:55Z"
    finalizers:
    - httproute.k8s.aws/resources
```

Hu1buerger:
Is a workaround known to remove the GatewayClass?

Hu1buerger:
Yes i did. when deleting the GatewayClass kubectl just locks

Hu1buerger:
it survies a reset and reboot.

zijun726911:
This project `aws-application-networking-k8s` (`aws-gateway-api-controller`) only manages GatewayGlass with` controllerName: application-networking.k8s.aws/gateway-api-controller`. 

https://github.com/aws/aws-application-networking-k8s/blob/8c2ad0f102526e41ab3975799732652b69963cde/pkg/controllers/gatewayclass_controller.go#L68

But your GatewayGlass has `Controller Name:  gateway.envoyproxy.io/gatewayclass-controller`, which is not managed by the `aws-gateway-api-controller` . You could search answer and ask in the envoy gateway repo : 
https://github.com/envoyproxy/gateway 
https://gateway.envoyproxy.io/


But from my limited knowledge in k8s, you probably could try to do `kubectl edit gatewayclass eg` and delete these lines and save it:
```
Finalizers:
    gateway-exists-finalizer.gateway.networking.k8s.io
``` 
[...]

PR #116, opened 2023-03-05
handle deletes of gateway, httproute if gatewayclass is deleted

aaroniscode:
*Issue #, if available:* #115

Tested and this resolves the reproduce steps in the issue above.

*Description of changes:*
This is much simpler change than it appears, due to removing a nested if/else

This changes the logic of 2 controllers: `Gateway` and `HTTPRoute` very simply.
* During a reconcile, the resource is considered in scope if it has the VPC Lattice finalizer on it

Before, the reconcile logic would skip any `Gateway` or `HTTPRoute` if there wasn't a matching `GatewayClass`. While this makes sense for all creates, it doesn't make sense for a delete when we _know_ it's relevant because the controller put a finalizer on it there is a VPC Lattice resources created.

The change in `controllers/httproute_controller.go` is very straightforward.

The logic change is the same for `controllers/gateway_controller.go` but there was nested if/else clauses

The `reconcile` function had a nested if/else:

```
if gwClass.Spec.ControllerName == config.LatticeGatewayControllerName {

	if !gw.DeletionTimestamp.IsZero() {
```

The `gwClass.Spec.ControllerName == config.LatticeGatewayControllerName` logic was moved to a new function `isGatewayRelevant` that makes it clear the conditions that decide whether to reconcile or ignore a Gateway.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

### DIFF
--- a/controllers/gateway_controller.go
+++ b/controllers/gateway_controller.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/source"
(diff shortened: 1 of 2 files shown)


liwenwu-amazon:
I am closing this for now.  We can re-open for discussion in future if needed.

Issue #86, opened 2023-02-10
Need to support named `targetPort`


liwenwu-amazon:
The lattice controller does not allow you to use a named (i.e. string) `targetPort` in your kubernetes `service` resource, and have to use pod port number (i.e. int) directly - we use named ports a lot as they give a level of abstraction, so would like to see this supported:

works:

ports:
- name: http
  protocol: TCP
  port: 80
  targetPort: 80

does not work:

ports:
- name: http
  protocol: TCP
  port: 80
  targetPort: http

PR #243, opened 2023-05-12
Fix TG creation when using named targetPorts

solmonk:
**What type of PR is this?**

bugfix

Addresses #86. Currently named targetPorts cause error because it triggers CreateTargetGroup request with port 0. With this PR it will trigger request with port 80 instead.

1. TargetPort actually does not matter at all. To understand this we need to understand how ports are working in Lattice TGs. Ports defined in Lattice TGs only acts as a "default port" when no port is specified in RegisterTargets API. Since we always populate port when creating targets, the port on TG does nothing. We do not need an accurate number on this field.

2. But can't we still query the endpoint and find the port name? This path seems complicated. For example, there could be multiple pods using the same port name but different port numbers. In this case picking any of the port number will be technically wrong. This way also requires an (practically unnecessary) update on targetGroup when just targets are changing.

3. Setting to port 80 matches [Lattice documentation](https://docs.aws.amazon.com/vpc-lattice/latest/APIReference/API_TargetGroupConfig.html#vpclattice-Type-TargetGroupConfig-port) as well.

Overall I would treat this as a bugfix and deploy the simplest mitigation, unless we need this field in some way later.

* `make e2etest`
* Gateway conformance tests
* manual testing to confirm service targetPort (either named or unnamed) does not matter on target registration.

**Automation added to e2e**:

### DIFF
--- a/pkg/gateway/model_build_targetgroup.go
+++ b/pkg/gateway/model_build_targetgroup.go
@@ -187,7 +187,8 @@ func (t *targetGroupModelBuildTask) BuildTargetGroup(ctx context.Context) error 
 		Type: latticemodel.TargetGroupTypeIP,
 		Config: latticemodel.TargetGroupConfig{
 			VpcID: config.VpcID,
-			//Port:            backendServicePort,
+			// Fill in default HTTP port as we are using target port anyway.
+			Port:                80,
 			IsServiceImport:     false,
 			IsServiceExport:     true,
 			K8SServiceName:      t.serviceExport.Name,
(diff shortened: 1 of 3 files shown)


pkg/gateway/model_build_targetgroup.go
```diff
@@ -196,10 +197,18 @@ func (t *targetGroupModelBuildTask) BuildTargetGroup(ctx context.Context) error
...
-		svc.Spec.Ports[0].TargetPort.IntVal)
-
-	tgSpec.Config.Port = svc.Spec.Ports[0].TargetPort.IntVal
+	var backendServicePort int32
+	if port := svc.Spec.Ports[0].TargetPort.IntVal; port != 0 {
+		backendServicePort = port
+		glog.V(6).Infof("Found backend service port %v, intval %v\n", svc.Spec.Ports[0].TargetPort,
+			svc.Spec.Ports[0].TargetPort.IntVal)
+	} else {
+		// Fallback to default HTTP port. This can be any number, because we always use the port defined in targets.
```
liwenwu-amazon, line 206: if port is irrelevant for target group lattice config, can we just always use the "else" portion of logic?

Issue #251, opened 2023-05-16
bug: correctly infer namespace of HTTPRoute backendRef


solmonk:
### To reproduce
Create gateway in `default`, service in `non-default` namespace.

Create the following httproute in `non-default` namespace:
```yaml
apiVersion: gateway.networking.k8s.io/v1beta1
kind: HTTPRoute
metadata:
  name: inventory
spec:
  parentRefs:
  - name: my-hotel
    namespace: default
    sectionName: http
  rules:
  - backendRefs:
    - name: inventory-ver1
      kind: Service
      port: 8090
      weight: 10
```

### Current behavior
This leads to the following error when registering targets:
```
I0515 19:08:07.547245   87545 model_build_targets.go:107] errmsg: Build Targets failed because K8S service default/inventory-ver1 does not exist
I0515 19:08:07.547252   87545 model_build_targetgroup.go:162] Error buildTargets for backend ref service {{{0xc000b5c3b0 0xc000b5c3c0 inventory-ver1 <nil> 0xc000bbf240} 0xc000bbf244} []}
```
Seems the controller infers `default` namespace by default. [example1](https://github.com/aws/aws-application-networking-k8s/blob/main/pkg/gateway/model_build_targetgroup.go#L143) [example2](https://github.com/aws/aws-application-networking-k8s/blob/main/pkg/gateway/model_build_targetgroup.go#L281)

### What to expect
Since the route itself is in `non-default` namespace, the backendRefs [should also implicitly refer to](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.BackendObjectReference) `non-default`.

### Related
Possibly related to #244

PR #256, opened 2023-05-19
Remove assumption of default namespaces in httproutes

### LINKED ISSUE #244: bug: Target not automatically registered in Target Group
I am seeing errors like this in gateway api controller logs and the target is not automatically registered in target group

```
E0511 21:55:34.435754       1 request.go:539] Failed request: VpcLattice/ListTargets, Payload: {  TargetGroupIdentifier: ""}, Error: InvalidParameter: 1 validation error(s) found.
93
- minimum field size of 17, ListTargetsInput.TargetGroupIdentifier.
```

Please advise, will provide more info upon request.

P/S: I've already created a support ticket in AWS (Case ID 12717202271), posting this here for visibility.

solmonk:
**What type of PR is this?**
bug

Fixes #244 #251 

Will remove reference of hardcoded "default" and use its own namespace whenever it needs inference

* `make e2etest`
* pending gw conformance test

**Automation added to e2e**:
Will add e2etest in a separate PR.

```release-note

```

### DIFF
--- a/controllers/gateway_controller.go
+++ b/controllers/gateway_controller.go
@@ -24,7 +24,6 @@ import (
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
 
-	"github.com/aws/aws-application-networking-k8s/pkg/latticestore"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -35,6 +34,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 	gateway_api "sigs.k8s.io/gateway-api/apis/v1beta1"
+
+	"github.com/aws/aws-application-networking-k8s/pkg/latticestore"
 
 	"github.com/aws/aws-application-networking-k8s/controllers/eventhandlers"
 	"github.com/aws/aws-application-networking-k8s/pkg/aws"
@@ -149,7 +150,7 @@ func (r *GatewayReconciler) reconcile(ctx context.Context, req ctrl.Request) err
 					continue
 				}
 				gwName := types.NamespacedName{
-					Namespace: httpRoute.Namespace,
+					Namespace: gw.Namespace,
 					Name:      string(httpRoute.Spec.ParentRefs[0].Name),
 				}
 
@@ -331,7 +332,7 @@ func (r *GatewayReconciler) updateBadStatus(ctx context.Context, message string,
 func UpdateHTTPRouteListenerStatus(ctx context.Context, k8sclient client.Client, httproute *gateway_api.HTTPRoute) error {
 	gw := &gateway_api.Gateway{}
 
-	gwNamespace := "default"
+	gwNamespace := httproute.Namespace
 	if httproute.Spec.ParentRefs[0].Namespace != nil {
 		gwNamespace = string(*httproute.Spec.ParentRefs[0].Namespace)
 	}
--- a/controllers/gateway_controller.go
+++ b/controllers/gateway_controller.go
@@ -149,8 +149,12 @@ func (r *GatewayReconciler) reconcile(ctx context.Context, req ctrl.Request) err
 				if len(httpRoute.Spec.ParentRefs) <= 0 {
 					continue
 				}
+				gwNamespace := httpRoute.Namespace
+				if httpRoute.Spec.ParentRefs[0].Namespace != nil {
+					gwNamespace = string(*httpRoute.Spec.ParentRefs[0].Namespace)
+				}
 				gwName := types.NamespacedName{
-					Namespace: gw.Namespace,
+					Namespace: gwNamespace,
 					Name:      string(httpRoute.Spec.ParentRefs[0].Name),
 				}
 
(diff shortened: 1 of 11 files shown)


test/pkg/test/framework.go
```diff
@@ -86,7 +89,7 @@ func NewFramework(ctx context.Context) *Framework {
 	SetDefaultEventuallyTimeout(180 * time.Second)
 	SetDefaultEventuallyPollingInterval(10 * time.Second)
 	BeforeEach(func() { framework.ExpectToBeClean(ctx) })
-	AfterEach(func() { framework.ExpectToBeClean(ctx) })
+	AfterEach(func() { framework.CleanTestEnvironment(ctx); framework.ExpectToBeClean(ctx) })
```
liwenwu-amazon, line 92: Please revert this change.  We purposely do it as `pause on failure` and keep all lattice resource(s), so that we can debug in case of failure!

solmonk: This was included totally by mistake, I will remove it.

solmonk:
Used `inventory-ver1.yaml` (in non-default), `my-hotel-gateway.yaml` (in default) and the following httproute (in non-default):
```
kind: HTTPRoute
metadata:
  name: inventory
spec:
  parentRefs:
  - name: my-hotel
    namespace: default
    sectionName: http
  rules:
  - backendRefs:
    - name: inventory-ver1
      kind: Service
      port: 8090
      weight: 10
```
Confirmed that target groups are created with correct name including non-default, and targets are registered.

pkg/deploy/lattice/listener_synthesizer.go
```diff
@@ -65,9 +66,8 @@ func (l *listenerSynthesizer) Synthesize(ctx context.Context) error {
 
 		glog.V(6).Infof("ListenerSynthesize >>>> delete staled sdk listener %v\n", *sdkListener)
 		l.listener.Delete(ctx, sdkListener.ListenerID, sdkListener.ServiceID)
-		k8sname, k8snamespace := latticeName2k8s(sdkListener.Name)
 
-		l.latticestore.DelListener(k8sname, k8snamespace, sdkListener.Port, sdkListener.Protocol)
+		l.latticestore.DelListener(sdkListener.Name, sdkListener.Namespace, sdkListener.Port, sdkListener.Protocol)
```
liwenwu-amazon, line 70: I do not think sdkListener.Namespace is populated by l.getSDKListeners().  Not sure why unit test did not catch this? Also, i think the e2e should catch this too. (not sure why )

liwenwu-amazon: can u confirm this from the debugging output at line#58

liwenwu-amazon: Also,  for all these xxxSDKxxx() routine(s), I think we should use the name or tags return from lattice to figure out the namespace.

solmonk: I made the change to that function just right below actually, I think it is working?

On listener creation:
```
I0519 15:20:51.334060   39938 listener_synthesizer.go:126] getSDKListeners result >> [0xc00090cf50]
I0519 15:20:51.334069   39938 listener_synthesizer.go:58] getSDKlistener: [0xc00090cf50], err: <nil>
I0519 15:20:51.334074   39938 listener_synthesizer.go:86] findMatchListener: found matching moduleListener &{{0xc000a59280 AWS::VPCServiceNetwork::Listener inventory-non-default-80-HTTP} {inventory non-default 80 HTTP {false inventory-ver1 non-default}} <nil>}
```

liwenwu-amazon:
Also, please update or add unit test(s).

pkg/deploy/lattice/listener_synthesizer.go
liwenwu-amazon: The objective of these xxxSDKxxx() is to figure out if Lattice resource(s) can map back to K8S resource.  It should `not` imply nor inherent any value from K8S object.

solmonk: It is impossible to figure out exact namespace because of the naming scheme of lattice and k8s. Lets say we have lattice listener `a-b-c-d-e-80-HTTP`, the resource can be either one of:
`a/b-c-d-e`
`a-b/c-d-e`
`a-b-c/d-e`
`a-b-c-d/e`
I think we need to rely on k8s information to figure out namespace. it should be the same as the lattice service anyways.

liwenwu-amazon: then we need to explicit using `tag`.  The whole idea is to compare between lattice resources and K8S object(s).  And perform cleanup lattice resources if they are not matching K8S objects.  

If we start infer lattice resource(s) from K8S object(s), we will end up leaking more lattice resource(s)

controllers/gateway_controller.go
```diff
@@ -149,7 +150,7 @@ func (r *GatewayReconciler) reconcile(ctx context.Context, req ctrl.Request) err
 					continue
 				}
 				gwName := types.NamespacedName{
-					Namespace: httpRoute.Namespace,
+					Namespace: gw.Namespace,
```
liwenwu-amazon, line 153: this needs to be `httpRoute.Spec.ParentRefs[0].Namespace`  if exist, or default to httpRoute.Namespace

liwenwu-amazon: We need a e2e to cover this

liwenwu-amazon (approved):

Issue #261, opened 2023-05-26
Config discovery doesnt work for Fargate


zmingxi:
Controller currently checks if its running in a cluster. If it is, then it assumes EC2, while IMDS is not available to Fargate pods.

PR #264, opened 2023-06-01
Enable passing config into controller

zmingxi:
**What type of PR is this?**
bug

https://github.com/aws/aws-application-networking-k8s/issues/261

This PR enables controller to take input from user

`make e2etest`
`unit test`
`helm install --set awsRegion=us-west-2 --set awsAccountId=123456789 --set clusterVpcId=vpc-12345678 --dry-run --debug helm --generate-name`
`conformance test`

**Automation added to e2e**:
N/A

No

No
No

```release-note

```


pkg/config/controller_config.go
```diff
@@ -43,21 +43,66 @@ func GetClusterLocalGateway() (string, error) {
 	return DefaultServiceNetwork, nil
 }
 
-func ConfigInit() {
-	// discover VPC using environment first
-	VpcID = os.Getenv("CLUSTER_VPC_ID")
```
liwenwu-amazon, line 48: Is there any reason to change this?  Today we documented [how to use these environments](https://github.com/aws/aws-application-networking-k8s/blob/main/docs/enviroment.md)

zmingxi: This document is for When running AWS Gateway API Controller `outside` the Kubernetes Cluster, this specify the VPC of the cluster. This code changes is to take the input while running AWS Gateway API Controller  `inside` the he Kubernetes Cluster.

pkg/config/controller_config.go
```diff
@@ -43,21 +43,66 @@ func GetClusterLocalGateway() (string, error) {
...
+			VpcID, err = metadata.VpcID()
+			glog.V(2).Infoln("CLUSTER_VPC_ID from IMDS config discovery :", VpcID)
+			if err != nil {
+				glog.V(2).Infoln("IMDS config discovery is NOT AVAILABLE :", err)
+				return
+			}
+		} else {
+			VpcID = os.Getenv("CLUSTER_VPC_ID")
+			glog.V(2).Infoln("CLUSTER_VPC_ID from local dev environment: ", VpcID)
+		}
```
solmonk, line 70: From what I have seen, the most common practice on config is to have priority order `flag > env > auto-resolve`. I think env variable should be a way of configuring things regardless of dev environment. This way we can simply set env var as a default value of a flag, and run discovery if the current value is empty.

Actually... we have a lot of configuration variables (longname mode for example) and I think they all should be configurable by both flag and env. It makes things bigger so we can leave it to other task though.

solmonk: Since we have two ways of configuring the same value I think the naming should be consistent

e.g. flag `aws-vpc-id` -> env `AWS_VPC_ID`

helm/templates/deployment.yaml
```diff
@@ -77,3 +77,12 @@ spec:
       {{ if .Values.deployment.priorityClassName -}}
       priorityClassName: {{ .Values.deployment.priorityClassName }}
       {{ end -}}
+      {{- if .aws.region }}
```
solmonk, line 80: Should these be somewhere inside like `args` block?

pkg/config/controller_config.go
zmingxi: Yes, I discussed with Liwen, we plan to go with env first, i am going to change this using env var.

docs/developer.md
```diff
@@ -6,17 +6,16 @@ make help
...
 
 # Run this before submitting code
 make presubmit
 
-# Install CRDs (which only need once)
+# Install CRDs (which only need once) 
 kubectl apply -f config/crds/bases/k8s-gateway-v0.6.1.yaml
 kubectl apply -f config/crds/bases/multicluster.x-k8s.io_serviceexports.yaml
 kubectl apply -f config/crds/bases/multicluster.x-k8s.io_serviceimports.yaml
-kubectl apply -f examples/gatewayclass.yaml
```
liwenwu-amazon, line 19: why do u delete this?

docs/developer.md
```diff
@@ -35,20 +34,11 @@ And use "EnvFile" GoLand plugin to read the env variables from the generated `.e
...
-namespace if it doesn't exist.
-
-NOTE: You'll need to allow in-bound traffics from lattice prefix list in the security
-groups of your cluster.
-
-```bash
-# create non-default namespace if it hasn't existed yet
-kubectl create namespace non-default
-
+You should set up the correct `REGION` env variable
```
liwenwu-amazon, line 37: why do u delete all these?

docs/developer.md
```diff
@@ -91,4 +81,4 @@ make docker-build
 ```
 make build-deploy
 ```
-Then follow [Deploying the AWS Gateway API Controller](https://github.com/aws/aws-application-networking-k8s/blob/main/docs/deploy.md) to configure and deploy the docker image
+Then follow [Deploying the AWS Gateway API Controller](https://github.com/aws/aws-application-networking-k8s/blob/main/docs/deploy.md) to configure and deploy the docker image
```
liwenwu-amazon, line 84: Please remove all these white-space diff

liwenwu-amazon:
Please also add or modify unit test for the change

docs/deploy.md
```diff
@@ -79,7 +79,12 @@ Run through them again for a second cluster to use with the extended example sho
    helm install gateway-api-controller \
       oci://public.ecr.aws/aws-application-networking-k8s/aws-gateway-controller-chart\
       --version=v0.0.12 \
-      --set=aws.region=$AWS_REGION --set=serviceAccount.create=false --namespace aws-application-networking-system
+      --set=serviceAccount.create=false --namespace aws-application-networking-system \
+      # Region, clusterVpcId, awsAccountId are required for fargate use case
```
liwenwu-amazon, line 83: can you replace "fargate use case" to  something " for the case where IMDS are not available for discovering Region, clusterVpcId, awsAccountId,  for example  fargate usecase"

pkg/config/controller_config.go
```diff
@@ -17,9 +17,9 @@ const (
 )
 
 // TODO endpoint, region
-var VpcID = "vpc-xxxx"
-var AccountID = "yyyyyy"
-var Region = "us-west-2"
+var VpcID = ""
```
liwenwu-amazon, line 20: NIT,can use use a constant for ""? e.g. UNKNOWN_VPCID=""

pkg/config/controller_config.go
```diff
@@ -44,20 +44,49 @@ func GetClusterLocalGateway() (string, error) {
...
-	AccountID = os.Getenv("AWS_ACCOUNT_ID")
-	if AccountID == "" {
-		AccountID = os.Getenv("AWS_ACCOUNT") // Fallback to AWS_ACCOUNT for compatibility
+	sess, _ := session.NewSession()
+	metadata := NewEC2Metadata(sess)
+	var err error
+
+	// VpcId
+	VpcID = os.Getenv("CLUSTER_VPC_ID")
+	if VpcID != "" {
```
liwenwu-amazon, line 54: NIT, use constant instead of ""

pkg/config/controller_config.go
```diff
@@ -17,9 +17,9 @@ const (
 )
 
 // TODO endpoint, region
-var VpcID = "vpc-xxxx"
```
liwenwu-amazon, line 20: can u also remove //TODO comment? thanks

liwenwu-amazon (approved):

Issue #241, opened 2023-05-11
Targets are built for all ports of service instead of just port specified in backendRefs


rmullinnix461332:
We are setting up hashicorp vault as a service in lattice network.  The HTTPRoute document has a backendRefs with port 8200.  Targets are created for all ports of the service - 8200, 8201, 9091, and 9102.  Accessing the service through the lattice results in failures except for when routed to port 8200.  Expected behavior is that targets are only created for the port specified in the backend ref.

For a workaround, we created another service that just exposes port 8200 and routes to port 8200 on the vault service.

### Issue
Service
vault                   ClusterIP   XXX.XX.XX.XXX   <none>        8200/TCP,8201/TCP,9091/TCP,9102/TCP

      rules:
      - backendRefs:
        - name: vault
          kind: Service
          namespace: secrets-system
          port: 8200

Results in registered targets for each cluster member for each port
Registered Targets
  XXX.XX.XX.XXX.    8200    
  XXX.XX.XX.XXX.    8201    
  XXX.XX.XX.XXX.    9091    
  XXX.XX.XX.XXX.    9102    

Expected Result
Registered Targets
  XXX.XX.XX.XXX.    8200 

### Workaround
vault-headless          ClusterIP   XXX.XX.XX.XX    <none>        8200/TCP

solmonk:
Thanks for bringing this to attention. We currently do create targets for every address/port combination of the service endpoint, which will result in splitting traffic across all ports in a multi-port service. We will discuss the best path forward on this issue, and keep you posted here.

liwenwu-amazon:
@rmullinnix461332 is it possible you can show the 
```
kubectl get svc value -n secrets-system -o yaml
```

rmullinnix461332:
`apiVersion: v1
kind: Service
metadata:
  annotations:
    banzaicloud.com/last-applied: XXX
  creationTimestamp: "2023-05-10T19:05:25Z"
  finalizers:
  - service.ki8s.aws/resources
  labels:
    app.kubernetes.io/name: vault
    global_service: "true"
    vault_cr: vault
  name: vault
  namespace: secrets-system
  ownerReferences:
  - apiVersion: vault.banzaicloud.com/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: Vault
    name: vault
    uid: 3d3a6ef6-4394-428d-a83d-75320104043e
  resourceVersion: "95123970"
[...]

PR #286, opened 2023-07-04
Aws#241Added support for HTTPRoute BackendRef ports

sesmaili:
**What type of PR is this?** bug

**Which issue does this PR fix**: #241 

**What does this PR do / Why do we need it**: It gives the user the ability to control what which posts are registered as targets in Lattice's Target Group.

``` GO
ok      github.com/aws/aws-application-networking-k8s/pkg/aws/services  0.020s  coverage: 2.9% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/config        0.008s  coverage: 77.4% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/deploy        0.019s  coverage: 20.6% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/deploy/lattice        0.043s  coverage: 86.1% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/gateway       0.054s  coverage: 78.5% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/latticestore  0.004s  coverage: 73.3% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/model/core    0.004s  coverage: 40.9% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/model/core/graph      0.003s  coverage: 17.2% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/utils/log     0.003s  coverage: 0.0% of statements [no tests to run]
```

**Automation added to e2e**: Yes

```release-note

```


solmonk:
Thank you for the pull request.

Here I am worried about a scenario when a service is referenced by two different backendRefs with different `backendRef.Port`. I believe only one target group is created per service right now (TG name is like `k8s-{service_name}-{namespace}`) - so this change might cause another conflict issue.

I think separate target groups should be created either per each backendRef or per each port to fully address the issue. Per port seems better for me personally.

sesmaili:
> Thank you for the pull request.
> 
> Here I am worried about a scenario when a service is referenced by two different backendRefs with different `backendRef.Port`. I believe only one target group is created per service right now (TG name is like `k8s-{service_name}-{namespace}`) - so this change might cause another conflict issue.
> 
> I think separate target groups should be created either per each backendRef or per each port to fully address the issue. Per port seems better for me personally.

Is this a scenario we support today? Multiple httproutes using the same backendref?

zijun726911:
> Thank you for the pull request.
> 
> Here I am worried about a scenario when a service is referenced by two different backendRefs with different `backendRef.Port`. I believe only one target group is created per service right now (TG name is like `k8s-{service_name}-{namespace}`) - so this change might cause another conflict issue.
> 
> I think separate target groups should be created either per each backendRef or per each port to fully address the issue. Per port seems better for me personally.

This PR only confine num of targets that intend to be regiestred and didn't change model_build_target_group behavior, does it can indeed introduce new target group creation conflict? For this scenario: "when a service is referenced by two different backendRefs with different `backendRef.Port`", (if replicas: 1) controller could register 2 targets(same ip different ports) in one same target group?

solmonk:
After some discussion we have decided that we can move forward with disabling such usage (multiple backendRefs) for now.

Can you please update all example configs under `examples/` so that they correctly refer to service port? It was okay until now because we have been ignoring port under backendRef, but I am seeing most of the values are incorrect which will start to break things.  `examples/inventory-route.yaml` for example:

```
kind: HTTPRoute
metadata:
  name: inventory
spec:
  parentRefs:
  - name: my-hotel
    sectionName: http
  rules:
  - backendRefs:
    - name: inventory-ver1
      kind: Service
      port: 8090 # wrong, should be service port instead of target port.
      weight: 10
```

pkg/gateway/model_build_targets.go
```diff
@@ -111,19 +112,35 @@ func (t *latticeTargetsModelBuildTask) buildLatticeTargets(ctx context.Context)
...
+			//portsAnnotations := strings.Split(serviceExport.ObjectMeta.Annotations["multicluster.x-k8s.io/Ports"], ",")
+			definedPorts[0], err = strconv.ParseInt(serviceExport.ObjectMeta.Annotations[portAnnotationsKey], 10, 64)
+			if err != nil {
+				glog.V(6).Infof("Failed to read Annotaions/Port:%v, err:%s\n ", serviceExport.ObjectMeta.Annotations[portAnnotationsKey], err)
+			}
+			glog.V(6).Infof("Build Targets - portAnnotations: %v \n", definedPorts)
 		}
-		glog.V(6).Infof("Build Targets - portAnnotations: %v \n", portAnnotations)
+	} else if tg.ByBackendRef {
+		if (t.httpRoute != nil) && (t.httpRoute.Spec.Rules != nil) && (0 < len(t.httpRoute.Spec.Rules)) {
```
solmonk, line 131: nit: I think the parentheses are unnecessary.

pkg/gateway/model_build_targets.go
```diff
@@ -111,19 +112,35 @@ func (t *latticeTargetsModelBuildTask) buildLatticeTargets(ctx context.Context)
...
+			definedPorts = []int64{}
+			for _, rule := range t.httpRoute.Spec.Rules {
+				for _, ref := range rule.BackendRefs {
+					if ref.Port != nil {
```
solmonk, line 135: I think we should find the right rule by matching name of backendRef, instead of picking the first one. Also a test case for this would be great

pkg/gateway/model_build_targets.go
```diff
@@ -111,19 +112,35 @@ func (t *latticeTargetsModelBuildTask) buildLatticeTargets(ctx context.Context)
...
-	portAnnotations := undefinedPort
-	serviceExport := &mcs_api.ServiceExport{}
-	err = t.Client.Get(ctx, namespacedName, serviceExport)
-	if err != nil {
-		glog.V(6).Infof("Failed to find Service export in the DS. Name:%v, Namespace:%v - err:%s\n ", t.tgName, t.tgNamespace, err)
-	} else {
-		// TODO: Change the code to support multiple comma separated ports instead of a single port
-		//portsAnnotations := strings.Split(serviceExport.ObjectMeta.Annotations["multicluster.x-k8s.io/Ports"], ",")
-		portAnnotations, err = strconv.ParseInt(serviceExport.ObjectMeta.Annotations[portAnnotationsKey], 10, 64)
+	definedPorts := []int64{undefinedPort}
```
solmonk, line 115: nit: `definedPorts := make([]int64, 0)` will work fine here

test/pkg/test/nginxapp.go
```diff
@@ -19,7 +19,7 @@ type ElasticSearchOptions struct {
 	MergeFromService    []*v1.Service
 }
 
-func (env *Framework) NewElasticeApp(options ElasticSearchOptions) (*appsv1.Deployment, *v1.Service) {
+func (env *Framework) NewNginxeApp(options ElasticSearchOptions) (*appsv1.Deployment, *v1.Service) {
```
solmonk, line 22: should be `NginxApp`?

pkg/gateway/model_build_targets.go
```diff
@@ -146,11 +163,11 @@ func (t *latticeTargetsModelBuildTask) buildLatticeTargets(ctx context.Context)
 						TargetIP: address.IP,
 						Port:     int64(port.Port),
 					}
-					if portAnnotations == undefinedPort || int64(target.Port) == portAnnotations {
+					if definedPorts[0] == undefinedPort || portdefined(definedPorts, int64(target.Port)) {
```
solmonk, line 166: nit: int64 casting seems redundant

pkg/gateway/model_build_targets.go
```diff
@@ -146,11 +163,11 @@ func (t *latticeTargetsModelBuildTask) buildLatticeTargets(ctx context.Context)
...
 						targetList = append(targetList, target)
-						glog.V(6).Infof("portAnnotations:%v, target.Port:%v\n", portAnnotations, target.Port)
+						glog.V(6).Infof("Found a port match, registering the target - port:%v, containerPort:%v, taerget:%v ***\n", int64(target.Port), definedPorts[0], target)
```
solmonk, line 168: typo: `taerget`

solmonk (changes requested):

liwenwu-amazon:
I think we should test e2e on following config `mutation(s)` and both in happy path and non-happy-path/mismatch and back and forth 

* modify the port in backendref of HTTPRoute 
* modify the port in K8S service object

test/pkg/test/nginxapp.go
sesmaili: done

pkg/gateway/model_build_targets.go
sesmaili: Fixed

pkg/gateway/model_build_targets.go
sesmaili: Fixed

pkg/gateway/model_build_targets.go
sesmaili: Done

test/pkg/test/nginxapp.go
sesmaili: yes

pkg/gateway/model_build_targets.go
rmullinnix461332: The model_build_targetgroup.go BuildTargets is already iterating through the list of backendRefs.  If you add 'port' to latticeTargetsModelBuildTask, you can reference that when iterating through the endPoint ports, instead of iterating through the rules again.  I made some changes a while back but did not submit PR.  I'll submit just for reference.

solmonk: Thank you for the reference, I think the approach makes sense. Currently there is a separate ongoing effort regarding this matter, will make sure this is considered.

rlymbur:
Issue was corrected in #335, closing.

Issue #282, opened 2023-06-29
HTTPRoute status update triggered indefinitely


solmonk:
Likely caused by #275 - HTTPRoute status update triggers another reconciliation which ends up triggering another HTTPRoute status update. The status update should happen only once as no specifications are changing.

Possible root cause is that status conditions keeps getting updated due to the timestamp value.

jansvive:
Thanks for reporting the issue, we will be investigating it further. Please watch this channel for more updates, and feel free to reach out.

PR #287, opened 2023-07-05
Fix condition handling of gateways and routes

### LINKED ISSUE #277: Set status condition on HTTPRoute resources
Need to state different kind of reasons when a route is accepted but configs are invalid inside.

solmonk:
**What type of PR is this?**

bug

Follow-up to a fix of #282 - also partially fixes #277 

Total rework of conditions update - now conditions are handled individually, and got rid of infinite loop and missing updates

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/controllers/gateway_controller.go
+++ b/controllers/gateway_controller.go
@@ -415,6 +415,7 @@ func UpdateGWListenerStatus(ctx context.Context, k8sclient client.Client, gw *ga
 				Status:             metav1.ConditionFalse,
 				Reason:             string(gateway_api.ListenerReasonInvalidRouteKinds),
 				ObservedGeneration: gw.Generation,
+				LastTransitionTime: metav1.Now(),
 			}
 			listenerStatus.SupportedKinds = supportedkind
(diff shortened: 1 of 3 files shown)


controllers/gateway_controller.go
```diff
@@ -258,21 +256,15 @@ func (r *GatewayReconciler) cleanupGatewayResources(ctx context.Context, gw *gat
 
 func (r *GatewayReconciler) updateGatewayStatus(ctx context.Context, serviceNetworkStatus *latticestore.ServiceNetwork, gw *gateway_api.Gateway) error {
 
-	glog.V(6).Infof("updateGatewayStatus: updating last transition time \n")
```
liwenwu-amazon, line 261: can we leave this for debugging?

liwenwu-amazon (approved):
LGTM, had some minor comments that we should keep some debugging msgs for trouble shooting.

controllers/gateway_controller.go
solmonk: updateGatewayStatus is only called in a single place and there is a log beforehand, so it is a duplicate IMO

PR #306, opened 2023-08-02
[Draft don't merge, demo purpose only] e2etest improvement

zijun726911:


test/pkg/test/framework.go
```diff
@@ -167,7 +171,7 @@ func (env *Framework) ExpectToBeClean(ctx context.Context) {
 	}).Should(Succeed())
 }
 
-func (env *Framework) CleanTestEnvironment(ctx context.Context) {
+func (env *Framework) DeleteAllTestCasesCreatedK8sResourceAndVpcLatticeResource(ctx context.Context) {
```
mikhail-aws, line 174: please use shorter names, hard to read

test/pkg/test/framework.go
```diff
@@ -313,6 +318,36 @@ func (env *Framework) GetTargets(ctx context.Context, targetGroup *vpclattice.Ta
 	return found
 }
 
+func (env *Framework) ExpectEventuallyVpcOfCurrentClusterAssociateWithServiceNetwork(ctx context.Context, sn *vpclattice.ServiceNetworkSummary) {
+	currentClusterVpcId := os.Getenv("CLUSTER_VPC_ID")
+	fmt.Println("ExpectEventuallyVpcOfCurrentClusterAssociateWithServiceNetwork currentClusterVpcId:", currentClusterVpcId)
```
mikhail-aws, line 323: 1. please use log not fmt 
2. ExpectEventuallyVpcOfCurrentClusterAssociateWithServiceNetwork can you use shorter names please?
 expect - eventually - these words are implementation details, not a function name

test/pkg/test/framework.go
```diff
@@ -313,6 +318,36 @@ func (env *Framework) GetTargets(ctx context.Context, targetGroup *vpclattice.Ta
...
+}
+
+func (env *Framework) ExpectEventuallyAllLatticeTargetsHealthy(ctx context.Context, tg *vpclattice.TargetGroupSummary) {
+	fmt.Printf("ExpectEventuallyAllLatticeTargetsHealthy for targetGroup: %+v\n", tg)
+	Eventually(func(g Gomega) {
+		targets, err := env.LatticeClient.ListTargetsAsList(ctx, &vpclattice.ListTargetsInput{TargetGroupIdentifier: tg.Id})
+		g.Expect(err).To(BeNil())
+		g.Expect(targets).ToNot(BeEmpty())
+		for _, target := range targets {
+			log.Println("target:", target)
```
mikhail-aws, line 345: log.Printf("target: %v (or %+v)", target)

test/pkg/test/framework.go
```diff
@@ -313,6 +318,36 @@ func (env *Framework) GetTargets(ctx context.Context, targetGroup *vpclattice.Ta
...
+			g.Expect(*target.Status).To(Equal(vpclattice.TargetStatusHealthy))
+		}
+	}).WithPolling(time.Second * 5).WithTimeout(time.Minute * 3).Should(Succeed())
```
mikhail-aws, line 348: I dont think pulling every 5 seconds is required here, can be relaxed to 10-15

test/pkg/test/framework.go
```diff
@@ -313,6 +318,36 @@ func (env *Framework) GetTargets(ctx context.Context, targetGroup *vpclattice.Ta
...
+		g.Expect(vpcAssociations).To(HaveLen(1))
+		for _, association := range vpcAssociations {
+			g.Expect(*association.Status).To(Equal(vpclattice.ServiceNetworkVpcAssociationStatusActive))
+		}
+	}).Should(Succeed())
+
+}
+
+func (env *Framework) ExpectEventuallyAllLatticeTargetsHealthy(ctx context.Context, tg *vpclattice.TargetGroupSummary) {
+	fmt.Printf("ExpectEventuallyAllLatticeTargetsHealthy for targetGroup: %+v\n", tg)
```
mikhail-aws, line 339: log.Printf

test/pkg/test/framework.go
```diff
@@ -528,12 +563,13 @@ func (env *Framework) DeleteAllFrameworkTracedTargetGroups(ctx aws.Context) {
 					g.Expect(err.(awserr.Error).Code()).To(Equal(vpclattice.ErrCodeResourceNotFoundException))
 				}
 			}
-		}).WithTimeout(360 * time.Second).Should(Succeed())
+		}).WithTimeout(8 * time.Minute).Should(Succeed())
```
mikhail-aws, line 566: please check body of the "eventually", if err == nil function succeed?

test/pkg/test/framework.go
```diff
@@ -528,12 +563,13 @@ func (env *Framework) DeleteAllFrameworkTracedTargetGroups(ctx aws.Context) {
...
 
 	}
 	env.TestCasesCreatedServiceNames = make(map[string]bool)
 }
 
+// TODO: Change this function name to GetVpcLatticeServiceDnsFromHttpRoute()
```
mikhail-aws, line 572: change function name as TODO? why not now?

test/suites/integration/deregister_targets_test.go
```diff
@@ -42,33 +42,33 @@ var _ = Describe("Deregister Targets", func() {
...
-		// Verify VPC Lattice Service exists
-		vpcLatticeService = testFramework.GetVpcLatticeService(ctx, pathMatchHttpRoute)
-
-		Expect(*vpcLatticeService.DnsEntry).To(ContainSubstring(
-			latticestore.LatticeServiceName(pathMatchHttpRoute.Name, pathMatchHttpRoute.Namespace)))
+		Eventually(func(g Gomega) {
+			// Put vpcLatticeService verification logic in the Eventually block(), because the controller need some time to create vpcLattice resource
+			vpcLatticeService = testFramework.GetVpcLatticeService(ctx, pathMatchHttpRoute)
+			g.Expect(vpcLatticeService).NotTo(BeNil())
+			Expect(*vpcLatticeService.DnsEntry).To(ContainSubstring(
```
mikhail-aws, line 49: g.Expect

test/suites/integration/deregister_targets_test.go
```diff
@@ -42,33 +42,33 @@ var _ = Describe("Deregister Targets", func() {
...
-		time.Sleep(3 * time.Minute) // Wait for creation of VPCLattice resources
-
-		// Verify VPC Lattice Service exists
-		vpcLatticeService = testFramework.GetVpcLatticeService(ctx, pathMatchHttpRoute)
-
-		Expect(*vpcLatticeService.DnsEntry).To(ContainSubstring(
-			latticestore.LatticeServiceName(pathMatchHttpRoute.Name, pathMatchHttpRoute.Namespace)))
+		Eventually(func(g Gomega) {
+			// Put vpcLatticeService verification logic in the Eventually block(), because the controller need some time to create vpcLattice resource
+			vpcLatticeService = testFramework.GetVpcLatticeService(ctx, pathMatchHttpRoute)
```
mikhail-aws, line 47: hidden assertions in GetVpcLatticeService, please use latticeClient directly (same for other comments about nested assertions)

test/suites/integration/deregister_targets_test.go
```diff
@@ -42,33 +42,33 @@ var _ = Describe("Deregister Targets", func() {
...
 			deployment,
 		)
-		time.Sleep(3 * time.Minute) // Wait for creation of VPCLattice resources
-
-		// Verify VPC Lattice Service exists
-		vpcLatticeService = testFramework.GetVpcLatticeService(ctx, pathMatchHttpRoute)
-
-		Expect(*vpcLatticeService.DnsEntry).To(ContainSubstring(
-			latticestore.LatticeServiceName(pathMatchHttpRoute.Name, pathMatchHttpRoute.Namespace)))
+		Eventually(func(g Gomega) {
```
mikhail-aws, line 45: you probably need to split one large eventually into smaller ones for each API call

test/suites/integration/deregister_targets_test.go
```diff
@@ -42,33 +42,33 @@ var _ = Describe("Deregister Targets", func() {
...
+			g.Expect(vpcLatticeService).NotTo(BeNil())
+			Expect(*vpcLatticeService.DnsEntry).To(ContainSubstring(
+				latticestore.LatticeServiceName(pathMatchHttpRoute.Name, pathMatchHttpRoute.Namespace)))
 
-		// Verify VPC Lattice Target Group exists
-		targetGroup = testFramework.GetTargetGroup(ctx, service)
-		Expect(*targetGroup.VpcIdentifier).To(Equal(os.Getenv("CLUSTER_VPC_ID")))
-		Expect(*targetGroup.Protocol).To(Equal("HTTP"))
+			// Verify VPC Lattice Target Group exists
+			targetGroup = testFramework.GetTargetGroup(ctx, service)
```
mikhail-aws, line 53: hidden assertions here without using g.*

test/suites/integration/deregister_targets_test.go
```diff
@@ -83,16 +83,16 @@ var _ = Describe("Deregister Targets", func() {
 		Eventually(func(g Gomega) {
 			log.Println("Verifying VPC lattice Targets deregistered")
 			targets := testFramework.GetTargets(ctx, targetGroup, deployment)
```
mikhail-aws, line 85: hidden assertions here

test/suites/integration/deregister_targets_test.go
```diff
@@ -83,16 +83,16 @@ var _ = Describe("Deregister Targets", func() {
...
-			Expect(len(targets) == 0)
-		}).WithTimeout(5*time.Minute + 10*time.Second)
+			g.Expect(targets).To(HaveLen(0))
+		}).WithTimeout(5*time.Minute + 10*time.Second).Should(Succeed())
```
mikhail-aws, line 87: A general comments about eventually. Please use eventually as assertion block and not side effect of it - a nested assertions.
```
Eventually(func() int {
  targets, err := latticeClient.getTargets()
  if err != nil { return -1 }
  return len(targets)
}).WithPolling(time.Minute).WithTimeout(10*time.Minutes).Should(Equal(0))
```

test/suites/integration/httproute_header_match_test.go
```diff
@@ -30,9 +27,16 @@ var _ = Describe("HTTPRoute header matches", func() {
 			service3,
 			deployment3)
 
-		time.Sleep(3 * time.Minute)
-		vpcLatticeService := testFramework.GetVpcLatticeService(ctx, headerMatchHttpRoute)
+		var vpcLatticeService *vpclattice.ServiceSummary
+		Eventually(func(g Gomega) {
+			// Put vpcLatticeService verification logic in the Eventually block(), because the controller need some time to create vpcLattice resource
+			vpcLatticeService = testFramework.GetVpcLatticeService(ctx, headerMatchHttpRoute)
```
mikhail-aws, line 33: nested assertions

test/suites/integration/httproute_path_match_test.go
```diff
@@ -38,36 +36,32 @@ var _ = Describe("HTTPRoute path matches", func() {
...
 		)
-		time.Sleep(3 * time.Minute) //Need some time to wait for VPCLattice resources to be created
-
-		// Verify VPC Lattice Resource
-		vpcLatticeService := testFramework.GetVpcLatticeService(ctx, pathMatchHttpRoute)
-		Expect(*vpcLatticeService.DnsEntry).To(ContainSubstring(latticestore.LatticeServiceName(pathMatchHttpRoute.Name, pathMatchHttpRoute.Namespace)))
+		var vpcLatticeService *vpclattice.ServiceSummary
+		Eventually(func(g Gomega) {
+			// Put vpcLatticeService verification logic in the Eventually block(), because the controller need some time to create vpcLattice resource
+			vpcLatticeService = testFramework.GetVpcLatticeService(ctx, pathMatchHttpRoute)
```
mikhail-aws, line 42: nested assertions

test/pkg/test/framework.go
zijun726911: ok, will change to log

I think using `expect - eventually` could make the caller more clearer to know that, within this function, it use Expect() to do some assertion and use eventually to do the retry

test/pkg/test/framework.go
zijun726911: will change

test/pkg/test/framework.go
zijun726911: For vpc lattice api, delete success could have 2 kind of respone:

- 204 Not content (err==nil) 
- 404 Not Found (err!=nil and err with vpclattice.ErrCodeResourceNotFoundException)
I think both scenarios mean delete success 
image

image

test/pkg/test/framework.go
zijun726911: Ok, will do name change in current or next PR

test/pkg/test/framework.go
mikhail-aws: In this case I would use name like EnsureSnVpcAssociation. Function returns something, for example error or bool if there is no association, then you propagate this to caller into a main test function, and to eventually block there.

test/pkg/test/framework.go
mikhail-aws: Not a fan of relying on default polling rates, API's are different. Changes in default can cause large blast radius. Please be explicit here

mikhail-aws:
whats the state of PR?

Issue #310, opened 2023-08-03
dont use global variables as config


mikhail-aws:
https://github.com/aws/aws-application-networking-k8s/blob/v0.0.15/pkg/config/controller_config.go

Using variables like this considered as anti-pattern of handling configurations.
https://dave.cheney.net/practical-go/presentations/qcon-china.html#_avoid_package_level_state

```
var VpcID = UnknownInput
var AccountID = UnknownInput
var Region = UnknownInput
var logLevel = defaultLogLevel
var DefaultServiceNetwork = UnknownInput
var UseLongTGName = false
```

We need to wrap these configs into a struct and propagate struct through the code. Can be done through context, manager, controllers.

PR #325, opened 2023-08-11
fix: issue #310 remove config global vars

rmullinnix461332:
**What type of PR is this?**
cleanup

#310 

removes global vars.  adds getters and a few setter functions to retrieve var values.

existing tes cases
**Automation added to e2e**:

no, no functional change

no
no

no
```release-note
Removed global config vars and replaced with GetXXX() functions that read env vars.  Set function for VPC only called from test case.
```


mikhail-aws:
The original problem is that we stick values to the package globals. Wrapping globals into getter's and setter's does not address underlying issue. I do like wrapping each variable into own function, looks neat.

To address globals issue we need:
- create a struct that will hold configurations, not package variables
- initialize struct in main.go
- propagate config through constructors or context to the places where we need then

rmullinnix461332 (commented):
the main issue I see with this is the amount of calls that originate from _test.go modules.  The init() in main() doesn't cover the test runs.  You end up having to InitConfig() in every test function or add a TestMain().  Would it be ok to have a separate test config that is like what I coded - just reading from env vars.  Then the controller_config for main() and use in packages.  Trying to pass the config struct through context in test cases, especially the mock client was problematic.

mikhail-aws:
>  You end up having to InitConfig() in every test function or add a TestMain()

There are other options. For example you can create a config struct in controller_config.go
```go
type Config struct {
	VpcId                 string
	AccountId             string
	Region                string
	LogLevel              string
	DefaultServiceNetwork string
	UserLongTGName        bool
}
```

in tests for a package create an instance of config - global package variable for tests. For example pkg/deploy/lattice/config_test.go
```go
var testConfig = Config { ... populate with fields ... }
```

And then in current test file we only need to replace "config.VpcId" with "testConfig.VpcId". Which is basic find/replace.
```go
latticeDataStore.AddServiceNetwork(tt.meshName, config.AccountID, tt.meshArn, tt.meshId, latticestore.DATASTORE_SERVICE_NETWORK_CREATED)

// will become

latticeDataStore.AddServiceNetwork(tt.meshName, testConfig.AccountID, tt.meshArn, tt.meshId, latticestore.DATASTORE_SERVICE_NETWORK_CREATED)
```

this helps with 2 things:
1. decouple variables from package into struct, with minimal changes in current code
2. makes it easier to propagate config in future refactoring

mikhail-aws (changes requested):
It's very large diff. Considering lots of changes currently happening it's pretty hard to do it in one shot. I still recommend to cut down a scope and do little changes starting from controller_config.go wrapping configs into a struct. A start config propagation little by little. It's more than 1 PR.

mikhail-aws:
closing due to inactivity, also some of configuration changes already merged into cloud.go

PR #327, opened 2023-08-15
add aws-sdk-go to tests

mikhail-aws:
fix for missing aws-sdk in tests module after removing sdk override


rlymbur:
Thanks for the fix!

Issue #168, opened 2023-03-27
EKS IPv6 target support


Tim-AWS:
When we deploy IPv6 clusters in EKS the pods are only given IPv6 but currently Lattice Target Groups for EKS only support IPv4 therefore IPv6 pods can't be targets. Would be great to add IPv6 to Lattice for EKS backends.

solmonk:
Should be supported with #329

PR #329, opened 2023-08-16
feat: add IPv6 target group support

scottlaiaws:
**What type of PR is this?**
feature

#168 

This PR adds IPv6 support for Service

`make e2etest` ran in an IPv6 clustere outputs:

```
Ran 11 of 11 Specs in 4135.156 seconds
--- PASS: TestIntegration (4135.90s)
PASS
ok      github.com/aws/aws-application-networking-k8s/test/suites/integration   4135.930s
```

**Automation added to e2e**: N/A. All existing e2e tests can be run a IPv6 cluster

**Will this break upgrades or downgrades. Has updating a running cluster been tested?**: N/A

**Does this PR introduce any user-facing change?**: Yes

```release-note
IPv6 Service and Pod supported
```

### DIFF
--- a/docs/getstarted.md
+++ b/docs/getstarted.md
@@ -230,3 +230,32 @@ The following figure illustrates this:
    Requsting to Pod(inventory-ver1-74fc59977-wg8br): Inventory-ver1 handler pod....
    ```
    You can see that the traffic is distributed between *inventory-ver1* and *inventory-ver2* as expected.
+
+## IPv6 Support
+
+IPv6 address type is automatically used for your services and pods if
+[your cluster is configured to use IPv6 addresses](https://docs.aws.amazon.com/eks/latest/userguide/cni-ipv6.html).
+
+```bash
+# To create an IPv6 cluster
+kubectl apply -f examples/ipv6-cluster.yaml
+```
+
+If your cluster is configured to be dual-stack, you can set the IP address type
+of your service using the `ipFamilies` field. For example:
+
+```yaml
+apiVersion: v1
+kind: Service
+metadata:
+  name: ipv4-target-in-dual-stack-cluster
+spec:
+  ipFamilies:
+    - "IPv4"
+  selector:
+    app: parking
+  ports:
+    - protocol: TCP
+      port: 80
+      targetPort: 8090
+```
--- a/examples/ipv6-cluster.yaml
+++ b/examples/ipv6-cluster.yaml
@@ -1,0 +1,25 @@
+---
+apiVersion: eksctl.io/v1alpha5
+kind: ClusterConfig
+
+metadata:
+  name: my-ipv6-cluster
+  region: us-west-2
+
+kubernetesNetworkConfig:
+  ipFamily: IPv6
+
+addons:
+  - name: vpc-cni
+    version: latest
+  - name: coredns
+    version: latest
+  - name: kube-proxy
+    version: latest
+
+iam:
+  withOIDC: true
+
+managedNodeGroups:
+  - name: my-ipv6-nodegroup
+    instanceType: t3.medium
--- a/pkg/gateway/model_build_targetgroup.go
+++ b/pkg/gateway/model_build_targetgroup.go
@@ -2,6 +2,8 @@ package gateway
 
 import (
 	"context"
+	"errors"
+	"fmt"
 
 	"github.com/golang/glog"
 
@@ -186,6 +188,12 @@ func (t *targetGroupModelBuildTask) BuildTargetGroup(ctx context.Context) error 
 		return err
 	}
 
+	ipAddressType, err := buildTargetGroupIpAdressType(svc)
+
+	if err != nil {
+		return err
+	}
+
 	//if t.serviceExport.
 	tgSpec := latticemodel.TargetGroupSpec{
 		Name: tgName,
@@ -200,6 +208,7 @@ func (t *targetGroupModelBuildTask) BuildTargetGroup(ctx context.Context) error 
 			K8SServiceNamespace: t.serviceExport.Namespace,
 			Protocol:            "HTTP",
 			ProtocolVersion:     vpclattice.TargetGroupProtocolVersionHttp1,
+			IpAddressType:       ipAddressType,
 		},
 	}
 
@@ -235,7 +244,6 @@ func (t *targetGroupModelBuildTask) BuildTargetGroup(ctx context.Context) error 
 
 // Build target group for backend service ref used in HTTPRoute
 func (t *latticeServiceModelBuildTask) buildTargetGroup(ctx context.Context, client client.Client) (*latticemodel.TargetGroup, error) {
-
 	for _, httpRule := range t.route.Spec().Rules() {
 		glog.V(6).Infof("buildTargetGroup: %v\n", httpRule)
 
@@ -304,6 +312,9 @@ func (t *latticeServiceModelBuildTask) buildTargetGroupSpec(ctx context.Context,
 	} else {
 		isDeleted = true
 	}
+
+	ipAddressType := vpclattice.IpAddressTypeIpv4
+
 	if backendKind == "ServiceImport" {
 		isServiceImport = true
 		namespaceName := types.NamespacedName{
@@ -348,6 +359,15 @@ func (t *latticeServiceModelBuildTask) buildTargetGroupSpec(ctx context.Context,
 				return latticemodel.TargetGroupSpec{}, err
 			}
 		}
+
+		var err error
+
+		ipAddressType, err = buildTargetGroupIpAdressType(svc)
+
+		// Ignore error for creation request
+		if !isDeleted && err != nil {
+			return latticemodel.TargetGroupSpec{}, err
+		}
 	}
 
 	tgName := latticestore.TargetGroupName(string(httpBackendRef.Name()), namespace)
@@ -367,7 +387,8 @@ func (t *latticeServiceModelBuildTask) buildTargetGroupSpec(ctx context.Context,
 			Protocol:              "HTTP",
 			ProtocolVersion:       vpclattice.TargetGroupProtocolVersionHttp1,
 			// Fill in default HTTP port as we are using target port anyway.
-			Port: 80,
+			Port:          80,
+			IpAddressType: ipAddressType,
 		},
 		IsDeleted: isDeleted,
 	}, nil
@@ -380,5 +401,27 @@ func (t *latticeServiceModelBuildTask) buildTargetGroupName(_ context.Context, b
 	} else {
 		return latticestore.TargetGroupName(string(backendRef.Name()),
 			t.route.Namespace())
+	}
+}
+
+func buildTargetGroupIpAdressType(svc *corev1.Service) (string, error) {
+	ipFamilies := svc.Spec.IPFamilies
+
+	glog.V(6).Infof("buildTargetGroupIpAddressType ipFamilies: %v\n", ipFamilies)
+
+	if len(ipFamilies) != 1 {
+		return "", errors.New("Lattice Target Group only support single stack ip addresses")
+	}
+
+	// IpFamilies will always have at least 1 element
+	ipFamily := ipFamilies[0]
+
+	switch ipFamily {
+	case corev1.IPv4Protocol:
+		return vpclattice.IpAddressTypeIpv4, nil
+	case corev1.IPv6Protocol:
+		return vpclattice.IpAddressTypeIpv6, nil
+	default:
+		return "", fmt.Errorf("unknown ipFamily: %v", ipFamily)
 	}
 }
--- a/controllers/gateway_controller.go
+++ b/controllers/gateway_controller.go
@@ -29,6 +29,7 @@ import (
 	"github.com/aws/aws-application-networking-k8s/pkg/model/core"
 	latticemodel "github.com/aws/aws-application-networking-k8s/pkg/model/lattice"
 	lattice_runtime "github.com/aws/aws-application-networking-k8s/pkg/runtime"
+	"github.com/aws/aws-application-networking-k8s/pkg/utils"
 	"github.com/aws/aws-application-networking-k8s/pkg/utils/gwlog"
 
 	"github.com/aws/aws-application-networking-k8s/controllers/eventhandlers"
@@ -268,7 +269,7 @@ func (r *GatewayReconciler) updateGatewayStatus(ctx context.Context, serviceNetw
 
 	gwOld := gw.DeepCopy()
 
-	gw.Status.Conditions = updateCondition(gw.Status.Conditions, metav1.Condition{
+	gw.Status.Conditions = utils.GetNewConditions(gw.Status.Conditions, metav1.Condition{
 		Type:               string(gateway_api.GatewayConditionProgrammed),
 		Status:             metav1.ConditionTrue,
 		ObservedGeneration: gw.Generation,
@@ -307,7 +308,7 @@ func (r *GatewayReconciler) updateGatewayAcceptStatus(ctx context.Context, gw *g
 			Reason:             string(gateway_api.GatewayReasonInvalid),
 		}
 	}
-	gw.Status.Conditions = updateCondition(gw.Status.Conditions, cond)
+	gw.Status.Conditions = utils.GetNewConditions(gw.Status.Conditions, cond)
 
 	if err := r.client.Status().Patch(ctx, gw, client.MergeFrom(gwOld)); err != nil {
 		return fmt.Errorf("update gateway status error, gw: %s, accepted: %t, err: %s", gw.Name, accepted, err)
@@ -458,7 +459,7 @@ func UpdateGWListenerStatus(ctx context.Context, k8sclient client.Client, gw *ga
 				gw.Status.Listeners[i].AttachedRoutes = listenerStatus.AttachedRoutes
 				gw.Status.Listeners[i].SupportedKinds = listenerStatus.SupportedKinds
 				// Only have one condition in the logic
-				gw.Status.Listeners[i].Conditions = updateCondition(gw.Status.Listeners[i].Conditions, listenerStatus.Conditions[0])
+				gw.Status.Listeners[i].Conditions = utils.GetNewConditions(gw.Status.Listeners[i].Conditions, listenerStatus.Conditions[0])
 				found = true
 			}
 		}
--- a/controllers/route_controller.go
+++ b/controllers/route_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+
 	"github.com/aws/aws-application-networking-k8s/pkg/utils/gwlog"
 
 	"github.com/pkg/errors"
@@ -32,6 +33,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 	gateway_api_v1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+	gateway_api "sigs.k8s.io/gateway-api/apis/v1beta1"
 	gateway_api_v1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 	mcs_api "sigs.k8s.io/mcs-api/pkg/apis/v1alpha1"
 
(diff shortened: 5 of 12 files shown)


scottlaiaws:
`make presubmit` output:

```
go mod tidy
go generate ./...
go vet ./...
go fmt ./...
go test ./pkg/... -coverprofile coverage.out
ok      github.com/aws/aws-application-networking-k8s/pkg/aws/services  0.226s  coverage: 4.4% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/config        50.484s coverage: 77.4% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/deploy        0.382s  coverage: 20.6% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/deploy/externaldns    0.584s  coverage: 77.8% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/deploy/lattice        0.791s  coverage: 86.2% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/gateway       0.945s  coverage: 78.5% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/latticestore  0.843s  coverage: 73.3% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/model/core    1.052s  coverage: 22.6% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/model/core/graph      1.048s  coverage: 17.2% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/utils/log     0.811s  coverage: 0.0% of statements [no tests to run]
```

controllers/httproute_controller.go
```diff
@@ -379,3 +409,45 @@ func isExternalDNSSupported(mgr ctrl.Manager) (bool, error) {
 	}
 	return false, nil
 }
+
+func (r *HTTPRouteReconciler) areBackendRefsIpFamiliesValid(ctx context.Context, httpRoute *gateway_api.HTTPRoute) bool {
```
mikhail-aws, line 413: can you return error here not bool, with explicit error message

controllers/httproute_controller.go
```diff
@@ -254,6 +254,30 @@ func (r *HTTPRouteReconciler) reconcileHTTPRouteResource(ctx context.Context, ht
...
+		updateRouteCondition(httpRoute, metav1.Condition{
+			Type:               string(gateway_api.RouteConditionAccepted),
+			Status:             metav1.ConditionFalse,
+			ObservedGeneration: httpRoute.Inner().Generation,
+			Reason:             string(gateway_api.RouteReasonUnsupportedValue),
+			Message:            fmt.Sprintf("Dual stack Service is not supported"),
+		})
+
+		if err := r.client.Status().Patch(ctx, httpRoute.Inner(), client.MergeFrom(httpRouteOld)); err != nil {
+			glog.V(2).Infof("updateHTTPRouteStatus: Patch() received err %v \n", err)
```
mikhail-aws, line 273: Use error log. No need "\n" at the end. %s, not %v for errors.

controllers/httproute_controller.go
```diff
@@ -254,6 +254,30 @@ func (r *HTTPRouteReconciler) reconcileHTTPRouteResource(ctx context.Context, ht
...
+
+			return errors.Wrapf(err, "failed to update httproute status")
+		}
+
+		return errors.New("Dual stack Service is not supported")
```
mikhail-aws, line 278: can you add to error message what kind of dual stack it is

controllers/httproute_controller.go
```diff
@@ -379,3 +409,45 @@ func isExternalDNSSupported(mgr ctrl.Manager) (bool, error) {
...
+
+		for _, backendRef := range backendRefs {
+			// For now we skip checking service import
+			if *backendRef.Kind == "ServiceImport" {
+				continue
+			}
+
+			svc := &corev1.Service{}
+
+			glog.Infof("backendRef: %v", backendRef)
```
mikhail-aws, line 427: redundant log

controllers/httproute_controller.go
```diff
@@ -379,3 +409,45 @@ func isExternalDNSSupported(mgr ctrl.Manager) (bool, error) {
...
+
+			key := types.NamespacedName{
+				// Namespace: string(*backendRef.Namespace),
```
mikhail-aws, line 430: why commented?

controllers/httproute_controller.go
```diff
@@ -379,3 +409,45 @@ func isExternalDNSSupported(mgr ctrl.Manager) (bool, error) {
...
+			}
+
+			if backendRef.Namespace != nil {
+				key.Namespace = string(*backendRef.Namespace)
+			} else {
+				key.Namespace = httpRoute.Namespace
+			}
+
+			if err := r.client.Get(ctx, key, svc); err != nil {
+				// Ignore error since Service might not be created yet
```
mikhail-aws, line 441: if service is not created we should requeue reconcile request, otherwise you assume here that route is valid, even if it's dual stack

controllers/httproute_controller.go
```diff
@@ -379,3 +409,45 @@ func isExternalDNSSupported(mgr ctrl.Manager) (bool, error) {
...
+				glog.Infof("error: %v", err)
```
mikhail-aws, line 442: if you dont care about error, why logging it? also should be debug log

pkg/gateway/model_build_targetgroup.go
```diff
@@ -179,6 +180,14 @@ func (t *targetGroupModelBuildTask) BuildTargetGroup(ctx context.Context) error
 		return err
 	}
 
+	glog.V(6).Infof("svc.Spec.IPFamilies: %v\n", svc.Spec.IPFamilies)
```
mikhail-aws, line 183: please dont use %v and \n, be specific with logging, no need to dump entire struct into log, choose fields that reader cares about

pkg/gateway/model_build_targetgroup.go
```diff
@@ -375,3 +399,29 @@ func (t *latticeServiceModelBuildTask) buildTargetGroupName(_ context.Context, b
...
+func buildTargetGroupIpAddressType(_ context.Context, svc *corev1.Service) (string, error) {
+	ipFamilies := svc.Spec.IPFamilies
+
+	isIPv6 := false
+
+	glog.V(6).Infof("buildTargetGroupIpAddressType ipFamilies: %v\n", ipFamilies)
+
+	// Since Lattice Target Group doesn't support dual stack
+	if len(ipFamilies) > 1 {
+		return "", errors.New("Invalid IpFamilies, Lattice Target Group doesn't support dual stack ip addresses")
```
mikhail-aws, line 412: looks like it's second place that return dual stack error, if we catch dual stack before, why it's happening here?

pkg/gateway/model_build_targetgroup.go
```diff
@@ -375,3 +399,29 @@ func (t *latticeServiceModelBuildTask) buildTargetGroupName(_ context.Context, b
...
+			isIPv6 = true
+			break
+		}
+	}
+
+	if isIPv6 {
+		return vpclattice.IpAddressTypeIpv6, nil
+	}
+
+	return vpclattice.IpAddressTypeIpv4, nil
```
mikhail-aws, line 426: considering ipFamilies should be size of 1, you dont need range

pkg/gateway/model_build_targetgroup.go
```diff
@@ -341,6 +355,15 @@ func (t *latticeServiceModelBuildTask) buildTargetGroupSpec(ctx context.Context,
 				return latticemodel.TargetGroupSpec{}, err
 			}
 		}
+
+		var err error
```
mikhail-aws, line 359: nit: you can remove it

pkg/gateway/model_build_targetgroup.go
```diff
@@ -341,6 +355,15 @@ func (t *latticeServiceModelBuildTask) buildTargetGroupSpec(ctx context.Context,
...
+
+		ipAddressType, err = buildTargetGroupIpAddressType(ctx, svc)
+
+		// Ignore error for creation request
```
mikhail-aws, line 363: not clear why should ignore error

mikhail-aws: also looking into buildTargetGroupIpAddressType you return dual stack error, which seems to be hard stop, confusing

controllers/httproute_controller.go
scottlaiaws: > what kind of dual stack it is

I don't understand what do you mean by that?

controllers/httproute_controller.go
scottlaiaws: Thanks, will remove it in next revision

controllers/httproute_controller.go
scottlaiaws: Thanks, I forgot to remove that while I was doing manual testing\

pkg/gateway/model_build_targetgroup.go
scottlaiaws: Will remove in next revision. Was using that for debugging purpose 😅

pkg/gateway/model_build_targetgroup.go
scottlaiaws: For deletion requests, we still want to build the model just like the code above

```go
		svc := &corev1.Service{}
		if err := t.Client.Get(ctx, serviceNamespaceName, svc); err != nil {
			glog.V(6).Infof("Error finding backend service %v error :%v \n", serviceNamespaceName, err)
			if !isDeleted {
				//Return error for creation request only,
				//For k8sService deletion request, we should go ahead to build TargetGroupSpec model
				return latticemodel.TargetGroupSpec{}, err
			}
		}
```

pkg/gateway/model_build_targetgroup.go
scottlaiaws: The reason i can't remove is because
```
ipAddressType, err := ...
```

will make `ipAddressType` a new variable inside the scope of the else statement instead of the one outside

pkg/gateway/model_build_targetgroup.go
```diff
@@ -287,6 +296,8 @@ func (t *latticeServiceModelBuildTask) buildTargetGroupSpec(ctx context.Context,
 	backendKind := string(*httpBackendRef.Kind())
 	glog.V(6).Infof("buildTargetGroupSpec,  kind %s\n", backendKind)
 
+	glog.V(6).Infof("buildHTTPTargetGroupSpec,  kind %s\n", backendKind)
+
```
zijun726911, line 300: In here it print `backendKind` two times?

controllers/httproute_controller.go
mikhail-aws: for example you have this error in another place, which looks nicer and more informative
`errors.New("Invalid IpFamilies, Lattice Target Group doesn't support dual stack ip addresses")`

pkg/gateway/model_build_targetgroup.go
mikhail-aws: Sorry I dont get it. buildTargetGroupIpAddressType returns single error - dual stack exception which seems critical error. This is the only place I see you use buildTargetGroupIpAddressType. So the error does not make any difference.
So you either need to handle error, or dont return it.

zijun726911: `// Ignore error for creation request`

Your comment should be : `// Ignore error for deletion request` ?

pkg/gateway/model_build_targetgroup.go
```diff
@@ -375,3 +399,29 @@ func (t *latticeServiceModelBuildTask) buildTargetGroupName(_ context.Context, b
...
+			break
+		}
+	}
+
+	if isIPv6 {
+		return vpclattice.IpAddressTypeIpv6, nil
+	}
+
+	return vpclattice.IpAddressTypeIpv4, nil
+}
```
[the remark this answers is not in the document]
zijun726911: Also, do you need the first parameter `_ context.Context`?

zijun726911:
Have you checked the vpc lattice console (or use vpcLattice.listTG() api) and can see the ipv6 type tgs created by the k8s controller?

docs/getstarted.md
```diff
@@ -230,3 +230,27 @@ The following figure illustrates this:
...
    You can see that the traffic is distributed between *inventory-ver1* and *inventory-ver2* as expected.
+
+## IPv6 Support
+
+IPv6 address type is automatically used for your services and pods if
+[your cluster is configured to use IPv6 addresses](https://docs.aws.amazon.com/eks/latest/userguide/cni-ipv6.html).
+
+If your cluster is configured to be dual-stack, you can set the IP address type
+of your service using the `ipFamilies` field. For example:
+
```
zijun726911, line 241: Seems EKS only support either ipv4 or ipv6 single stack? no dual stack?
https://docs.aws.amazon.com/eks/latest/userguide/cni-ipv6.html

```
Amazon EKS doesn't support dual-stacked Pods or services, even though Kubernetes does in version 1.23 and later. As a result, you can't assign both IPv4 and IPv6 addresses to your Pods and services. 
```

pkg/gateway/model_build_targetgroup.go
zijun726911: I think Scott's current code could work. Or change to this still work: 
```ipAddressType, _ = buildTargetGroupIpAddressType(ctx, svc)```

for the targetGroup deletion request, the `svc` is empty and buildTargetGroupIpAddressType(ctx, svc) return `"", err`, and in the `target_group_manager.go` it is able to handle the `targetGroup.Spec.Config.IpAddressType == ""`

For dual stack validation, `areBackendRefsIpFamiliesValid()` could handle it and  retrun early at  httproute_controller.go before  doing the`buildAndDeployModel()`

Currently the `buildTargetGroupSpec()` have a little bit high cyclomatic complexity and hard to read, we could refactor it and make it more clear to show how it build targetGroupSpec following 4 scenarios:
1.  ServiceImport creation
2. ServiceImport deletion
3. k8sService creation
4. k8sService deletion

File: pkg/gateway/model_build_targetgroup.go [hunk 16]
```diff
@@ -297,6 +308,9 @@ func (t *latticeServiceModelBuildTask) buildTargetGroupSpec(ctx context.Context,
 	} else {
 		isDeleted = true
 	}
+
+	ipAddressType := vpclattice.IpAddressTypeIpv4
+
```
zijun726911, line 313: Change to `ipAddressType := ""` can still work and more clear?

controllers/httproute_controller.go
scottlaiaws: i see, I'll change the error message to be the same as that.

controllers/httproute_controller.go
```diff
@@ -254,6 +254,30 @@ func (r *HTTPRouteReconciler) reconcileHTTPRouteResource(ctx context.Context, ht
 		r.eventRecorder.Event(httpRoute.K8sObject(), corev1.EventTypeWarning, k8s.HTTPRouteventReasonFailedAddFinalizer, fmt.Sprintf("Failed add finalizer due to %v", err))
 	}
 
+	ok := r.areBackendRefsIpFamiliesValid(ctx, httpRoute.Inner())
+
+	if !ok {
+		httpRouteOld := httpRoute.Inner().DeepCopy()
```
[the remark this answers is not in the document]
scottlaiaws: no, for `false` scenario, I think we need to have some more condition before we set the status of http route

docs/getstarted.md
scottlaiaws: Self-host cluster on EKS can support dual stack

pkg/gateway/model_build_targetgroup.go
mikhail-aws: lets keep it simple, please remove this error from buildTargetGroupIpAddressType since you dont use it

controllers/httproute_controller.go
scottlaiaws: Why is that?

mikhail-aws: Bool value does not tell much. If it's invalid and you return error, this error can be propagated back to the caller, caller will decide what to do. We should propagate error up to the stack and log at the controller level.

pkg/gateway/model_build_targetgroup.go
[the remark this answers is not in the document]
solmonk: switch case looks even better here

examples/ipv6-cluster.yaml
solmonk, line 1: can you maybe briefly mention in the doc on what this example means?

controllers/httproute_controller.go
```diff
@@ -254,6 +254,30 @@ func (r *HTTPRouteReconciler) reconcileHTTPRouteResource(ctx context.Context, ht
...
+			Message:            fmt.Sprintf("Dual stack Service is not supported"),
+		})
+
+		if err := r.client.Status().Patch(ctx, httpRoute.Inner(), client.MergeFrom(httpRouteOld)); err != nil {
+			glog.Errorf("updateHTTPRouteStatus: Patch() received err %s", err)
+
+			return errors.Wrapf(err, "failed to update httproute status")
+		}
+
+		errors.New("Invalid IpFamilies, Lattice Target Group doesn't support dual stack ip addresses")
```
solmonk, line 278: This line will not do anything. You may want to return earlier because we can't really create target group.

zijun726911 (approved):
LGTM, tho I think buildTargetGroupSpec() need some refactoring afterward

controllers/httproute_controller.go
solmonk: I think creation of the service should trigger reconcile request anyways. Looks like this function should only determine if the ipFamily is valid.

solmonk (changes requested):

controllers/httproute_controller.go
scottlaiaws: Thanks, forgot to put a `return`  there.

> I think creation of the service should trigger reconcile request anyways. Looks like this function should only determine if the ipFamily is valid.

I can create a `HTTPRoute` with `BackendRefs` to a non-existing Service right? In that case, the `Get` call would return an error which we want to ignore

pkg/gateway/model_build_targetgroup.go
scottlaiaws: We still use the error in the case of Creation requests

controllers/httproute_controller.go
solmonk: yeah I think it will eventually lead to lattice_retry on stack deployer later

controllers/httproute_controller.go
```diff
@@ -378,3 +408,40 @@ func isExternalDNSSupported(mgr ctrl.Manager) (bool, error) {
 	}
 	return false, nil
 }
+
+func (r *HTTPRouteReconciler) getBackendRefsIpFamiliesError(ctx context.Context, httpRoute *gateway_api.HTTPRoute) error {
```
mikhail-aws, line 412: getXxxError kind of weird name, yes it returns error, but function name should tell meaning not implementation.
Just looking at code I would assume it does validation for IP families, that can be a name

controllers/httproute_controller.go
```diff
@@ -253,6 +253,30 @@ func (r *HTTPRouteReconciler) reconcileHTTPRouteResource(ctx context.Context, ht
 		r.eventRecorder.Event(httpRoute.K8sObject(), corev1.EventTypeWarning, k8s.HTTPRouteventReasonFailedAddFinalizer, fmt.Sprintf("Failed add finalizer due to %v", err))
 	}
 
+	backendRefIPFamiliesErr := r.getBackendRefsIpFamiliesError(ctx, httpRoute.Inner())
+
+	if backendRefIPFamiliesErr != nil {
+		httpRouteOld := httpRoute.Inner().DeepCopy()
+
+		setupRouteStatusParents(httpRoute.Inner())
```
mikhail-aws, line 261: inverse ownership: mutating struct outside of struct, you should add a method to httpRoute with reference receiver:

```
func(r *httpRoute) setupRouteStatusParents() { ... }
```

controllers/httproute_controller.go
```diff
@@ -253,6 +253,30 @@ func (r *HTTPRouteReconciler) reconcileHTTPRouteResource(ctx context.Context, ht
...
+
+		updateRouteCondition(httpRoute, metav1.Condition{
```
mikhail-aws, line 263: same thing as above, update route condition should be method of http route that accept condition as argument.

controllers/httproute_controller.go
```diff
@@ -253,6 +253,30 @@ func (r *HTTPRouteReconciler) reconcileHTTPRouteResource(ctx context.Context, ht
...
+			Status:             metav1.ConditionFalse,
+			ObservedGeneration: httpRoute.Inner().Generation,
+			Reason:             string(gateway_api.RouteReasonUnsupportedValue),
+			Message:            fmt.Sprintf("Dual stack Service is not supported"),
+		})
+
+		if err := r.client.Status().Patch(ctx, httpRoute.Inner(), client.MergeFrom(httpRouteOld)); err != nil {
+			glog.Errorf("updateHTTPRouteStatus: Patch() received err %s", err)
+
+			return errors.Wrapf(err, "failed to update httproute status")
```
mikhail-aws, line 274: dont log when return error

controllers/httproute_controller.go
```diff
@@ -253,6 +253,30 @@ func (r *HTTPRouteReconciler) reconcileHTTPRouteResource(ctx context.Context, ht
...
+		}
+
+		return backendRefIPFamiliesErr
```
mikhail-aws, line 277: can you explain how two errors above interact with each other:
`if backendRefIPFamiliesErr != nil and r.client.Status().Patch != nil`

it seems if backend error happens we should not get to Patch call, but we do in fact, or I'm missing something

pkg/gateway/model_build_targetgroup.go
```diff
@@ -375,3 +396,25 @@ func (t *latticeServiceModelBuildTask) buildTargetGroupName(_ context.Context, b
 			t.route.Namespace())
 	}
 }
+
+func buildTargetGroupIpAddressType(_ context.Context, svc *corev1.Service) (string, error) {
```
mikhail-aws, line 400: - naming, what we are building here? to looks like getIpAddrType from Service
- context not in use, please remove
- please add unit test

zijun726911 (approved):
LGTM

mikhail-aws (approved):

mikhail-aws:
please run e2e tests with latest changes from main

PR #330, opened 2023-08-16
E2E test improvement, get rid of fixed wait time, trying to avoid lattice targets draining,fixing deregister_targets_test

zijun726911:
**What type of PR is this?**

E2E test code improvement

- Add two helper functions in the test framework IsVpcAssociateWithServiceNetwork() and IsAllLatticeTargetsHealthy()
- Get rid of above 3min fixed wait time, instead, use above helper functions and Eventually() block to make use vpc lattice connectivity setup is fully ready
- Fix bug in the deregister_targets_test.go test cases

**What does this PR do / Why do we need it**: E2E test code improvement

No

No

```release-note

```

## Test:
`make e2etest` success
```

Ran 10 of 11 Specs in 4600.977 seconds
SUCCESS! -- 10 Passed | 0 Failed | 0 Pending | 1 Skipped
--- PASS: TestIntegration (4601.32s)
PASS
ok  	github.com/aws/aws-application-networking-k8s/test/suites/integration	4601.876s
```

### DIFF
--- a/test/pkg/test/framework.go
+++ b/test/pkg/test/framework.go
@@ -2,6 +2,7 @@ package test
 
 import (
 	"context"
+	"fmt"
 	"log"
 	"os"
 	"reflect"
@@ -48,13 +49,14 @@ type TestObject struct {
 }
 
 var (
-	TestObjects = []TestObject{
-		{&v1.Service{}, &v1.ServiceList{}},
+	CurrentClusterVpcId = os.Getenv("CLUSTER_VPC_ID")
+	TestObjects         = []TestObject{
+		{&v1beta1.HTTPRoute{}, &v1beta1.HTTPRouteList{}},
 		{&v1alpha1.ServiceExport{}, &v1alpha1.ServiceExportList{}},
 		{&v1alpha1.ServiceImport{}, &v1alpha1.ServiceImportList{}},
-		{&appsv1.Deployment{}, &appsv1.DeploymentList{}},
-		{&v1beta1.HTTPRoute{}, &v1beta1.HTTPRouteList{}},
 		{&v1beta1.Gateway{}, &v1beta1.GatewayList{}},
+		{&appsv1.Deployment{}, &appsv1.DeploymentList{}},
+		{&v1.Service{}, &v1.ServiceList{}},
 	}
 )
 
@@ -67,7 +69,9 @@ type Framework struct {
 	TestCasesCreatedServiceNetworkNames map[string]bool //key: ServiceNetworkName; value: not in use, meaningless
 	TestCasesCreatedServiceNames        map[string]bool //key: ServiceName; value not in use, meaningless
 	TestCasesCreatedTargetGroupNames    map[string]bool //key: TargetGroupName; value: not in use, meaningless
-	TestCasesCreatedK8sResource         []client.Object
+	// TODO: instead of using one big list TestCasesCreatedK8sResource to track all created k8s resource,
+	//  we should create different lists for different kind of k8s resource i.e., httproute have 1 list, service have another list etc.
+	TestCasesCreatedK8sResource []client.Object
 }
 
 func NewFramework(ctx context.Context) *Framework {
@@ -86,8 +90,7 @@ func NewFramework(ctx context.Context) *Framework {
 		TestCasesCreatedServiceNames:        make(map[string]bool),
 		TestCasesCreatedTargetGroupNames:    make(map[string]bool),
 	}
-
-	SetDefaultEventuallyTimeout(180 * time.Second)
+	SetDefaultEventuallyTimeout(3 * time.Minute)
 	SetDefaultEventuallyPollingInterval(10 * time.Second)
 	BeforeEach(func() { framework.ExpectToBeClean(ctx) })
 	AfterEach(func() { framework.ExpectToBeClean(ctx) })
@@ -102,9 +105,8 @@ func (env *Framework) ExpectToBeClean(ctx context.Context) {
 		env.EventuallyExpectNoneFound(ctx, testObject.ListType)
 	})
 
-	currentClusterVpcId := os.Getenv("CLUSTER_VPC_ID")
 	retrievedServiceNetworkVpcAssociations, _ := env.LatticeClient.ListServiceNetworkVpcAssociationsAsList(ctx, &vpclattice.ListServiceNetworkVpcAssociationsInput{
-		VpcIdentifier: aws.String(currentClusterVpcId),
+		VpcIdentifier: aws.String(CurrentClusterVpcId),
 	})
 	Logger(ctx).Infof("Expect VPC used by current cluster don't have any ServiceNetworkVPCAssociation, if it has you should manually delete it")
 	Expect(len(retrievedServiceNetworkVpcAssociations)).To(Equal(0))
@@ -121,7 +123,7 @@ func (env *Framework) ExpectToBeClean(ctx context.Context) {
 
 				value, ok := retrievedTags.Tags[lattice.K8SServiceNetworkOwnedByVPC]
 				if ok {
-					g.Expect(*value).To(Not(Equal(currentClusterVpcId)))
+					g.Expect(*value).To(Not(Equal(CurrentClusterVpcId)))
 				}
 			}
 		}
@@ -137,7 +139,7 @@ func (env *Framework) ExpectToBeClean(ctx context.Context) {
 				Logger(ctx).Infof("Found Tags for service %v tags: %v", *service.Name, retrievedTags)
 				value, ok := retrievedTags.Tags[lattice.K8SServiceOwnedByVPC]
 				if ok {
-					g.Expect(*value).To(Not(Equal(currentClusterVpcId)))
+					g.Expect(*value).To(Not(Equal(CurrentClusterVpcId)))
 				}
 			}
 		}
@@ -145,8 +147,8 @@ func (env *Framework) ExpectToBeClean(ctx context.Context) {
 		retrievedTargetGroups, _ := env.LatticeClient.ListTargetGroupsAsList(ctx, &vpclattice.ListTargetGroupsInput{})
 		for _, tg := range retrievedTargetGroups {
 			Logger(ctx).Infof("Found TargetGroup: %s, checking it whether it's created by current EKS Cluster", *tg.Id)
-			if tg.VpcIdentifier != nil && currentClusterVpcId != *tg.VpcIdentifier {
-				Logger(ctx).Infof("Target group VPC Id: %s, does not match current EKS Cluster VPC Id: %s", *tg.VpcIdentifier, currentClusterVpcId)
+			if tg.VpcIdentifier != nil && CurrentClusterVpcId != *tg.VpcIdentifier {
+				Logger(ctx).Infof("Target group VPC Id: %s, does not match current EKS Cluster VPC Id: %s", *tg.VpcIdentifier, CurrentClusterVpcId)
 				//This tg is not created by current EKS Cluster, skip it
 				continue
 			}
@@ -175,14 +177,16 @@ func (env *Framework) CleanTestEnvironment(ctx context.Context) {
 	// Kubernetes API Objects
 	namespaces := &v1.NamespaceList{}
 	Expect(env.List(ctx, namespaces)).WithOffset(1).To(Succeed())
+	// TODO: instead of using one big list TestCasesCreatedK8sResource to track all created k8s resource,
+	//  we should create different lists for different kind of k8s resource i.e., httproute have 1 list, service have another list etc.
 	for _, object := range env.TestCasesCreatedK8sResource {
 		Logger(ctx).Infof("Deleting k8s resource %s %s/%s", reflect.TypeOf(object), object.GetNamespace(), object.GetName())
 		env.Delete(ctx, object)
 		//Ignore resource-not-found error here, as the test case logic itself could already clear the resources
 	}
 
-	//Theoretically, Deleting all k8s resource by `env.ExpectDeleteAllToSucceed()`, will make controller delete all related VPC Lattice resource,
-	//but the controller is still developing in the progress and may leaking some vPCLattice resource, need to invoke vpcLattice api to double confirm and delete leaking resource.
+	//Theoretically, Deleting all k8s resource by above `env.Delete(ctx, object)`, will make controller delete all related VPC Lattice resource,
+	//but the controller is still developing in the progress and may leaking some VPCLattice resource, need to invoke vpcLattice api to double confirm and delete leaking resource.
 	env.DeleteAllFrameworkTracedServiceNetworks(ctx)
 	env.DeleteAllFrameworkTracedVpcLatticeServices(ctx)
 	env.DeleteAllFrameworkTracedTargetGroups(ctx)
@@ -285,6 +289,7 @@ func (env *Framework) GetTargetGroup(ctx context.Context, service *v1.Service) *
 	return found
 }
 
+// TODO: Create a new function that only verifying deployment len(podList.Items)==*deployment.Spec.Replicas, and don't do lattice.ListTargets() api call
 func (env *Framework) GetTargets(ctx context.Context, targetGroup *vpclattice.TargetGroupSummary, deployment *appsv1.Deployment) []*vpclattice.TargetSummary {
 	var found []*vpclattice.TargetSummary
 	Eventually(func(g Gomega) {
@@ -302,7 +307,6 @@ func (env *Framework) GetTargets(ctx context.Context, targetGroup *vpclattice.Ta
 		retrievedTargets, err := env.LatticeClient.ListTargetsAsList(ctx, &vpclattice.ListTargetsInput{TargetGroupIdentifier: targetGroup.Id})
 		g.Expect(err).To(BeNil())
 		g.Expect(retrievedTargets).To(HaveLen(int(*deployment.Spec.Replicas)))
-
 		podIps := lo.Map(podList.Items, func(pod v1.Pod, _ int) string { return pod.Status.PodIP })
 		targetIps := lo.Filter(retrievedTargets, func(target *vpclattice.TargetSummary, _ int) bool {
 			return lo.Contains(podIps, *target.Id) &&
@@ -315,8 +319,42 @@ func (env *Framework) GetTargets(ctx context.Context, targetGroup *vpclattice.Ta
 	return found
 }
 
+func (env *Framework) IsVpcAssociatedWithServiceNetwork(ctx context.Context, vpcId string, serviceNetwork *vpclattice.ServiceNetworkSummary) (bool, error) {
+	Logger(ctx).Infof("IsVpcAssociatedWithServiceNetwork vpcId:%v serviceNetwork: %v \n", vpcId, serviceNetwork)
+	vpcAssociations, err := env.LatticeClient.ListServiceNetworkVpcAssociationsAsList(ctx, &vpclattice.ListServiceNetworkVpcAssociationsInput{
+		ServiceNetworkIdentifier: serviceNetwork.Id,
+		VpcIdentifier:            &vpcId,
+	})
+	if err != nil {
+		return false, err
+	}
+	if len(vpcAssociations) != 1 {
+		return false, fmt.Errorf("Expect to have one VpcServiceNetworkAssociation len(vpcAssociations): %d", len(vpcAssociations))
+	}
+	association := vpcAssociations[0]
+	if *association.Status != vpclattice.ServiceNetworkVpcAssociationStatusActive {
+		return false, fmt.Errorf("Current cluster should have one Active status association *association.Status: %s, err: %w", *association.Status, err)
+	}
+	return true, nil
+}
+
+func (env *Framework) AreAllLatticeTargetsHealthy(ctx context.Context, tg *vpclattice.TargetGroupSummary) (bool, error) {
+	Logger(ctx).Infof("Checking whether AreAllLatticeTargetsHealthy for targetGroup: %v", tg)
+	targets, err := env.LatticeClient.ListTargetsAsList(ctx, &vpclattice.ListTargetsInput{TargetGroupIdentifier: tg.Id})
+	if err != nil {
+		return false, err
+	}
+	for _, target := range targets {
+		Logger(ctx).Infof("Checking target: %v", target)
+		if *target.Status != vpclattice.TargetStatusHealthy {
+			return false, nil
+		}
+	}
+	return true, nil
+}
+
 func (env *Framework) DeleteAllFrameworkTracedServiceNetworks(ctx aws.Context) {
-	log.Println("DeleteAllFrameworkTracedServiceNetworks ", env.TestCasesCreatedServiceNetworkNames)
+	Logger(ctx).Infof("DeleteAllFrameworkTracedServiceNetworks %v", env.TestCasesCreatedServiceNetworkNames)
 	sns, err := env.LatticeClient.ListServiceNetworksAsList(ctx, &vpclattice.ListServiceNetworksInput{})
 	Expect(err).ToNot(HaveOccurred())
 	filteredSns := lo.Filter(sns, func(sn *vpclattice.ServiceNetworkSummary, _ int) bool {
--- a/test/suites/integration/deregister_targets_test.go
+++ b/test/suites/integration/deregister_targets_test.go
@@ -1,20 +1,16 @@
 package integration
 
 import (
-	"log"
-	"os"
-	"time"
-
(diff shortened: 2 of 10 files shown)


test/pkg/test/framework.go
```diff
@@ -86,11 +88,8 @@ func NewFramework(ctx context.Context) *Framework {
 		TestCasesCreatedServiceNames:        make(map[string]bool),
 		TestCasesCreatedTargetGroupNames:    make(map[string]bool),
 	}
-
-	SetDefaultEventuallyTimeout(180 * time.Second)
+	SetDefaultEventuallyTimeout(5 * time.Minute)
 	SetDefaultEventuallyPollingInterval(10 * time.Second)
-	BeforeEach(func() { framework.ExpectToBeClean(ctx) })
-	AfterEach(func() { framework.ExpectToBeClean(ctx) })
 	return framework
```
zijun726911, line 93: Moved  `BeforeEach()` and `AfterEach()` to the `test/suites/integration/suite_test.go`, will re-considering whether we really need it later.

test/pkg/test/framework.go
```diff
@@ -102,13 +101,12 @@ func (env *Framework) ExpectToBeClean(ctx context.Context) {
 		env.EventuallyExpectNoneFound(ctx, testObject.ListType)
 	})
 
-	currentClusterVpcId := os.Getenv("CLUSTER_VPC_ID")
```
zijun726911, line 105: Just moved that block in the `Eventually()`

test/suites/integration/deregister_targets_test.go
```diff
@@ -78,21 +75,42 @@ var _ = Describe("Deregister Targets", func() {
 		)
 	})
 
-	It("Kubernetes Service deletion deregisters targets", func() {
+	It("Kubernetes Service deletion triggers targets de-registering", func() {
+		Skip("Currently controller have a bug, service deletion do NOT triggers targets de-registering, need to further investigate the root cause")
 		testFramework.ExpectDeleted(ctx, service)
```
zijun726911, line 80: After fixed the obvious mistake for this test case (by adding the `.Should(Succeed())`), I found currently controller don't have this behaviour "Kubernetes Service deletion triggers targets de-registering"

I also did manual test several times and could confirm that is a bug.

I think we should temprarily skip() this test case and investigate the root cause in another issue:
https://github.com/aws/aws-application-networking-k8s/issues/331

test/suites/integration/deregister_targets_test.go
```diff
@@ -78,21 +75,42 @@ var _ = Describe("Deregister Targets", func() {
...
+			Expect(err).To(BeNil())
+			log.Println("targets:", targets)
+			if len(targets) == 0 {
+				g.Expect(true).To(BeTrue())
+			} else {
+				for _, target := range targets {
+					g.Expect(*target.Status).To(Equal(vpclattice.TargetStatusDraining))
+				}
+			}
+		}).Should(Succeed())
```
zijun726911, line 114: we also regard `TargetStatusDraining` as this test pass, so don't need to wait for `.WithTimeout(5*time.Minute + 10*time.Second)`

test/suites/integration/httproute_path_match_test.go
```diff
@@ -38,39 +37,33 @@ var _ = Describe("HTTPRoute path matches", func() {
...
-		for _, target := range targetsV2 {
-			Expect(*target.Port).To(BeEquivalentTo(service2.Spec.Ports[0].TargetPort.IntVal))
-			Expect(*target.Status).To(Or(
-				Equal(vpclattice.TargetStatusInitial),
-				Equal(vpclattice.TargetStatusHealthy),
-			))
-		}
-
-		Eventually(func(g Gomega) {
+		deployments := []*appsv1.Deployment{deployment1, deployment2}
```
zijun726911, line 40: Just remove the time.Sleep(3 * time.Minute) and move some verification logic in the `Eventually()` block

test/suites/integration/https_listener_weighted_rule_with_service_export_import_test.go
```diff
@@ -99,55 +108,56 @@ var _ = Describe("Test 2 listeners gateway with weighted httproute rules and ser
...
-			dnsName := testFramework.GetVpcLatticeServiceDns(httpRoute.Name, httpRoute.Namespace)
-
-			pods := testFramework.GetPodsByDeploymentName(deployment0.Name, deployment0.Namespace)
-			Expect(len(pods)).To(BeEquivalentTo(1))
-			log.Println("client pod name:", pods[0].Name)
-			protocols := []string{"http", "https"}
-			for _, protocol := range protocols {
+			Eventually(func(g Gomega) { // Put traffic verification logic in the Eventually() block, because connectivity setup may need some time to fully propagate to vpc lattice dataplane
+				log.Println("Verifying Weighted rule traffic")
+				dnsName := testFramework.GetVpcLatticeServiceDns(httpRoute.Name, httpRoute.Namespace)
```
zijun726911, line 127: Just move the verification logic in the `Eventually()` block, verification logic itself didn't change

mikhail-aws:
whats the improvement on actual e2e tests time? Last time I ran 11 tests it took me 4966.49s
```
--- PASS: TestIntegration (4966.49s)
```

test/pkg/test/framework.go
```diff
@@ -86,11 +88,8 @@ func NewFramework(ctx context.Context) *Framework {
 		TestCasesCreatedServiceNames:        make(map[string]bool),
 		TestCasesCreatedTargetGroupNames:    make(map[string]bool),
 	}
-
-	SetDefaultEventuallyTimeout(180 * time.Second)
+	SetDefaultEventuallyTimeout(5 * time.Minute)
```
mikhail-aws, line 91: not following, I dont know why it was 3 minutes before and now it should be 5, some magic numbers

test/pkg/test/framework.go
mikhail-aws: whats the problem of having it here?

test/pkg/test/framework.go
```diff
@@ -102,13 +101,12 @@ func (env *Framework) ExpectToBeClean(ctx context.Context) {
...
-		VpcIdentifier: aws.String(currentClusterVpcId),
-	})
-	Logger(ctx).Infof("Expect VPC used by current cluster don't have any ServiceNetworkVPCAssociation, if it has you should manually delete it")
-	Expect(len(retrievedServiceNetworkVpcAssociations)).To(Equal(0))
 	Eventually(func(g Gomega) {
+		retrievedServiceNetworkVpcAssociations, _ := env.LatticeClient.ListServiceNetworkVpcAssociationsAsList(ctx, &vpclattice.ListServiceNetworkVpcAssociationsInput{
+			VpcIdentifier: aws.String(CurrentClusterVpcId),
+		})
+		Logger(ctx).Infof("Expect VPC used by current cluster don't have any ServiceNetworkVPCAssociation, if it has you should manually delete it")
+		g.Expect(len(retrievedServiceNetworkVpcAssociations)).To(Equal(0))
```
mikhail-aws, line 109: Not sure if it's improvement, I dont see any good of making Eventually assertions even more bigger. We should get away from these gigantic Eventually blocks, it does not looks idiomatic from ginkgo framework. I see that we use eventually as a big retry on error block, that's not exact purpose of Eventually

test/suites/integration/deregister_targets_test.go
mikhail-aws: I dont think we should skip test because of bug, we need to address bug. Let it fail until fixed.

test/suites/integration/deregister_targets_test.go
```diff
@@ -78,21 +75,42 @@ var _ = Describe("Deregister Targets", func() {
...
-			targets := testFramework.GetTargets(ctx, targetGroup, deployment)
-			Expect(len(targets) == 0)
-		}).WithTimeout(5*time.Minute + 10*time.Second)
+			targets, err := testFramework.LatticeClient.ListTargetsAsList(ctx, &vpclattice.ListTargetsInput{
+				TargetGroupIdentifier: targetGroup.Id,
+			})
+			Expect(err).To(BeNil())
+			log.Println("targets:", targets)
+			if len(targets) == 0 {
+				g.Expect(true).To(BeTrue())
```
mikhail-aws, line 89: true to be true... please use correct assertion on list len

test/suites/integration/deregister_targets_test.go
mikhail-aws: this code looks identical to code above at line 83, probably deserve own function

test/suites/integration/httproute_creation_test.go
```diff
@@ -101,12 +99,12 @@ func verifyResourceCreation(
...
-
-	vpcLatticeService = testFramework.GetVpcLatticeService(ctx, httpRoute)
-	Expect(*vpcLatticeService.DnsEntry).To(ContainSubstring(latticestore.LatticeServiceName(httpRoute.Name, httpRoute.Namespace)))
+	Eventually(func(g Gomega) {
+		vpcLatticeService = testFramework.GetVpcLatticeService(ctx, httpRoute)
+		g.Expect(*vpcLatticeService.DnsEntry).To(ContainSubstring(latticestore.LatticeServiceName(httpRoute.Name, httpRoute.Namespace)))
+		targetGroup = testFramework.GetTargetGroup(ctx, service)
+		g.Expect(*targetGroup.VpcIdentifier).To(Equal(os.Getenv("CLUSTER_VPC_ID")))
+		g.Expect(*targetGroup.Protocol).To(Equal("HTTP"))
+	})
```
mikhail-aws, line 108: why replaced with Eventually?

mikhail-aws (commented):
I have a mixed impression about this PR. There are good parts for typos and handy methods.

But overall "improvement" direction does not look to me as improvement. Our tests still overloaded, so many assertions happen in single test, Eventually blocks everywhere which looks to me as bandaid on API calls that fail and we need retry-wait. It's hard to read these tests.

I cant tell that new code looks more readable to me. Same code blocks reshuffled. Not sure how to move forward with it. Probably split into pieces.

test/suites/integration/deregister_targets_test.go
zijun726911: But before we fix it it always leave a failed test cases? and other developer could be confuse on whether their code change break something but actually not? Also if we leave it not skip(), in the Eventually() block it found the targets still existing over and over again, and finally failed on Eventually() block timeout, it can take 3 min, I think this 3 min wait time is meaningless

test/suites/integration/deregister_targets_test.go
zijun726911: good point, will change

zijun726911:
> I have a mixed impression about this PR. There are good parts for typos and handy methods.
> 
> But overall "improvement" direction does not look to me as improvement. Our tests still overloaded, so many assertions happen in single test, Eventually blocks everywhere which looks to me as bandaid on API calls that fail and we need retry-wait. It's hard to read these tests.
> 
> I cant tell that new code looks more readable to me. Same code blocks reshuffled. Not sure how to move forward with it. Probably split into pieces.
> 
> I would rather refactor single test, drop it into smaller pieces, clean assertions (just a few per "It"). Something like Ryan did https://github.com/aws/aws-application-networking-k8s/blob/main/test/suites/integration/httproute_creation_test.go. Easy to read and comprehend.


`Eventually blocks everywhere which looks to me as bandaid on API calls that fail and we need retry-wait.` The api call didn't failed. It's intention is: instead of wait for 3min then check lattice resource once, it  sleep for some shorter interval and check lattice resource creation more frequently. The k8s controller follow the design pattern of  "reconciliation" and "eventually state", so "eventually" all lattice resource could be created. Usually the controller don't need 3min (180s) to create all lattice resource, maybe it just need 10sec sleep interval * 10 time retry == 100 sec ==1.6 min in eventually block to create all lattice. 3min compare to 1.6 min may not be a big time saver, considering the whole test suite total time. but it indeed same some time.   

Some big time savers I could come up with:
- Combine httproute_header_match_test.go to httproute_path_match_test.go into one file and use one gateway and one httproute, use one It() to verify lattice resource creation, one `It()` to verify lattice resource, one `It()` to verify header_match traffic, one `It()` to verify path_match traffic, and when Eric F's code merge, use another `It()` to verify method_match traffic
- Combine httproute_creation_test.go, httproute_mutation_do_not_leak_target_group_test.go, httproute_update_test.go to reuse one gateway
- Only do wait for draining target one time in the AfterSuite()

Above improvement seems quite big, don't plan to add them in the current PR.



Some ideas I can come up with to improve current PR:
- Option1
    - Fix some obvious test cases mistake 
    - Use Eventually() for deregister_targets_test.go and srvexport_port_annotation_targets_test.go and httproute_creation_test.go to wait for all lattice reosurce created
    - For test cases that test traffic, revert other traffic test change and only refactor `httproute_path_match_test.go` to use big Eventually block
 
- Option2
    - Fix some obvious test cases mistake 
    - Use Eventually() for deregister_targets_test.go and srvexport_port_annotation_targets_test.go and httproute_creation_test.go to wait for all lattice reosurce created
    - For test cases that test traffic, revert other traffic test change and only refactor `httproute_path_match_test.go` to use some small Eventually block to only verify TargetGroup healthy, vpcAssociation

- Option3
    - Fix some obvious test cases mistake 
    - Use Eventually() for deregister_targets_test.go and srvexport_port_annotation_targets_test.go and httproute_creation_test.go to wait for all lattice reosurce created
    - For test cases that test traffic, revert all traffic test change (improve traffic related test when later combine httproute_header_match_test.go to httproute_path_match_test.go), current Eventually() idea indeed is not the biggest time saver

Or, do you have some other ideas?

test/suites/integration/deregister_targets_test.go
zijun726911: I think  we need assert a true in the `if len(targets) == 0 { )` to make test logic more clear, I can do:
```
if len(targets) == 0 {
			g.Expect(len(targets)).To(Equal(0))
		}
```

test/suites/integration/httproute_creation_test.go
zijun726911: The intention is: instead of wait for 3min then check lattice resource only once, it  sleep for some shorter interval and check lattice resource creation more frequently. The k8s controller follow the design pattern of  "reconciliation" and "eventually state", so "eventually" all lattice resource could be created. Usually the controller don't need 3min (180s) to create all lattice resource, maybe it just need 10sec sleep interval * 10 time retry == 100 sec ==1.6 min in eventually block to verify all lattice resources have been created.

zijun726911:
Another point is, if we leave the `time.Sleep(3 * time.Minute)` e2e code example, when other people write new e2e test case their copy and paste that example and make the e2etest time keep bloating

(That might not be a really important point if we adopt some other big time savers in e2etest I describe above)

test/suites/integration/deregister_targets_test.go
mikhail-aws: Then we should fix it as first priority, disabling tests because of bugs is not a good idea.

test/suites/integration/deregister_targets_test.go
mikhail-aws: or just why single does not work? `g.Expect(len(targets)).To(Equal(0))`

mikhail-aws:
> Some big time savers I could come up with:
Sounds good to me.

> Option2
lets go for it

We need to brainstorm a bit how do we handle async resource creation - the "Create-Wait-Till-Created" loop. I'm thinking that we should handle it in the source code. For example ddb table API has a [wait-exists](https://docs.aws.amazon.com/cli/latest/reference/dynamodb/wait/table-exists.html) API call, which is part of SDK, we might need something like this, may be not part of sdk, but implement in source of gateway-api. This approach should help with removing eventually blocks and LATTICE_RETRY for resources in progress. It will be a normal sync call.

mikhail-aws:
> Another point is, if we leave the time.Sleep(3 * time.Minute) e2e code example, when other people write new e2e test case they copy and paste that example and make the e2etest time keep bloating

This assumption is true, if we keep writing large Eventually blocks, which I dont think we should. Smaller eventually blocks for 1-2 API calls should not wait for 5 minutes.

mikhail-aws:
Here is [example](https://github.com/projectcontour/contour/blob/main/test/e2e/framework.go#L316) from contour, they use public library from k8s [apimachinery](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/wait/poll.go#L110). They dont use eventually from ginkgo.

test/suites/integration/deregister_targets_test.go
zijun726911: Because it need a OR condition:
- if len(targets) == 0 regard this test case pass,
- Or if len(targets) != 0 && all targets are in Draining status we regard this test case pass
I think it's hard to write one Gomega assertion to express that

test/suites/integration/httproute_creation_test.go
solmonk: don't we need Should(Succeed()) here?

solmonk: Anyways I'm also not a fan of `Eventually(func).Should(Succeed())` approach, I would like to have assertions outside the func whenever possible.

If it is hard to do so, I think this kind of `Eventually()` function should be split. We are testing two separate calls here - getting lattice service and target group. It reduces the need of having complex logic inside Eventually() so that we may be able to pull out some.

solmonk:
Some thoughts:

* I'm personally fine with spamming `Eventually()`, what I don't prefer is `Eventually()` clause being too large and having to many assertions inside it. Ideally I would expect `Eventually()` act as a simple data retriever, and assertions happen outside. 

> We need to brainstorm a bit how do we handle async resource creation - the "Create-Wait-Till-Created" loop. I'm thinking that we should handle it in the source code. For example ddb table API has a [wait-exists](https://docs.aws.amazon.com/cli/latest/reference/dynamodb/wait/table-exists.html) API call, which is part of SDK, we might need something like this, may be not part of sdk, but implement in source of gateway-api. This approach should help with removing eventually blocks and LATTICE_RETRY for resources in progress. It will be a normal sync call.

* I would be more careful about this. Each step of resource creation (CREATE_IN_PROGRESS, CREATED) leaves a thing to do in controller - updating resource status, etc. Sync-waiting doesn't seem ideal either for me.

* I would suggest to avoid increasing the scope of PR too much. We know there are whole bunch of rooms for improvements in this codebase but we should try to deal with them separately. If the PR gets rid of fixed wait time I think it serves its purpose. I would be fine with this PR if Eventually() is just a bit more fine-grained.

test/suites/integration/deregister_targets_test.go
zijun726911: I think we could use `Fail()`

test/pkg/test/framework.go
```diff
@@ -315,8 +315,42 @@ func (env *Framework) GetTargets(ctx context.Context, targetGroup *vpclattice.Ta
...
+func (env *Framework) AreAllLatticeTargetsHealthy(ctx context.Context, tg *vpclattice.TargetGroupSummary) (bool, error) {
+	Logger(ctx).Infof("Checking whether AreAllLatticeTargetsHealthy for targetGroup: %v", tg)
+	targets, err := env.LatticeClient.ListTargetsAsList(ctx, &vpclattice.ListTargetsInput{TargetGroupIdentifier: tg.Id})
+	if err != nil {
+		return false, err
+	}
+	for _, target := range targets {
+		Logger(ctx).Infof("Checking target: %v", target)
+		if *target.Status != vpclattice.TargetStatusHealthy {
+			return false, fmt.Errorf("Target is not healthy yet, target: %v", target)
```
mikhail-aws, line 346: I think err should be nil here, since call is successful but target is not healthy yet. If you still want to return error here, then you can remove bool, since True happens only when there are no errors, which means err==nil

test/suites/integration/deregister_targets_test.go
```diff
@@ -78,21 +73,32 @@ var _ = Describe("Deregister Targets", func() {
...
+		})
+		g.Expect(err).To(BeNil())
+		log.Println("targets:", targets)
+		if len(targets) == 0 {
+			g.Expect(len(targets)).To(Equal(0))
+		} else {
+			for _, target := range targets {
+				g.Expect(*target.Status).To(Equal(vpclattice.TargetStatusDraining))
+			}
+		}
```
mikhail-aws, line 102: `for _, target := range` already covers targets with len=0, you dont need if/else here

test/suites/integration/suite_test.go
```diff
@@ -3,11 +3,11 @@ package integration
 import (
 	"context"
 	"os"
-	"testing"
 
 	"github.com/aws/aws-application-networking-k8s/test/pkg/test"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"testing"
```
mikhail-aws, line 10: interesting that formater you are using mixing up imports order, usually it's std lib, then public packages, then local packages, you might need to adjust IDE setting for go formatting

mikhail-aws:
>     * I would suggest to avoid increasing the scope of PR too much. We know there are whole bunch of rooms for improvements in this codebase but we should try to deal with them separately. If the PR gets rid of fixed wait time I think it serves its purpose. I would be fine with this PR if Eventually() is just a bit more fine-grained.

Yes, I commented that we should reduce scope of this PR. But brainstorm for future how do we handle async resources creation in controller and tests. I'm not exactly sure that returning "LATTICE_RETRY" error and re-queue request is better than just wait for resource creation in a sync call. Of course it's outside of scope of this PR.

test/suites/integration/suite_test.go
```diff
@@ -16,5 +15,8 @@ func TestIntegration(t *testing.T) {
 	ctx = test.NewContext(t)
 	testFramework = test.NewFramework(ctx)
 	RegisterFailHandler(Fail)
+	//TODO: re-consider whether we really need ExpectToBeClean() in BeforeEach/AfterEach
+	BeforeEach(func() { testFramework.ExpectToBeClean(ctx) })
+	AfterEach(func() { testFramework.ExpectToBeClean(ctx) })
```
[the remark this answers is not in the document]
zijun726911: now I move it back to the framework.go, planing to remove it later

test/pkg/test/framework.go
zijun726911: changed

test/suites/integration/deregister_targets_test.go
```diff
@@ -68,7 +63,11 @@ var _ = Describe("Deregister Targets", func() {
 	})
 
 	AfterEach(func() {
-		testFramework.CleanTestEnvironment(ctx)
+		// Lattice targets draining time for test cases in this file is actually unavoidable,
+		//because these test cases logic themselves test lattice targets de-registering before delete the httproute
+		testFramework.ExpectDeleted(ctx, gateway, pathMatchHttpRoute)
+		time.Sleep(30 * time.Second) // Use a trick to delete httpRoute first and then delete the service and deployment to avoid draining lattice targets
+		testFramework.ExpectDeleted(ctx, service, deployment)
```
mikhail-aws, line 70: Can you explain (probably again) why do you need to use trick for 30 seconds wait? And whats the deal to not use trick?

test/suites/integration/deregister_targets_test.go
mikhail-aws: Looking again into this problem, why dont we focus on fixing this bug and come back to this PR after?

mikhail-aws: I think bug fixing and e2e test changes should come together, not in this PR

mikhail-aws (approved):

zijun726911:
I just plan to update... Current running spend about 2500 sec

Ran 11 of 11 Specs in 2572.128 seconds
--- PASS: TestIntegration (2572.46s)
PASS

Ran 11 of 11 Specs in 2321.826 seconds
--- PASS: TestIntegration (2322.22s)

test/suites/integration/deregister_targets_test.go
zijun726911: Some backgroud: If a latticeService's listener's rule can route to that targetGroup, targets in that tg will have `healthy`/`unhealthy` status, derigister a healthy status target can cause  5 min draining. if no rule can route to that targetGroup, the target would be `unused` status, we coud immediately de-register it without draining time.


Some other backgroud: when delete a httproute, it following this order to delete : latticeService -> listener -> rules -> targets -> target group that is pretty streamline, target status chagne to  `unused` first, and then could be immediately deleted.

However, if we try to delete a k8sDeplotment or k8sService prior than deleting a httproute, controller will try to de-register a `healthy` status target, that cause draining status targets

mikhail-aws: interesting, should we change controller's logic to removing route prior to registration?

Issue #123, opened 2023-03-12
Need to support HTTP method-based routing


liwenwu-amazon:
Need to support HTTP method-based routing

erikfuller:
This should be resolved

PR #332, opened 2023-08-16
Http method routing

erikfuller:
**What type of PR is this?**

Feature: Adds the ability to route requests by HTTP method. Also fixes a race condition I came across in target group deletion where attempts to list targets for an already-deleted target group cause deletion to fail. I've organized the HTTP method changes and the race condition changes into two separate commits.

Addresses https://github.com/aws/aws-application-networking-k8s/issues/123

Added unit test and integration test. Ran e2e tests. Ran several manual tests manually editing and pushing YAML files.

**Automation added to e2e**:

New integration tests in ```httproute_method_match_test.go```

```
Ran 12 of 12 Specs in 5194.655 seconds
SUCCESS! -- 12 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestIntegration (5195.10s)
PASS
ok  	github.com/aws/aws-application-networking-k8s/test/suites/integration	5195.788s
```

No

Downgrades would reject any routes that had been created with a method match. These routes would need to be updated to remove the method match.

```release-note
HTTP Method-based routing supported
```


pkg/aws/services/vpclattice_mocks.go
erikfuller, line 1: hmm, I'm not actually sure this should be here. Let me double-check what's up here.

erikfuller: Not sure when this popped up but wasn't meant to be part of this PR. Looks fixed now.

pkg/deploy/lattice/rule_manager.go
```diff
@@ -360,8 +359,13 @@ func isRulesSame(modelRule *latticemodel.Rule, sdkRuleDetail *vpclattice.GetRule
 		}
 	}
 
-	// Header Match
+	// Method match
+	if aws.StringValue(sdkRuleDetail.Match.HttpMatch.Method) != modelRule.Spec.Method {
+		glog.V(6).Infof("Method mismatch '%v' != '%v'\n", modelRule.Spec.Method, sdkRuleDetail.Match.HttpMatch.Method)
```
solmonk, line 364: The way of logging will change pretty soon anyways but for the note, trailing `\n` is not required - also I would recommend avoiding use of `%v` if possible.

solmonk (commented):
This looks good to me in general, @xWink can you check if this is conflicting with any of your ongoing work?

pkg/deploy/lattice/rule_manager.go
erikfuller: So far I've tried to stick with whatever convention was in the file. Happy to follow up on these as part of the ongoing logging improvements.

liwenwu-amazon:
Are u going to add more logic in these routines?  
https://github.com/aws/aws-application-networking-k8s/blob/main/pkg/deploy/lattice/rule_manager.go#L313

erikfuller:
> Are u going to add more logic in these routines? https://github.com/aws/aws-application-networking-k8s/blob/main/pkg/deploy/lattice/rule_manager.go#L313

Yes, it's in the diff.
https://github.com/aws/aws-application-networking-k8s/pull/332/files#diff-c0535a89c8f62832ed3ff41ab0ce85347947a611a3422d61c9f4e1cc340f15f7

pkg/deploy/lattice/target_group_manager.go
```diff
@@ -163,54 +163,64 @@ func (s *defaultTargetGroupManager) Delete(ctx context.Context, targetGroup *lat
...
-			return errors.New(LATTICE_RETRY)
+	} else {
+		// deregister targets
+		var targets []*vpclattice.Target
+		for _, t := range listResp {
+			if t.Status != nil && *t.Status != vpclattice.TargetStatusUnused {
+				glog.V(6).Infof("TargetGroupManager: The targetGroup [%v] has non-Unused status target(s), which means this targetGroup is still in use by latticeService, it cannot be deleted now\n", targetGroup.Spec.LatticeID)
+				// Before call the defaultTargetGroupManager.Delete(), we always call the latticeServiceManager.Delete() first,
+				//  *t.Status != vpclattice.TargetStatusUnused means previous delete latticeService still in the progress, we could wait for 20 seconds and then retry
+				return errors.New(LATTICE_RETRY)
```
mikhail-aws, line 185: I saw a comment on follow up on logging improvement. It would be nice to have retry error with original error. There are no examples yet in the source code, but here is concept I want to use. So we dont need to do logging before returning retry.

```go
var LATTICE_RETRY = errors.New("lattice_retry")

// ... later in code
err := fmt.Errorf("%w, target status is 'unused', previous delete latticeService still in progress", LATTICE_RETRY)

// the %w will be recognized by retry error handler, for example this call will return true, even if LATTICE_RETRY in the middle of errors chain
errors.Is(err, LATTICE_RETRY) == true

```

pkg/deploy/lattice/rule_manager.go
```diff
@@ -360,8 +359,13 @@ func isRulesSame(modelRule *latticemodel.Rule, sdkRuleDetail *vpclattice.GetRule
 		}
 	}
 
-	// Header Match
+	// Method match
```
mikhail-aws, line 362: Any thoughts on current structure of validations? we just keep adding them with more if blocks?
Does it make sense to convert sdkRuleDetail into latticemodel.Rule and do something like reflect.deepEqual? https://pkg.go.dev/reflect#DeepEqual

pkg/deploy/lattice/target_group_manager.go
erikfuller: This pattern is pretty pervasive and I'd be reluctant to block this PR on that improvement, but happy to follow up separately with this change.

test/suites/integration/httproute_method_match_test.go
```diff
@@ -0,0 +1,124 @@
...
+		testFramework.ExpectCreated(ctx,
+			gateway,
+			methodMatchHttpRoute,
+			getService,
+			deployment1,
+			postService,
+			deployment2,
+		)
+
+		time.Sleep(3 * time.Minute)
```
xWink, line 33: Why do we sleep here?

test/suites/integration/suite_test.go
```diff
@@ -12,6 +13,13 @@ import (
 var testFramework *test.Framework
 var ctx context.Context
 
+var _ = BeforeSuite(func() {
+	vpcid := os.Getenv("CLUSTER_VPC_ID")
+	if vpcid == "" {
+		Fail("CLUSTER_VPC_ID environment variable must be set to run integration tests")
```
xWink, line 19: Thank you for adding this!

xWink (approved):
Thank you for the bug fixes. HTTP method routing implementation LGTM. Just left one non-blocking question on why wait during a test!

test/suites/integration/httproute_method_match_test.go
erikfuller: Every routing test does this, I'm just following the crowd. Agree this isn't ideal and shouldn't be our approach as it could artificially extend the time the test takes to run.

erikfuller:
Thanks for all the comments, folks.

Issue #336, opened 2023-08-17
controller_config_test.go fails when run with access to Instance Metadata Service


klyubin:
`pkg/config/controller_config.go` falls back to Instance Metadata Service when the AWS_ACCOUNT_ID, CLUSTER_VPC_ID, or REGION env vars are not set. Its unit test suite -- `pkg/config/controller_config_test.go` -- assumes that it runs in environment where Instance Metadata Service is not accessible. This makes it fail in environments where Instance Metadata Service is accessible:
```bash
--- FAIL: Test_config_init_with_partial_env_var (0.00s)
    controller_config_test.go:22: 
        	Error Trace:	/workspace/pkg/config/controller_config_test.go:22
        	Error:      	Expected value not to be nil.
        	Test:       	Test_config_init_with_partial_env_var
--- FAIL: Test_config_init_no_env_var (0.00s)
    controller_config_test.go:32: 
        	Error Trace:	/workspace/pkg/config/controller_config_test.go:32
        	Error:      	Expected value not to be nil.
        	Test:       	Test_config_init_no_env_var
FAIL
FAIL	github.com/aws/aws-application-networking-k8s/pkg/config	0.010s
```

klyubin:
Fixed in https://github.com/aws/aws-application-networking-k8s/pull/338

PR #338, opened 2023-08-18
Make IMDS unavailable in controller config tests (#336)

klyubin:
The code under test falls back to reading AWS region, account ID, and VPC ID from Instance Metadata Service. The tests currently assume that this service is unavailable. This means they fail when run in environments where this service is available to tests.

This commit fixes this issue by making the tests use a mock Instance Metadata Service which returns an error. This ensures that these tests work as expected regardless of whether Instance Metadata Service is available/reachable.

**What type of PR is this?**

Fixes broken tests

pkg/config/controller_config_test.go fails in environments with access to AWS EC2 Instance Metadata Service.

Unbreaks tests in environments with IMDS accessible.

`go test ./...`

**Automation added to e2e**:

No need

No
```release-note
pkg/config tests now pass in environments which have access to AWS Instance Metadata Service.
```

### DIFF
--- a/pkg/config/controller_config.go
+++ b/pkg/config/controller_config.go
@@ -41,6 +41,10 @@ func GetClusterLocalGateway() (string, error) {
 func ConfigInit() error {
 	sess, _ := session.NewSession()
 	metadata := NewEC2Metadata(sess)
+	return configInit(metadata)
+}
+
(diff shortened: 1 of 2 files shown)


zijun726911 (approved):
LGTM

Issue #25, opened 2022-11-11
Support gateway GRPCroute object


liwenwu-amazon:
Add support for gateway API [gRPC routing](https://gateway-api.sigs.k8s.io/guides/grpc-routing/)

PR #341, opened 2023-08-18
add gRPC route controller

xWink:
**What type of PR is this?**
feature

#25 (partial)

- Adds GRPCRoute CRD
- Replaces HTTPRoute Controller and EventHandler with a consolidated Route controller and event handler
    - This allows for all registered route types to be handled by the same controller and event handler, due to the abstraction of HTTP and GRPC routes
    - Creation, Update, and Deletion of GRPCRoutes in Kubernetes now results in the same process as that of an HTTPRoute
- Adds rule builder logic for GRPCRouteMatch which converts K8s route matches to Lattice rule matches
- Fixes example grpc-route.yaml so that it deploys a GRPCRoute instead of HTTPRoute
- Fixes example grpc-server.yaml so that it uses a working image and HTTPS port
- Adds `-v 6` flag when running in developer mode
- Adds Equals() method to many route abstraction constructs to enable comparisons without relying on DeepEquals(), which may fail due to reliance of reflection
- General cleanup (spacing, indentation, log migration from glog to zap, improved log contents)

1. `kubectl apply -f examples/my-hotel-gateway.yaml`
2. `kubectl apply -f examples/grpc-server.yaml`
3. `kubectl apply -f examples/grpc-route.yaml`

Should create VPC Lattice Service routing to GRPC server. If a service and method is specified in `grpc-route.yaml`, the rule using a PathMatchExact value of `/<service>/<method>` is expected. If a service is provided, but no method is provided, a PatchMatchPrefix value of `/<service>/` is expected. If neither a service, nor a method is provided, a PathMatchPrefix value of `/` is expected. If no service is provided, but a method is provided, this should fail to reconcile since Lattice doesn't support Regex or Suffix matches.

Additionally, deletion of these resources should succeed, with both Lattice and K8s resources being deleted.

Ran e2e tests using latest changes:
```
• [224.537 seconds]

Ran 11 of 11 Specs in 2631.797 seconds
```

And ran the manual steps mentioned above.

**Automation added to e2e**:

No, this is to be done separately.

No.

The new CRD must be deployed to the cluster. The following commands will do so:
```
kubectl apply -f config/crds/bases/k8s-gateway-v0.6.1.yaml
```

Yes, though the feature is not complete yet.
Remaining work in followup PR:
- Reconcile GRPCRoutes on Gateway deletion
- TargetGroupModelBuilder for Lattice GRPC Target Group
- TargetGroupSynthesizer checks for GRPCRoute alongside existing check for HTTPRoute
- Service EventHandler generification for all route types

```release-note
action required
Added support for creating, updating, and deleting GRPCRoutes.
```


cmd/aws-application-networking-k8s/main.go
```diff
@@ -151,6 +153,13 @@ func main() {
 		os.Exit(1)
 	}
 
+	grpcRouteReconciler := controllers.NewGrpcRouteReconciler(cloud, mgr.GetClient(), mgr.GetScheme(), mgr.GetEventRecorderFor("grpcroute"), finalizerManager, latticeDataStore)
+
+	if err = grpcRouteReconciler.SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "GRPCRoute")
+		os.Exit(1)
+	}
+
```
mikhail-aws, line 162: I did refactoring for other controllers, not http route yet, with single call controller registration. Can you follow that pattern please?

https://github.com/aws/aws-application-networking-k8s/blob/main/cmd/aws-application-networking-k8s/main.go#L142-L145C3

You also need to add logger argument and use it in controller, rather than glog

controllers/eventhandlers/grpcroute.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func NewEnqueueRequestGRPCRouteEvent(client client.Client) handler.EventHandler {
+	return &enqueueRequestsForGRPCRouteEvent{
+		client: client,
+	}
+}
+
+func (h *enqueueRequestsForGRPCRouteEvent) Create(e event.CreateEvent, queue workqueue.RateLimitingInterface) {
+	glog.V(6).Info("GRPCRoute create")
```
mikhail-aws, line 31: Be more specific with logging, add something about event you receive. Same for other logs without specifics.

controllers/eventhandlers/grpcroute.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (h *enqueueRequestsForGRPCRouteEvent) Update(e event.UpdateEvent, queue workqueue.RateLimitingInterface) {
+	glog.V(6).Info("GRPCRoute update ")
+
+	oldGRPCRoute := e.ObjectOld.(*gateway_api_v1alpha2.GRPCRoute)
+	newGRPCRoute := e.ObjectNew.(*gateway_api_v1alpha2.GRPCRoute)
+
+	if !equality.Semantic.DeepEqual(oldGRPCRoute.Spec, newGRPCRoute.Spec) {
+		glog.V(6).Infof("--oldGRPCRoute %v \n", oldGRPCRoute.Spec)
+		glog.V(6).Infof("--newGRPCRoute %v \n", newGRPCRoute.Spec)
```
mikhail-aws, line 45: single log statement please, dont dump full spec into log, no \n

controllers/eventhandlers/grpcroute.go
```diff
@@ -0,0 +1,89 @@
...
+		if newGRPCRoute.Status.RouteStatus.Parents != nil &&
+			newGRPCRoute.Status.RouteStatus.Parents[0].Conditions[0].LastTransitionTime != ZeroTransitionTime {
```
mikhail-aws, line 47: is it guarantied that Parents and Conditions has size of 1 at least? if yes please attach reference. If not you need to do a check for size

controllers/grpcroute_controller.go
```diff
@@ -0,0 +1,376 @@
...
+type GRPCRouteReconciler struct {
+	client           client.Client
+	scheme           *runtime.Scheme
+	finalizerManager k8s.FinalizerManager
+	eventRecorder    record.EventRecorder
+	modelBuilder     gateway.LatticeServiceBuilder
+	stackDeployer    deploy.StackDeployer
+	latticeDataStore *latticestore.LatticeDataStore
+	stackMashaller   deploy.StackMarshaller
+}
```
mikhail-aws, line 63: Can you reuse http reconciler, dependencies are same, you can add conditional check when reconcile event comes, then branch into http/grpc. 
You may find some methods in controller are generic to route handling, where new route abstraction can help.

pkg/gateway/model_build_lattice_service.go
```diff
@@ -69,15 +69,13 @@ func (t *latticeServiceModelBuildTask) buildModel(ctx context.Context) error {
 	err := t.buildLatticeService(ctx)
 
 	if err != nil {
-		glog.V(2).Infof("latticeServiceModelBuildTask: Failed on buildLatticeService %v\n ", err)
-		return err
+		return fmt.Errorf("latticeServiceModelBuildTask: Failed on buildLatticeService %+v", err)
```
mikhail-aws, line 72: %s, not %+v

pkg/k8s/events.go
```diff
@@ -14,6 +14,14 @@ const (
 	HTTPRouteEventReasonFailedDeployModel = "FailedDeployModel"
 	HTTPRouteEventReasonRetryReconcile    = "Retry-Reconcile"
 
+	// GRPCRoute events
+	GRPCRouteEventReasonReconcile          = "Reconcile"
+	GRPCRouteEventReasonDeploySucceed      = "DeploySucceed"
+	GRPCRouteEventReasonFailedAddFinalizer = "FailedAddFinalizer"
+	GRPCRouteEventReasonFailedBuildModel   = "FailedBuildModel"
+	GRPCRouteEventReasonFailedDeployModel  = "FailedDeployModel"
+	GRPCRouteEventReasonRetryReconcile     = "Retry-Reconcile"
```
mikhail-aws, line 23: whats the difference from http?

controllers/grpcroute_controller.go
```diff
@@ -0,0 +1,376 @@
...
+//+kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=grpcroutes,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=grpcroutes/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=grpcroutes/finalizers,verbs=update
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// TODO(user): Modify the Reconcile function to compare the state specified by
+// the GRPCRoute object against the actual cluster state, and then
+// perform operations to make the cluster state reflect the state specified by
+// the user.
```
mikhail-aws, line 102: It seems quite a bit of code copy pasted, please delete what you dont need

pkg/model/lattice/rule.go
```diff
@@ -32,11 +32,12 @@ type RuleSpec struct {
...
 	PathMatchExact  bool   `json:"pathmatchexact"`
 	PathMatchPrefix bool   `json:"pathmatchprefix"`
-	// Header
+	// HeaderMatch
 	NumOfHeaderMatches int `json:"numofheadermatches"`
 	MatchedHeaders     [MAX_NUM_OF_MATCHED_HEADERS]vpclattice.HeaderMatch
-	// Method
-	Method     string     `json:"method"`
+	// MethodMatch
+	Method string `json:"method"`
```
mikhail-aws, line 39: nit: can we align properties, remove empty lines between them. comments can be added at the end of the line

mikhail-aws (commented):
I think it's worth to invest time on reusing existing http controller, there are similarities. Also amount of copy-pasted code good indicator for reuse.

controllers/grpcroute_controller.go
```diff
@@ -0,0 +1,376 @@
...
+			r.updateGRPCRouteStatus(ctx, serviceStatus.DNS, grpcRoute)
+		}
+	}
+
+	return err
+
+}
+
+func (r *GRPCRouteReconciler) updateGRPCRouteStatus(ctx context.Context, dns string, coreRoute *core.GRPCRoute) error {
+	glog.V(6).Infof("updateGRPCRouteStatus: grpcRoute %v, dns %v\n", coreRoute, dns)
```
zijun726911, line 294: What we cannot combine  and updateHTTPRouteStatus() and updateGRPCRouteStatus()  and use coreRoute core.Route as the last parameter? Or do you plan to combine it later?

cmd/aws-application-networking-k8s/main.go
xWink: Sure thing, will do next commit!

controllers/eventhandlers/grpcroute.go
xWink: Will do!

controllers/eventhandlers/grpcroute.go
xWink: Will do!

controllers/grpcroute_controller.go
xWink: The conditional check would look like this
```go
func (r *RouteReconciler) getRoute(ctx context.Context, req ctrl.Request) (core.Route, error) {
	var err error

	grpcRoute := &core.GRPCRoute{}
	if err = r.client.Get(ctx, req.NamespacedName, grpcRoute.K8sObject()); err == nil {
		return grpcRoute, nil
	}

	httpRoute := &core.HTTPRoute{}
	if err = r.client.Get(ctx, req.NamespacedName, httpRoute.K8sObject()); err == nil {
		return httpRoute, nil
	}

	return nil, err
}
```

My concern here is that this relies on the name and namespace of the resource. If I create an HTTPRoute with the same name and namespace as a GRPCRoute, this check cannot tell the difference between which one is being reconciled. As a result, I would rather opt for two separate controllers, but reuse some common functions between them.

pkg/gateway/model_build_lattice_service.go
xWink: Sure thing!

pkg/k8s/events.go
xWink: Fair point, consolidating these

pkg/model/lattice/rule.go
xWink: Sure thing

controllers/grpcroute_controller.go
xWink: Yup!

controllers/eventhandlers/grpcroute.go
xWink: This is doing exactly what we do for HTTPRoute. According to the doc, an empty list indicates that the route has not been attached to any gateway. In the case of a GRPCRoute update, I suspect this is outside the scope of this issue.

controllers/grpcroute_controller.go
xWink: Yup, I have a working consolidated controller and event handler for routes. Will include it in next commit!

controllers/grpcroute_controller.go
xWink: Yup! Done next revision

xWink:
Addressed all comments. E2E tests still pass:
```

Ran 11 of 11 Specs in 2448.453 seconds
--- PASS: TestIntegration (2449.31s)
```

cmd/aws-application-networking-k8s/main.go
```diff
@@ -144,11 +146,9 @@ func main() {
...
 	}
 
-	httpRouteReconciler := controllers.NewHttpRouteReconciler(cloud, mgr.GetClient(), mgr.GetScheme(), mgr.GetEventRecorderFor("httproute"), finalizerManager, latticeDataStore)
-
-	if err = httpRouteReconciler.SetupWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create controller", "controller", "HTTPRoute")
-		os.Exit(1)
+	err = controllers.RegisterAllRouteControllers(ctrlLog.Named("route"), cloud, latticeDataStore, finalizerManager, mgr)
+	if err != nil {
+		setupLog.Fatalf("route controller setup failed: %s", err)
```
mikhail-aws, line 151: thank you

controllers/route_controller.go
```diff
@@ -0,0 +1,394 @@
...
+		return client.IgnoreNotFound(err)
+	}
+
+	if !r.isRouteRelevant(ctx, route) {
+		// not relevant
+		return nil
+	}
+
+	if !route.DeletionTimestamp().IsZero() {
+		r.log.Info("Deleting")
```
mikhail-aws, line 163: for consistency with other controllers, r.log.Infow("reconcile, deleting", "name", req.Name)  
and log at the end that request reconciled successfully

controllers/route_controller.go
```diff
@@ -0,0 +1,394 @@
...
+		r.eventRecorder.Event(route.K8sObject(), corev1.EventTypeNormal,
+			k8s.RouteEventReasonReconcile, "Deleting Reconcile")
+		if err := r.cleanupRouteResources(ctx, route); err != nil {
+			return fmt.Errorf("failed to cleanup GRPCRoute %+v: %w", route, err)
+		}
+		updateRouteListenerStatus(ctx, r.client, route)
```
mikhail-aws, line 169: dropped error

controllers/route_controller.go
```diff
@@ -0,0 +1,394 @@
...
+		return err
+	}
+}
+
+func (r *RouteReconciler) getRoute(ctx context.Context, req ctrl.Request) (core.Route, error) {
+	switch r.routeType {
+	case HTTP:
+		httpRoute := &core.HTTPRoute{}
+		err := r.client.Get(ctx, req.NamespacedName, httpRoute.K8sObject())
+		return httpRoute, err
```
mikhail-aws, line 189: I know it's cumbersome to do all if err !=nil return, but lets keep it that way
if err != nil {
  return nil, err
}
return httpRoute,nil

controllers/route_controller.go
```diff
@@ -0,0 +1,394 @@
...
+}
+
+func (r *RouteReconciler) cleanupRouteResources(ctx context.Context, route core.Route) error {
+	_, _, err := r.buildAndDeployModel(ctx, route)
+	return err
+}
+
+func (r *RouteReconciler) isRouteRelevant(ctx context.Context, route core.Route) bool {
+	if len(route.Spec().ParentRefs()) == 0 {
+		r.log.Infof("Ignore Route which has no ParentRefs gateway %+v ", route.Spec())
```
mikhail-aws, line 226: +v on route spec too much for log info, should be debug, also route name probably is enough for logging

mikhail-aws (commented):
Looks much better! I'm a bit worried that we have lots of logic in controller and not covered by unit tests at all. We should have a follow up on this, after gRPC wrap up, it's technical dept that will come back to us:
- move everything into pkg, let controller be simple lunching pad
- start adding unit tests for all these functions

controllers/route_controller.go
xWink: This seems like a more usable log. Made the change in following commit!

controllers/route_controller.go
xWink: Done next commit

controllers/route_controller.go
xWink: Good point, fixed next revision!

zijun726911:
When I run the `kubectl apply -f config/crds/bases/k8s-gateway-v1alpha2.yaml`

it meet the:

```
Resource: "apiextensions.k8s.io/v1, Resource=customresourcedefinitions", GroupVersionKind: "apiextensions.k8s.io/v1, Kind=CustomResourceDefinition"
Name: "httproutes.gateway.networking.k8s.io", Namespace: ""
for: "config/crds/bases/k8s-gateway-v1alpha2.yaml": error when patching "config/crds/bases/k8s-gateway-v1alpha2.yaml": CustomResourceDefinition.apiextensions.k8s.io "httproutes.gateway.networking.k8s.io" is invalid: status.storedVersions[0]: Invalid value: "v1beta1": must appear in spec.versions

``` 

Is that normal?

zijun726911:
I think created TG ProtocolVersion should be `GRPC` or not? but you didn't change this part?:
https://github.com/aws/aws-application-networking-k8s/blob/662a1fb97b21d4f3476a81a1763b4a52b982bf49/pkg/gateway/model_build_targetgroup.go#L361
image

examples/grpc-route.yaml
```diff
@@ -1,14 +1,17 @@
-apiVersion: gateway.networking.k8s.io/v1beta1
-kind: HTTPRoute
+apiVersion: gateway.networking.k8s.io/v1alpha2
+kind: GRPCRoute
 metadata:
   name: grpc
```
zijun726911, line 4: Could you change this name to some thing like my-grpc-route?

xWink:
> I think created TG ProtocolVersion should be `GRPC` or not? but you didn't change this part?:
> 
> https://github.com/aws/aws-application-networking-k8s/blob/662a1fb97b21d4f3476a81a1763b4a52b982bf49/pkg/gateway/model_build_targetgroup.go#L361
> 
> Or, you plan to change it in a separate PR? image

Yes, this is part of the actions required for a future PR mentioned in the description. This PR is purely meant to set up the controller and matches, the next PR will be to enable gRPC traffic. There are multiple places that need changes, with the target group model builder being one of them.

examples/grpc-route.yaml
xWink: Why?

mikhail-aws (approved):

zijun726911 (approved):
LGTM, looking forward to your next few GRPCRoute PRs

PR #344, opened 2023-08-19
e2e test enhancements for better performance and reliability

erikfuller:
**What type of PR is this?**
cleanup

n/a

Improves dev quality of life, following on from https://github.com/aws/aws-application-networking-k8s/pull/330, this PR further improves the performance of e2e tests, their reliability, and resource cleanup. Main features of this PR:

* Provision a single gateway for the test suite
* Create all objects against the "non-default" namespace
* Move all persistent objects out of individual tests to allow cleanup even on failure
* Update all ```Eventually()``` blocks to use local expectations so failed assertions do not break out
* reliability enhancements
* minor refactoring

e2e tests
```
Ran 11 of 11 Specs in 1744.934 seconds
--- PASS: TestIntegration (1746.05s)
PASS
ok  	github.com/aws/aws-application-networking-k8s/test/suites/integration	1746.793s
```

n/a

n/a

no

### DIFF
--- a/test/suites/integration/httproute_method_match_test.go
+++ b/test/suites/integration/httproute_method_match_test.go
@@ -2,41 +2,49 @@ package integration
 
 import (
 	"fmt"
-	"log"
-	"time"
-
 	"github.com/aws/aws-application-networking-k8s/test/pkg/test"
 	"github.com/aws/aws-sdk-go/service/vpclattice"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/samber/lo"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"log"
+	"sigs.k8s.io/gateway-api/apis/v1beta1"
+	"time"
 )
 
 var _ = Describe("HTTPRoute method matches", func() {
-	It("HTTPRoute should route by HTTP method", func() {
-		gateway := testFramework.NewGateway("", k8snamespace)
-		deployment1, getService := testFramework.NewHttpApp(test.HTTPAppOptions{Name: "test-get", Namespace: k8snamespace})
-		deployment2, postService := testFramework.NewHttpApp(test.HTTPAppOptions{Name: "test-post", Namespace: k8snamespace})
-		methodMatchHttpRoute := testFramework.NewMethodMatchHttpRoute(gateway, getService, postService, "route-by-method", k8snamespace)
+
+	var (
+		methodMatchHttpRoute *v1beta1.HTTPRoute
+		deployment1          *appsv1.Deployment
+		getService           *corev1.Service
+		deployment2          *appsv1.Deployment
+		postService          *corev1.Service
+	)
+
+	BeforeEach(func() {
+		deployment1, getService = testFramework.NewHttpApp(test.HTTPAppOptions{Name: "test-get", Namespace: k8snamespace})
+		deployment2, postService = testFramework.NewHttpApp(test.HTTPAppOptions{Name: "test-post", Namespace: k8snamespace})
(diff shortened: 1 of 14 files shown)


mikhail-aws (approved):
nice, thank you

test/suites/integration/httproute_method_match_test.go
```diff
@@ -89,40 +99,38 @@ var _ = Describe("HTTPRoute method matches", func() {
...
+			g.Expect(stdout).To(ContainSubstring("test-post handler pod"))
+		}).WithTimeout(30 * time.Second).WithOffset(1).Should(Succeed())
+
+		Eventually(func(g Gomega) {
+			invalidCmd := fmt.Sprintf("curl -X DELETE %s", dnsName)
+			stdout, _, err := testFramework.PodExec(pods[0].Namespace, pods[0].Name, invalidCmd, true)
+			g.Expect(err).To(BeNil())
+			g.Expect(stdout).To(ContainSubstring("Not Found"))
+		}).WithTimeout(30 * time.Second).WithOffset(1).Should(Succeed())
+	})
```
zijun726911, line 123: I guess we can bear first few `curl` failed due to immature dataplane xds config and eventually expect traffic can work?

test/suites/integration/suite_test.go
```diff
@@ -3,26 +3,42 @@ package integration
...
 	if vpcid == "" {
 		Fail("CLUSTER_VPC_ID environment variable must be set to run integration tests")
 	}
+
+	testFramework.ExpectToBeClean(ctx)
+
+	// would be good to simply wait here until service network has been provisioned and associated
+	testGateway = testFramework.NewGateway("test-gateway", k8snamespace)
+	testFramework.ExpectCreated(ctx, testGateway)
 })
```
zijun726911, line 33: Should we use `Eventually IsVpcAssociatedWithServiceNetwork()` here to verify vpcAssociation?

zijun726911:
Thanks Eric! overall LGTM

test/suites/integration/httproute_method_match_test.go
erikfuller: Yes. There will be a short delay as configuration converges and Lattice is able to process the request successfully. We don't have a good way of programmatically knowing this convergence has happened, so we rely on getting there "Eventually" :)

test/suites/integration/suite_test.go
erikfuller: Thanks for flagging, I've updated the PR with your suggestion.

zijun726911: it should have `.Should(Succeed())`:

```
Eventually(func(g Gomega) {
}).Should(Succeed())
```

erikfuller: Added.

test/suites/integration/httproute_method_match_test.go
```diff
@@ -89,40 +99,38 @@ var _ = Describe("HTTPRoute method matches", func() {
...
-
-		testFramework.ExpectDeleted(ctx,
-			gateway,
+
+		Eventually(func(g Gomega) {
+			cmd := fmt.Sprintf("curl -X GET %s", dnsName)
+			stdout, _, err := testFramework.PodExec(pods[0].Namespace, pods[0].Name, cmd, true)
+			g.Expect(err).To(BeNil())
+			g.Expect(stdout).To(ContainSubstring("test-get handler pod"))
+		}).WithTimeout(30 * time.Second).WithOffset(1).Should(Succeed())
```
mikhail-aws, line 108: nit: eventually itself is assertion block, that can assert return value of the function inside, for simple tests like this we can drop a few things. Assertion on substring should be enough without checking error

```go
func curlPod(pod Pod, curlCmd string) (string, error) {
		stdout, _, err := testFramework.PodExec(pod.Namespace, pod.Name, curlCmd, true)
                return stdout, err
}

Eventually(func() string {
	out, _ := curlPod(pods[0], fmt.Sprintf("curl -X POST %s", dnsName))
	return out
}).Should(ContainSubstring("200 OK"))
```

Issue #342, opened 2023-08-18
Tag ServiceAssociation resources created in VPC Lattice


solmonk:
The controller needs to identify the resources that it owns, currently we are having (or planning to have) indirect methods such as through VPC IDs and account IDs.

Ideally we should have tags to all VPC Lattice resources that the controller creates, and have reliable identifiers for resource owner.

Related: #317

solmonk:
The tag itself is added, will track separately for any enhancements.

Issue #321, opened 2023-08-11
add golangci-lint


mikhail-aws:
We need to add set of linters to the project. A popular one is golangci-lint - a collection of different linters for bugs, style, complexity, errors, etc.

https://github.com/golangci/golangci-lint

PR #345, opened 2023-08-20
refactor service_manager.go

### LINKED ISSUE #317: Feature request for controller to acknowledge service networks and services managed by other accounts
Is there a recommended configuration for the k8s controller when lattice services are expected to be shared across accounts?

In an example scenario:
I have account A and account B

- account A uses kubernetes, account B uses ec2 instances. Each account has its own lattice service network already configured.
- account A uses the k8s controller to manage lattice services, account B uses terraform/manual configuration
- account A wants account B to be able to call its lattice service A
- account A shares lattice service A with account B
- account B associates service A with service network B -- but this here it seems the k8s controller does not like this. https://github.com/aws/aws-application-networking-k8s/blob/3927a6ea027d3f84ed363a64e4e43a6da23b806f/pkg/deploy/lattice/service_manager.go#L271-L283 per this code, the controller will delete the service<>service network association since it's not captured in the HTTPRoute of service A

I know there is the annotation that makes kubernetes ignore vpc associations to a gateway[0]. However even with this, the gateway/service network is still created in account A. In this scenario, account B wants to manage its own infrastructure as much as possible. With the current limitations it seems that all lattice resources must be created in 1 account and then shared out to other accounts.

Some possible solutions I can think of

1. consider foreign resources a special case such as having an option to force the controller to not delete a service<>service network association if the account ids of these resources are different
2. similar to the previous idea, allow a reference to a non-existent gateway which refers to a remote service network (on another account)
3. allow k8s controller to manage resources for multiple accounts. (I like this option less since it gives less independence to other accounts to manage their own AWS resources)

I am definitely open to other ideas I am not considering and would be happy to add more context
[0] https://github.com/aws/aws-application-networking-k8s/blob/main/docs/multi-sn.md

Note: after discussion with @liwenwu-amazon on slack is sounds like prefer solution 1

[In her words](https://kubernetes.slack.com/archives/C052EBZPAUR/p1691549259484649): "lattice controller not to disassociate lattice service from a lattice service network , if that lattice service network owned by a different account"

mikhail-aws:
Note:
- major change in service_manager.go and service_manager_test.go, completely rewritten, diff might be hard to read. Adding links to source files for readers [service_manager](https://github.com/mikhail-aws/aws-application-networking-k8s/blob/bug-fix-317/pkg/deploy/lattice/service_manager.go) and [test](https://github.com/mikhail-aws/aws-application-networking-k8s/blob/bug-fix-317/pkg/deploy/lattice/service_manager_test.go)

Testing:
- unit tests
- e2e tests
```
Ran 13 of 13 Specs in 2343.684 seconds
SUCCESS! -- 13 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestIntegration (2344.70s)
```


mikhail-aws:
#342

pkg/deploy/lattice/service_manager.go
```diff
@@ -2,318 +2,285 @@ package lattice
...
+type CreateSnSvcAssocReq = vpclattice.CreateServiceNetworkServiceAssociationInput
+type CreateSnSvcAssocResp = vpclattice.CreateServiceNetworkServiceAssociationOutput
+type DelSnSvcAssocReq = vpclattice.DeleteServiceNetworkServiceAssociationInput
+type DelSnSvcAssocResp = vpclattice.DeleteServiceNetworkServiceAssociationOutput
+type GetSvcReq = vpclattice.GetServiceInput
+type SvcSummary = vpclattice.ServiceSummary
+type GetTagsReq = vpclattice.ListTagsForResourceInput
+type GetTagsResp = vpclattice.ListTagsForResourceOutput
+type ListAssocsReq = vpclattice.ListServiceNetworkServiceAssociationsInput
+type SnSvcAssocSummary = vpclattice.ServiceNetworkServiceAssociationSummary
```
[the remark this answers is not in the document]
mikhail-aws: only for these two? tags and assocs?

docs/developer.md
```diff
@@ -1,57 +1,113 @@
-## Developer Guide
+# Developer Guide
```
erikfuller, line 1: I'm a little confused - why is this file showing up in the diff? I'm ont seeing any changes from what's on main?

pkg/deploy/lattice/service_manager.go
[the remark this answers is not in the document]
solmonk: Other changes are acceptable I think. Just want to make sure we are not obfuscating API names

pkg/deploy/lattice/service_manager.go
```diff
@@ -2,318 +2,301 @@ package lattice
...
-
-	if err == nil {
-		for _, r := range resp {
-			if aws.StringValue(r.Name) == serviceName {
-				glog.V(6).Info("service ", serviceName, " already exists with arn ", *r.Arn, "\n")
-				return r, err
-			}
+// update SN-Svc associations, if svc has no SN associations will delete all of them
+// does not delete associations that are not tagged by controller
+func (m *defaultServiceManager) updateSnSvcAssocs(ctx context.Context, svc *Service, svcSum *SvcSummary) error {
```
erikfuller, line 148: I'm not sure we should be using this method for both ```Create/Update``` and ```Delete```. On delete, shouldn't we be deleting all associations whether or not we created them?

pkg/deploy/lattice/service_manager.go
```diff
@@ -2,318 +2,301 @@ package lattice
...
+	err = m.updateSnSvcAssocs(ctx, svc, svcSum)
+	if err != nil {
+		return err
+	}
 
+	delInput := vpclattice.DeleteServiceInput{
+		ServiceIdentifier: svcSum.Id,
 	}
-	return nil
+	_, err = m.cloud.Lattice().DeleteServiceWithContext(ctx, &delInput)
```
erikfuller, line 300: At this point, could we be dealing with a service that doesn't have our management tags? Should we be checking? How do you envision upgrades working from a previous version of the controller to this one?

erikfuller (changes requested):
Thanks for putting this together. I'm still working my way through the tests but wanted to get my initial comments up. I think the big question is whether or not the proposed tag actually improves the situation - so far I'm not convinced. I do appreciate the refactoring changes in ```service_manager.go```, it's much more readable and has less duplication.

pkg/model/lattice/servicenetwork.go
```diff
@@ -5,6 +5,7 @@ import (
 )
 
 const (
+	TagManagedBy                = "ManagedBy"
 	K8SServiceNetworkOwnedByVPC = "K8SServiceNetworkOwnedByVPC"
 	K8SServiceOwnedByVPC        = "K8SServiceOwnedByVPC"
```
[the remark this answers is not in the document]
mikhail-aws: yes, it's a larger refactoring, need follow up on this

pkg/aws/cloud.go
```diff
@@ -37,37 +68,89 @@ func NewCloud(log gwlog.Logger) (Cloud, error) {
...
-	}, nil
+	lattice := services.NewDefaultLattice(sess, config.Region)
+	eks := services.NewDefaultEKS(sess, config.Region)
+	cl := NewDefaultCloud(lattice, eks, cfg)
+	return cl, nil
 }
 
-var _ Cloud = &defaultCloud{}
+// Used in testing and mocks
+func NewDefaultCloud(l services.Lattice, e services.EKS, cfg CloudConfig) Cloud {
```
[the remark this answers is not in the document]
mikhail-aws: agree, a bit too much

pkg/deploy/lattice/service_manager.go
```diff
@@ -2,318 +2,285 @@ package lattice
 
 import (
```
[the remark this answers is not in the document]
mikhail-aws: will do 2 prs

pkg/deploy/lattice/service_manager.go
```diff
@@ -2,318 +2,301 @@ package lattice
...
+	svcSum, err := m.getService(ctx, svc.LatticeName())
+	if err != nil {
+		return err
+	}
+	if svcSum == nil {
+		return nil
+	}
 
-		}
+	svc.Spec.ServiceNetworkNames = []string{}
```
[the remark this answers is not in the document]
mikhail-aws: I was having same debate with myself. Since you also mentioning this, I will change it

pkg/deploy/lattice/service_manager.go
mikhail-aws: Actually I'm using this on Update and Delete but not create. But I agree that delete should be more straightforward with deletion. So create/update/delete will have own little method.

pkg/deploy/lattice/service_manager.go
mikhail-aws: There is not tag check on service deletion, and it wasnt before as far as I can tell. About moving forward - while we still at alpha stage and breaking changes happens, there is little concern about backward compatibility. Starting from Beta we cant do that.

mikhail-aws: My initial thought was to make something like "ensureTag" that will go through existing associations and tag them if not tagged. But then controller does not about any associations right now. If I restart controller local storage is gone, and next reconcile event can already mutate list of SN's. So at this point we need to have any tag, that can be a starting point for future upgrades.

mikhail-aws:
@erikfuller @solmonk 

- updated this PR with only refactoring for service_manager, removed bug fix
- re-ran e2e tests with latest changed from main

pkg/deploy/lattice/service_manager.go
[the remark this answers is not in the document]
mikhail-aws: split update and delete into different methods, no longer mutate spec

pkg/deploy/lattice/service_manager.go
mikhail-aws: will do tags in another PR, removed from this

pkg/deploy/lattice/service_manager.go
```diff
@@ -2,318 +2,322 @@ package lattice
...
-	if err == nil {
-		for _, r := range resp {
-			if aws.StringValue(r.Name) == serviceName {
-				glog.V(6).Info("service ", serviceName, " already exists with arn ", *r.Arn, "\n")
-				return r, err
-			}
-		}
-	} else {
-		return nil, err
+func svcStatusFromCreateSvcResp(resp *CreateSvcResp) ServiceStatus {
```
erikfuller, line 119: I see somewhere else similar code is used, I would suggest changing the signature to take the arn, id, and dns - I don't think for example the ```resp``` value here can be nil and identical logic lives in ```updateServiceAndAsociations```.

pkg/deploy/lattice/service_manager.go
```diff
@@ -2,318 +2,322 @@ package lattice
...
+	status := ServiceStatus{}
+	if resp == nil {
+		return status
 	}
-	return nil, nil
+	status.Arn = aws.StringValue(resp.Arn)
```
erikfuller, line 124: So...I think we should rename all these to ```service``` and reserve ```status``` for when we're actually working with some kind of status.

pkg/deploy/lattice/service_manager_test.go
```diff
@@ -3,1035 +3,296 @@ package lattice
...
+		status, err := m.Create(ctx, svc)
+		assert.Nil(t, err)
+		assert.Equal(t, "arn", status.Arn)
+	})
+
+	// Update is more complex than create, we need to apply diff for Sn-Svc associations
+	// This test covers creation/deletion for multiple SN's
+	// Service is associated with sn-keep and sn-delete,
+	// for update we need to delete sn-delete and add sn-add
+	t.Run("update service's associations", func(t *testing.T) {
```
erikfuller, line 73: This is a nice test!

pkg/deploy/lattice/service_manager_test.go
```diff
@@ -3,1035 +3,296 @@ package lattice
...
+			CreateServiceWithContext(gomock.Any(), gomock.Any()).
+			Return(&CreateSvcResp{
+				Arn:      aws.String("arn"),
+				DnsEntry: &vpclattice.DnsEntry{DomainName: aws.String("dns")},
+				Id:       aws.String("svc-id"),
+			}, nil)
+
+		// assert that we call create association
+		lat.EXPECT().
+			CreateServiceNetworkServiceAssociationWithContext(gomock.Any(), gomock.Any(), gomock.Any()).
```
erikfuller, line 50: Would be nice to include some assertions in this test on arguments

pkg/deploy/lattice/service_manager_test.go
```diff
@@ -3,1035 +3,296 @@ package lattice
...
+					Arn:                aws.String("assoc-arn"),
+					Id:                 aws.String("assoc-id"),
+					ServiceNetworkName: aws.String("sn"),
+					Status:             aws.String(vpclattice.ServiceNetworkServiceAssociationStatusActive),
+				},
+			}, nil)
+
+		// assert we delete association and service
+		lat.EXPECT().
+			DeleteServiceNetworkServiceAssociationWithContext(gomock.Any(), gomock.Any(), gomock.Any()).
```
erikfuller, line 166: Worth adding some assertions on args?

pkg/deploy/lattice/service_manager_test.go
```diff
@@ -3,1035 +3,296 @@ package lattice
...
-		if !tt.errOnAssociating {
-			listMeshServiceAssociationsOutput = []*vpclattice.ServiceNetworkServiceAssociationSummary{}
-			for i := 0; i < len(tt.existingSNs); i++ {
-				listMeshServiceAssociationsOutput = append(listMeshServiceAssociationsOutput,
-					&vpclattice.ServiceNetworkServiceAssociationSummary{ServiceNetworkName: &tt.existingSNs[i].snName})
+}
 
-			}
-			mockVpcLatticeSess.EXPECT().ListServiceNetworkServiceAssociationsAsList(ctx, gomock.Any()).Return(listMeshServiceAssociationsOutput, nil)
+func TestSnSvcAssocsDiff(t *testing.T) {
```
erikfuller, line 239: Thanks for adding these - good to have full coverage on this logic.

erikfuller (changes requested):
Thanks for the PR, very readable! I've left a number of suggestions, but I think the ones I care most about are:

#### 1. Prefer pointer return types
In addition to being able to return ```nil``` instead of having a placeholder empty return, this will help ensure calling code can't act on an empty value, for example if the calling code forgets to check ```err```. With ```nil```, any attempts to act on the value will cause the code to fail, most likely very loudly, which I think is better than taking an incorrect action. Example:

```golang
func (...) updateServiceAndAssociations(...) (ServiceStatus, error) {
```
becomes
```golang
func (...) updateServiceAndAssociations(...) (*ServiceStatus, error) {
```

#### 2. Rename ```status ServiceStatus``` variables to ```service```
These are more services than statuses, and given we have actual status values around, it's confusing to read these variable names as they are right now.

Other comments can be considered optional, but if you feel like making changes based on them go for it.

mikhail-aws:
@erikfuller, thank you for review

> 1. Prefer pointer return types

Your reasoning is correct about failing on nil is more obvious than failing on empty struct. But it's debatable that having nil exception is better than empty struct. Also it's not exactly how errors should be handled in go. We should rely on error checking. Error checking can be (and should be) enforced by lint. Example https://github.com/kisielk/errcheck. Part of #321 

What I consider when return pointer or struct:

 - simple, small structs with primitive values always return by value. In go everything is mutable and cheap concurrency tend to create lots of goroutines with shared (mutable!) objects. So passing copies of data around is safer, and preferable. (frowning at lattice-datastore). Mutability one of the major concerns.
 - use pointer for complex structs with internal states, or expensive to copy
 - use pointer when nil is valid option, an analog to Optional<T>. In this case when we create service and there is no error it will return valid response. (aws sdk is full of pointers, everything is optional...)

Returning struct here struct is fine. I can rewrite code a bit to make it less "verbose"

pkg/aws/cloud.go
```diff
@@ -37,34 +47,41 @@ func NewCloud(log gwlog.Logger) (Cloud, error) {
...
+	return c.eks
 }
 
-func (d *defaultCloud) EKS() services.EKS {
-	return d.eksSess
+func (c *defaultCloud) Config() CloudConfig {
+	return c.cfg
 }
 
 func (d *defaultCloud) GetEKSClusterVPC(name string) string {
```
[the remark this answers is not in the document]
mikhail-aws: Will rename receiver, but it's used and will be used more in my planned refactoring for config propagation. #310 
```
pkg/deploy/lattice/service_manager.go
84:	sn, err := m.datastore.GetServiceNetworkStatus(snName, m.cloud.Config().AccountId)
```

pkg/deploy/lattice/error.go
```diff
@@ -1,9 +1,13 @@
...
+
 const (
 	LATTICE_CREATE_IN_PROGRESS = "LATTICE_CREATE_IN_PROGRESS"
 	LATTICE_DELETE_IN_PROGRESS = "LATTICE_DELETE_IN_PROGRESS"
 	LATTICE_NOT_FOUND          = "LATTICE_NOT_FOUND"
 	LATTICE_ALREADY_EXIST      = "LATTICE_ALREADY_EXIST"
 	LATTICE_RETRY              = "LATTICE_RETRY"
 )
+
+var RetryErr = errors.New(LATTICE_RETRY)
```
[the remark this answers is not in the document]
mikhail-aws: Works fine for golang standard library, io errors are reused. https://cs.opensource.google/go/go/+/refs/tags/go1.21.0:src/internal/oserror/errors.go

pkg/deploy/lattice/service_manager.go
```diff
@@ -2,318 +2,322 @@ package lattice
...
 		}
-
 	}
-	err = s.serviceNetworkAssociationMgr(ctx, service.Spec.ServiceNetworkNames, serviceID)
+	status := svcStatusFromCreateSvcResp(createSvcResp)
+	return status, nil
+}
 
+func (m *defaultServiceManager) createAssociation(ctx context.Context, svcId *string, snName string) error {
+	sn, err := m.datastore.GetServiceNetworkStatus(snName, m.cloud.Config().AccountId)
```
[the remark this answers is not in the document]
mikhail-aws: I think current implementation of datastore introduces more not-handled edge cases than solve problems. We are scratching a surface here. I believe we need to get back to drawing board at some point about entire flow.

pkg/deploy/lattice/service_manager.go
```diff
@@ -2,318 +2,322 @@ package lattice
...
-//		MeshServiceAssociationStatusDeleteFailed
-//		MeshServiceAssociationStatusCreateFailed
-//		MeshServiceAssociationStatusDeleteInProgress
-func (s *defaultServiceManager) Create(ctx context.Context, service *latticemodel.Service) (latticemodel.ServiceStatus, error) {
-
-	// check if exists
-	svcName := latticestore.LatticeServiceName(service.Spec.Name, service.Spec.Namespace)
-	serviceSummary, err := s.findServiceByName(ctx, svcName)
+// find service by name, if not found returns nil without error
+func (m *defaultServiceManager) getService(ctx context.Context, svcName string) (*vpclattice.ServiceSummary, error) {
```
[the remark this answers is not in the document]
mikhail-aws: In original code we dont store/get service from cache, will keep it in this PR, since it's only code refactoring with same semantics.

pkg/deploy/lattice/service_manager.go
mikhail-aws: Status with kind of status make sense :D
It comes from latticemodel, I just created 2 aliases to Service and ServiceStatus. Same as before. A hammer of refactoring will get there one day.

pkg/deploy/lattice/service_manager_test.go
mikhail-aws: Request part is covered in smaller unit test TestCreateSvcReq

mikhail-aws:
@erikfuller 
- renamed ServiceStatus alias to ServiceInfo [link](https://github.com/aws/aws-application-networking-k8s/blob/38d27477508eb341e12a5af8a9b8aa18b1d0fead/pkg/deploy/lattice/service_manager.go#L18)
- renamed "status" variables into svcInfo [link](https://github.com/aws/aws-application-networking-k8s/blob/38d27477508eb341e12a5af8a9b8aa18b1d0fead/pkg/deploy/lattice/service_manager.go#L118)
- added arguments assertion to Create and Delete integ tests

pkg/aws/cloud.go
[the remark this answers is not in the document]
mikhail-aws: sorry, my bad I thought you were asking about Config, but it was GetEKSClusterVPC(), yes it's not used. It's on my personal todo list

erikfuller (approved):
lgtm

Issue #269, opened 2023-06-14
ServiceExport deletion leaks targetgroup if the corresponding K8sService have already been deleted


zijun726911:
ServiceExport deletion leaks targetgroup if the corresponding K8sService have already been deleted.

More details on how to reproduce this bug (could 100% reproduce it):

### For the ServiceExport, it could have 2 scenarios:

- Scenario1, all good no bug: 
    -  Given: create one k8sService, and one its corresponding ServiceExport, create ServiceExport could trigger controller to create a target group success
    -  __Now the k8sService still exists__,  if I do `kubectl delete serviceexport myserviceexport` , could trigger target group deletion correctly

- Scenario2, has bug,  target group leaking
    -  Given: create 1 k8sService, and 1 its corresponding ServiceExport, create ServiceExport could trigger controller to create a target group success
    -  do `kubectl delete service myservice`
    -  __Now the k8sService does not exist__, if I do `kubectl delete serviceexport myserviceexport`,  target group deletion will not be triggered, which cause target group leaking

liwenwu-amazon:
We should use service finalizer to prevent K8S service being deleted,  if this service is referenced by HTTPRoute or ServiceExport

zijun726911:
> We should use service finalizer to prevent K8S service being deleted, if this service is referenced by HTTPRoute or ServiceExport
Can not quite get what the benefit of it? That can also increase the chance the user is blocked to delete K8S services?

solmonk:
Why would we interrupt k8s service lifecycle though? We already have a behavior of deregistering all targets when service is deleted. HTTPRoute will have a rule that points to empty upstream but it is fine because that's actually the way the user configured k8s config. We should update HTTPRoute resolvedRef status to false instead.

I believe we should move away of user experience of waiting forever. Finalizers are hooks that handle necessary cleanup process, and should not be used for blocking users.

liwenwu-amazon:
This is very similar to the logic that lattice had on target group and service that you are not allowed to delete rules or services if it references some target group.

Allow deleting K8S service while it is referenced by other objects (e.g. serviceexport, httproute) make those object handling logic more complicated and error prone.

liwenwu-amazon:
for "user experience of waiting forever.", i believe it is only come from kubectl command `kubectl delete xxxx`.  If you want a error msg (similar to lattice CLI), you need to add webhook for our controller.

solmonk:
[...]

PR #348, opened 2023-08-22
[Bugfix] Fix serviceexport deletion leaks targetgroup

zijun726911:
**What type of PR is this?**
bug fix for serviceexport deletion leaks targetgroup issue  #269 

**Changes**:

- Refactor (t *targetGroupModelBuildTask) BuildTargetGroup(), make sure it can build model for serviceExport deletion request (the key point is the built model should have `IsDeleted: true`,`tg.Spec.LatticeID = dsTG.ID` )
- Split the ServiceExport creation and delete logic into 2 smaller helper functions: `buildTargetGroupForServiceExportCreation` and `buildTargetGroupForServiceExportDeletion` to make code eaiser to read
- Make sure we add/delete/update the correct latticeDataStore TargetGroup cache, so that the garbage collector logic: `SynthesizeSDKTargetGroups()` could be able to identifed the correct stale tgs, check the code changes for detail

Run multiple times, all e2etest cases pass
```
[AfterSuite] PASSED [30.210 seconds]

Ran 13 of 13 Specs in 2320.742 seconds
SUCCESS! -- 13 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestIntegration (2321.24s)
```
Checked the aws vpc lattice console after e2etest finished, no target group leaking

### Manual test:

- Scenario 1: stand-alone serviceExport(No httproute created, in other words, created targetGroups don't have latticeServiceRule refer to it):

    - Scenario 1.1 create k8Service, serviceExport,tg created success, delete k8Service first, then delete serviceExport, tg, targets deleted success, no TG leak
    - Scenario 1.2 create k8Service, serviceExport, delete serviceExport first, then delete k8Service, , tg, targets deleted success, no TG leak

- Scenario 2: serviceExport's serviceImport referenced by a httproute
    - Scenario 2.1 Create httproute, k8Service, serviceImport serviceExport, all lattice created success. then, delete serviceExport first, tg will not be deleted as expected, since this tg is refered by latticeService. `kubectl delete serviceExport inventory-ver2` will not be stuck. Then, delete the httproute, the garbage collector logic `SynthesizeSDKTargetGroups()` be able to clean this inventory-ver2 tg, no any lattice resource leak
    - Scenario 2.2 Create httproute, k8Service, serviceImport serviceExport, all lattice created success. then, delete the the httproute first, then the  serviceExport become "stand-alone" status, delete this serviceExport could delete `inventory-ver2` tg success

```release-note

```

### DIFF
--- a/pkg/deploy/lattice/target_group_synthesizer.go
+++ b/pkg/deploy/lattice/target_group_synthesizer.go
@@ -286,7 +286,6 @@ func (t *targetGroupSynthesizer) SynthesizeSDKTargetGroups(ctx context.Context) 
 			}
 
 		}
-
 		// the routename for serviceimport is ""
 		if tg, err := t.latticeDataStore.GetTargetGroup(*sdkTG.getTargetGroupOutput.Name, "", true); err == nil {
 			glog.V(6).Infof("Ignore target group created by service import %v\n", tg)
@@ -423,7 +422,7 @@ func (t *targetGroupSynthesizer) SynthesizeTriggeredTargetGroupsDeletion(ctx con
 
 		if resTargetGroup.Spec.Config.IsServiceImport {
 			glog.V(2).Infof("Deleting service import target group from local datastore %v", resTargetGroup.Spec.LatticeID)
-			t.latticeDataStore.DelTargetGroup(resTargetGroup.Spec.Name, resTargetGroup.Spec.Config.K8SHTTPRouteName, resTargetGroup.Spec.Config.IsServiceImport)
+			t.latticeDataStore.DelTargetGroup(resTargetGroup.Spec.Name, "", resTargetGroup.Spec.Config.IsServiceImport)
 		} else {
 			// For delete TargetGroup request triggered by k8s service, invoke vpc lattice api to delete it, if success, delete the tg in the datastore as well
 			err := t.targetGroupManager.Delete(ctx, resTargetGroup)
--- a/pkg/gateway/model_build_targetgroup.go
+++ b/pkg/gateway/model_build_targetgroup.go
@@ -2,6 +2,7 @@ package gateway
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/golang/glog"
 
@@ -72,7 +73,7 @@ func (b *targetGroupBuilder) Build(ctx context.Context, srvExport *mcs_api.Servi
 		Client:    b.Client,
 	}
 
-	if err := task.run(ctx); err != nil {
+	if err := task.buildModel(ctx); err != nil {
 		return task.stack, task.targetGroup, corev1.ErrIntOverflowGenerated
 	}
 
@@ -80,39 +81,30 @@ func (b *targetGroupBuilder) Build(ctx context.Context, srvExport *mcs_api.Servi
 }
 
 // for serviceexport
-func (t *targetGroupModelBuildTask) run(ctx context.Context) error {
-	/*
-		if !t.serviceExport.DeletionTimestamp.IsZero() {
-			// TODO handle delete
-			return nil
-		}
-	*/
-
-	err := t.buildModel(ctx)
-
-	return err
-}
-
-// for serviceexport
 func (t *targetGroupModelBuildTask) buildModel(ctx context.Context) error {
-	err := t.BuildTargetGroup(ctx)
+	err := t.buildTargetGroupForServiceExport(ctx)
 
 	if err != nil {
 		glog.V(6).Infof("Failed to build TargetGroup when serviceExport buildModel for name %v namespace %v \n", t.serviceExport.Name, t.serviceExport.Namespace)
 		return err
 	}
+	if !t.serviceExport.DeletionTimestamp.IsZero() {
+		//for serviceExport deletion request, we don't need to build targets model
+		return nil
+	}
 
-	err = t.BuildTargets(ctx)
+	err = t.buildTargets(ctx)
 
 	if err != nil {
 		glog.V(6).Infof("Failed to build Targets when serviceExport buildModel for name %v namespace %v \n ", t.serviceExport.Name, t.serviceExport.Namespace)
 	}
 
 	return nil
+
 }
 
 // triggered from service exports/targetgroups
-func (t *targetGroupModelBuildTask) BuildTargets(ctx context.Context) error {
+func (t *targetGroupModelBuildTask) buildTargets(ctx context.Context) error {
 
 	targetTask := &latticeTargetsModelBuildTask{
 		Client:      t.Client,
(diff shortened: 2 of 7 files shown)


pkg/latticestore/latticestore.go
```diff
@@ -392,6 +402,11 @@ func (ds *LatticeDataStore) SetTargetGroupByBackendRef(name string, routeName st
 
 }
 
+// TODO: make this DelTargetGroup() method to be private, create seperate methods:
```
xWink, line 405: What's the purpose of these TODO items?

xWink (commented):
I don't see any issues, but I'm still learning about this logic flow. Would like to see other reviewers comment before approving. Note to Mikhail if he reviews, don't worry about glog usage, I will be fixing it as part of my PR anyways.

pkg/gateway/model_build_targetgroup_test.go
```diff
@@ -145,7 +145,10 @@ func Test_TGModelByServicexportBuild(t *testing.T) {
 			}
 
 			ds := latticestore.NewLatticeDataStore()
-
+			if !tt.svcExport.DeletionTimestamp.IsZero() {
+				tgName := latticestore.TargetGroupName(tt.svcExport.Name, tt.svcExport.Namespace)
+				ds.AddTargetGroup(tgName, "vpc-123456789", "123456789", "tg-123", false, "")
+			}
```
zijun726911, line 151: when test the ServiceExport deletion, the latticestore should already have a entry for it, so that the built tg.Spec.LatticeID will have this tgId, and it will be pass to the target_group_manage to delte this tg

pkg/gateway/model_build_targetgroup.go
```diff
@@ -172,65 +164,82 @@ func (t *latticeServiceModelBuildTask) buildTargets(ctx context.Context) error {
...
-			K8SServiceNamespace: t.serviceExport.Namespace,
-			Protocol:            "HTTP",
-			ProtocolVersion:     vpclattice.TargetGroupProtocolVersionHttp1,
+			Port:                  80,
+			IsServiceImport:       false,
+			IsServiceExport:       true,
+			K8SServiceName:        t.serviceExport.Name,
+			K8SServiceNamespace:   t.serviceExport.Namespace,
+			K8SHTTPRouteName:      "", // When build TG model for serviceExport, the httproutename should be ""
+			K8SHTTPRouteNamespace: "",
```
zijun726911, line 207: Explicitly show that  When build TG model for serviceExport, the httproutename should be "",  this is imporment, if we are not aware that it can cause some bug that hard to figure out. I will try to imporve our code to remove such "implicit convention"

pkg/gateway/model_build_targetgroup.go
```diff
@@ -72,47 +73,38 @@ func (b *targetGroupBuilder) Build(ctx context.Context, srvExport *mcs_api.Servi
...
+	if err := task.buildModel(ctx); err != nil {
 		return task.stack, task.targetGroup, corev1.ErrIntOverflowGenerated
 	}
 
 	return task.stack, task.targetGroup, nil
 }
 
-// for serviceexport
-func (t *targetGroupModelBuildTask) run(ctx context.Context) error {
-	/*
```
zijun726911, line 84: Removed this small wrapper function `run()`

pkg/latticestore/latticestore.go
zijun726911: Current we use this way to delete a tg cache record that this k8sSerivce backendref in a httproute: 			

`t.latticeDataStore.DelTargetGroup(resTargetGroup.Spec.Name, resTargetGroup.Spec.Config.K8SHTTPRouteName, false)`


we use this way to delete a tg cache record for which this serviceImport backendref in a httproute: 
`t.latticeDataStore.DelTargetGroup(resTargetGroup.Spec.Name, "", true)`		


And we use this way to delete a tg cache record for serviceExport: 
`t.latticeDataStore.DelTargetGroup(resTargetGroup.Spec.Name, "", false)`		


I think these functions are really confusing, we should improve them

pkg/deploy/lattice/target_group_synthesizer.go
```diff
@@ -423,7 +422,7 @@ func (t *targetGroupSynthesizer) SynthesizeTriggeredTargetGroupsDeletion(ctx con
 
 		if resTargetGroup.Spec.Config.IsServiceImport {
 			glog.V(2).Infof("Deleting service import target group from local datastore %v", resTargetGroup.Spec.LatticeID)
-			t.latticeDataStore.DelTargetGroup(resTargetGroup.Spec.Name, resTargetGroup.Spec.Config.K8SHTTPRouteName, resTargetGroup.Spec.Config.IsServiceImport)
+			t.latticeDataStore.DelTargetGroup(resTargetGroup.Spec.Name, "", resTargetGroup.Spec.Config.IsServiceImport)
 		} else {
```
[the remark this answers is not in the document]
mikhail-aws: Why dont we fix lattice data store at first place? wont it solve bug too?

pkg/gateway/model_build_targetgroup.go
```diff
@@ -172,65 +164,82 @@ func (t *latticeServiceModelBuildTask) buildTargets(ctx context.Context) error {
...
-			K8SServiceName:      t.serviceExport.Name,
-			K8SServiceNamespace: t.serviceExport.Namespace,
-			Protocol:            "HTTP",
-			ProtocolVersion:     vpclattice.TargetGroupProtocolVersionHttp1,
+			Port:                  80,
+			IsServiceImport:       false,
+			IsServiceExport:       true,
+			K8SServiceName:        t.serviceExport.Name,
+			K8SServiceNamespace:   t.serviceExport.Namespace,
+			K8SHTTPRouteName:      "", // When build TG model for serviceExport, the httproutename should be ""
```
mikhail-aws, line 206: we should wrap these into different constructor functions, that prevents to pass wrong arguments into struct

pkg/gateway/model_build_targetgroup.go
```diff
@@ -172,65 +164,82 @@ func (t *latticeServiceModelBuildTask) buildTargets(ctx context.Context) error {
...
-		// triggered by serviceexport delete
-		t.Datastore.SetTargetGroupByServiceExport(tgName, false, false)
-	} else {
-		// triggered by serviceexport add
-		t.Datastore.SetTargetGroupByServiceExport(tgName, false, true)
-	}
+	})
+	glog.V(6).Infof("buildTargetGroup, tg[%s], tgSpec%v \n", targetGroupName, tg)
+	// add targetgroup to Datastore local cache for service reconcile to reference
+	// for serviceexport, the httproutename is ""
```
mikhail-aws, line 214: please cleanup redundant comments that repeat code below

pkg/gateway/model_build_targetgroup.go
```diff
@@ -172,65 +164,82 @@ func (t *latticeServiceModelBuildTask) buildTargets(ctx context.Context) error {
...
+	t.Datastore.AddTargetGroup(targetGroupName, "", "", "", tg.Spec.Config.IsServiceImport, "")
```
mikhail-aws, line 215: we need to refactor datastore that it does not except wrong arguments, especially when we expect some of them being empty strings only

liwenwu-amazon:
I would recommend to have 2 PRs for this, 
* 1st one just to fix [one-line bug](https://github.com/aws/aws-application-networking-k8s/blob/main/pkg/gateway/model_build_targetgroup.go#L186),  where it should not return error if it is delete and k8s service not found
* 2nd one do what ever refactoring you think it is necessary

Right now, it is very hard to see what you are trying to fix. or you are trying to improve

pkg/gateway/model_build_targetgroup.go
```diff
@@ -72,47 +73,38 @@ func (b *targetGroupBuilder) Build(ctx context.Context, srvExport *mcs_api.Servi
...
+	err := t.buildTargetGroupForServiceExport(ctx)
 
 	if err != nil {
 		glog.V(6).Infof("Failed to build TargetGroup when serviceExport buildModel for name %v namespace %v \n", t.serviceExport.Name, t.serviceExport.Namespace)
 		return err
 	}
+	if !t.serviceExport.DeletionTimestamp.IsZero() {
+		//for serviceExport deletion request, we don't need to build targets model
+		return nil
+	}
```
zijun726911, line 94: for serviceExport deletion request, we don't need to build targets model because the target_group_manager could handle the deregistering targets 
https://github.com/aws/aws-application-networking-k8s/blob/a8eefa1d4c99c6de030f0a6264b5e1a6b10d1b2e/pkg/deploy/lattice/target_group_manager.go#L201

pkg/gateway/model_build_targetgroup.go
```diff
@@ -260,7 +267,13 @@ func (t *latticeServiceModelBuildTask) buildTargetGroup(ctx context.Context, cli
...
 			if t.route.DeletionTimestamp().IsZero() {
 				// to add
-				t.Datastore.SetTargetGroupByBackendRef(tgName, t.route.Name(), tgSpec.Config.IsServiceImport, true)
+				if *httpBackendRef.Kind() == "Service" {
+					t.Datastore.SetTargetGroupByBackendRef(tgName, t.route.Name(), tgSpec.Config.IsServiceImport, true)
+				} else if *httpBackendRef.Kind() == "ServiceImport" {
+					//When build targetGroup model for service Import byBackendRef
+					t.Datastore.SetTargetGroupByBackendRef(tgName, "", true, true)
+					t.Datastore.SetTargetGroupByBackendRef(tgName, "", false, true)
+				}
```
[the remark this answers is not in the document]
mikhail-aws: I appreciate comments you put around the code, but it should be either explicit in code, or these comments should be in place in source code describing public methods. Comments will be gone with PR, but logic will remain and keep confuse users.

zijun726911:
> I would recommend to have 2 PRs for this,
> 
>     * 1st one just to fix [one-line bug](https://github.com/aws/aws-application-networking-k8s/blob/main/pkg/gateway/model_build_targetgroup.go#L186),  where it should not return error if it is delete and k8s service not found
> 
>     * 2nd one do what ever refactoring you think it is necessary
> 
> 
> Right now, it is very hard to see what you are trying to fix. or you are trying to improve

I once tried to fix the one line fix (or keep diff as little as possible), but it actually didn't work, the reason it don't work is: we show pass the correct LatticeID to the target_group_manage to delete that TG, however I think we didn't assign the correct tgId in cache when we create cache record when we create ServiceExport. I need to fix the way how we add/update/delete latticeStore as well

mikhail-aws (changes requested):
I think we should refactor datastore first, it's hard to follow existing logic. Also please add description to PR page why it's happening. 

As I understand we pass non-empty string to data store that cause datastore return error, and we didnt build model. But when we create datastore do we pass non-empty string too? Why miss-match happens on deletion?

pkg/gateway/model_build_targetgroup.go
zijun726911: I think for create/ update serviceExport, the t.serviceExport.DeletionTimestamp.IsZero() == true? So update serviceExport can update the target?

For delete serviceExport t.serviceExport.DeletionTimestamp.IsZero() == false and we be sure need to de-register all targets for that tg and delelete that target group?

solmonk:
Please minimize the refactoring unless it is absolutely necessary. Also create issues instead of TODO comments which makes the diff too big

pkg/latticestore/latticestore.go
solmonk: please remove these TODO comments and create an issue instead

pkg/gateway/model_build_targetgroup.go
zijun726911: Talk with @solmonk , current datastore indeed hard to understand and need refactoring, but we should do it after GRPCRoute and TargetGroupPolicy delivered.  


And @solmonk also suggest to fix code diff as small as possible

What I am think is:
For Scenario 1: stand-alone serviceExport deletion leak tg, that can be fix by few lines change, but I am not 100% sure need to double confirm)
And For scenario 2: non-stand-alone serviceExport deletion leak tg (non-stand-alone means the serviceImport of that serviceExport referenced in a httproute), I think it need to change where we add/update/delete datastore targetGroup cache, only refactoring datastore could make that code easy to follow.

My next step would be try to use a seperate PR to fix Scenario 1 only(keep code diff as small as possible), and keep current PR open until we delivered the GRPCRoute and TargetGroupPolicy 

If quick fix for Scenario 1, I will just go ahead to work on something else

mikhail-aws:
still working on it?

zijun726911:
> still working on it?

Working on another PR to only solve standalone serviceExport deletion leak tg, that one will have smaller diff and more easy to review

zijun726911:
Partial fix for that bug: https://github.com/aws/aws-application-networking-k8s/pull/389

For the other part of fix (ServiceExport's serviceImport is referenced in the route), I prefer to wait until the target group datastore refactoring, maybe from @erikfuller ?

Issue #304, opened 2023-07-31
Support backend TLS connection


solmonk:
currently a connection from gateway to upstream is [plaintext HTTP by default, and users have no control over it](https://github.com/aws/aws-application-networking-k8s/blob/38f2b672595c1d68254b7f750cabbde2f528f4e6/pkg/gateway/model_build_targetgroup.go#L169-L198).

TLS re-encryption is not a part of gateway API spec yet but there is [an active discussion](https://github.com/kubernetes-sigs/gateway-api/issues/1897) going on, and we are observing a big user interest in this feature. In the meantime we could support this by extending the spec.

Issue #69, opened 2022-12-09
Controller need to configure target group health check


liwenwu-amazon:
Today,  lattice controller does `not` configure any health check for target group. So, target group uses lattice default configuration.  This will be `problem` and Pod stay in `Unhealthy` if. Pod does `NOT` respond to "/"  HTTP GET 

```
aws vpc-lattice get-target-group --target-group-identifier tg-028db3d23f0c48565 --endpoint-url=https://vpc-lattice.us-west-2.amazonaws.com
{
    "arn": "arn:aws:vpc-lattice:us-west-2:694065802095:targetgroup/tg-028db3d23f0c48565",
    "config": {
        "healthCheck": {
            "enabled": true,
            "healthCheckIntervalSeconds": 30,
            "healthCheckTimeoutSeconds": 5,
            "healthyThresholdCount": 5,
            "matcher": {
                "httpCode": "200"
            },
            "path": "/",
            "protocol": "HTTP",
            "unhealthyThresholdCount": 2
        },
        "port": 8090,
        "protocol": "HTTP",
        "protocolVersion": "HTTP1",
        "vpcIdentifier": "vpc-053e8d083fabe4780"
    },
    "createdAt": "2022-12-01T14:40:44.808000+00:00",
    "id": "tg-028db3d23f0c48565",
    "lastUpdatedAt": "2022-12-01T14:40:44.808000+00:00",
    "name": "k8s-nov28-parking-ver1-default",
    "serviceArns": [
        "arn:aws:vpc-lattice:us-west-2:694065802095:service/svc-086ea32d9c319060b"
    ],
    "status": "ACTIVE",
    "type": "IP"
}
```

PR #349, opened 2023-08-23
Watch TargetGroupPolicy CRD event on controllers

solmonk:
**What type of PR is this?**

feature

Related to #69 and #304 but not fixing the issue yet.

* This is the first part of TargetGroupPolicy change PR, just "watching" the event of TargetGroupPolicy change. Splitting out actual model changes because it is a lot bigger.
* The event handler is written a bit differently from other handlers, using controller library construct `handler.EnqueueRequestsFromMapFunc`. This clearly indicates the resource relationship, and enables easier unit testing IMO.
* The watch is happening only for Route/ServiceExport controllers which actually involve target group operations. 
* Also brings back DNSEndpoint watcher which got lost during route controller merge.

* `make e2etest` passing
* Added event handler unit tests as well - the controller package is now part of `make test`

**Automation added to e2e**:

```release-note

```


controllers/eventhandlers/targetgrouppolicy.go
```diff
@@ -0,0 +1,149 @@
...
+		return nil
+	}
+
+	svcName := types.NamespacedName{
+		Namespace: namespace,
+		Name:      string(targetRef.Name),
+	}
+	svc := &corev1.Service{}
+	if err := h.client.Get(context.TODO(), svcName, svc); err != nil {
+		return nil
```
mikhail-aws, line 49: should return *Service and error? it seems nil can be a good "nothing to return" and bad "error"

controllers/serviceexport_controller.go
```diff
@@ -179,12 +182,24 @@ func (r *ServiceExportReconciler) buildAndDeployModel(ctx context.Context, srvEx
...
 		For(&mcs_api.ServiceExport{}).
 		Watches(&source.Kind{Type: &corev1.Service{}}, svcEventsHandler).
-		Complete(r)
+		Watches(&source.Kind{Type: &v1alpha1.TargetGroupPolicy{}}, handler.EnqueueRequestsFromMapFunc(tgpEventHandler.MapToServiceExport))
+
+	if ok, err := k8s.IsGVKSupported(mgr, "application-networking.k8s.aws/v1alpha1", "TargetGroupPolicy"); ok {
+		builder.Watches(&source.Kind{Type: &v1alpha1.TargetGroupPolicy{}}, handler.EnqueueRequestsFromMapFunc(tgpEventHandler.MapToServiceExport))
+	} else {
+		if err != nil {
+			log.Infof("TargetGroupPolicy CRD is not found due to error: %s", err.Error())
```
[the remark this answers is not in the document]
solmonk: no, I think controller should still work if the CRD is not found

controllers/eventhandlers/targetgrouppolicy.go
solmonk: * The function's type is restricted to k8s-provided `handler.MapFunc` so that we cannot handle errors without panics
* I think it is better to gracefully handle the error and just skip the event. Instead we can put error-level logs here.

controllers/serviceexport_controller.go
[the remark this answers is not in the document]
solmonk: nvm actually I think terminal error here makes sense, will change

controllers/eventhandlers/targetgrouppolicy.go
erikfuller: > The function's type is restricted to k8s-provided handler.MapFunc so that we cannot handle errors without panics

I don't think ```getTargetRef``` is subject to this restriction? I see how the ```Map*``` functions would be.

controllers/eventhandlers/targetgrouppolicy.go
```diff
@@ -0,0 +1,158 @@
...
+	policyName := tgp.Namespace + "/" + tgp.Name
+
+	targetRef := tgp.Spec.TargetRef
+	if targetRef.Group != "" || targetRef.Kind != "Service" {
+		h.log.Infow("Detected non-Service TargetGroupPolicy attachment, skipping",
+			"policyName", policyName, "targetRef", targetRef)
+		return nil
+	}
+	namespace := tgp.Namespace
+	if targetRef.Namespace != nil && namespace != string(*targetRef.Namespace) {
```
erikfuller, line 38: what if ```targetRef.Namespace``` is ```nil```? Is that fine?

controllers/eventhandlers/targetgrouppolicy.go
```diff
@@ -0,0 +1,158 @@
...
+
+func NewTargetGroupPolicyEventHandler(log gwlog.Logger, client client.Client) *targetGroupPolicyEventHandler {
+	return &targetGroupPolicyEventHandler{log: log, client: client}
+}
+
+func (h *targetGroupPolicyEventHandler) getTargetRef(obj client.Object) *corev1.Service {
+	tgp := obj.(*v1alpha1.TargetGroupPolicy)
+	policyName := tgp.Namespace + "/" + tgp.Name
+
+	targetRef := tgp.Spec.TargetRef
```
erikfuller, line 31: Can this be ```nil```?

examples/target-group-policy.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: application-networking.k8s.aws/v1alpha1
+kind: TargetGroupPolicy
+metadata:
+    name: test-policy
+spec:
+    targetRef:
+        group: ""
```
erikfuller, line 7: If we leave the ```group``` field out does applying the YAML complain or does it simply set group to ""?

controllers/eventhandlers/targetgrouppolicy.go
```diff
@@ -0,0 +1,158 @@
...
+	if targetRef.Group != "" || targetRef.Kind != "Service" {
```
erikfuller, line 32: Would it be worth separating the ```Group``` condition and the ```Kind``` condition? Given this is attached to our TargetGroupPolicy CRD, is having values we don't support even a valid use case?

controllers/eventhandlers/targetgrouppolicy.go
erikfuller: I guess the result is the same, we return nil from our mapping function

controllers/eventhandlers/targetgrouppolicy.go
```diff
@@ -0,0 +1,158 @@
...
+	requests = append(requests, reconcile.Request{
+		NamespacedName: svcName,
+	})
+	return requests
+}
+
+func isServiceUsedByRoute(route core.Route, svc *corev1.Service) bool {
+	for _, rule := range route.Spec().Rules() {
+		for _, backendRef := range rule.BackendRefs() {
+			if backendRef.Kind() != nil && string(*backendRef.Kind()) != "Service" {
```
erikfuller, line 140: Don't we require ```Kind()``` to be ```"Service"```? Here it can be ```nil``` and still match.

controllers/eventhandlers/targetgrouppolicy.go
```diff
@@ -0,0 +1,158 @@
...
+	if err := h.client.Get(context.TODO(), svcName, svcExport); err != nil {
+		return nil
+	}
+	requests = append(requests, reconcile.Request{
+		NamespacedName: svcName,
+	})
+	return requests
+}
+
+func isServiceUsedByRoute(route core.Route, svc *corev1.Service) bool {
```
erikfuller, line 137: I'd say the approach of "eliminate all reasons not to match" is a bit of an anti-pattern that will likely allow things to fall through and return true that you don't want. Unless you really want "everything but these things" to return true, this logic should probably change to "here are the exact conditions I want to return true".

controllers/eventhandlers/targetgrouppolicy.go
```diff
@@ -0,0 +1,158 @@
...
+	if svc == nil {
+		return nil
+	}
+	svcName := types.NamespacedName{
+		Namespace: svc.Namespace,
+		Name:      svc.Name,
+	}
+	svcExport := &mcs_api.ServiceExport{}
+	if err := h.client.Get(context.TODO(), svcName, svcExport); err != nil {
+		return nil
```
erikfuller, line 129: In this case we would expect the ```.Get``` to succeed, right? is it worth logging here or is this actually expected?

controllers/eventhandlers/targetgrouppolicy.go
```diff
@@ -0,0 +1,158 @@
...
+
+	return svc
+}
+
+func (h *targetGroupPolicyEventHandler) MapToHTTPRoute(obj client.Object) []reconcile.Request {
+	var requests []reconcile.Request
+
+	svc := h.getTargetRef(obj)
+	if svc == nil {
+		return nil
```
erikfuller, line 71: Is there an external retry mechanisms for cases where we've errored internally and returned ```nil```?

erikfuller (changes requested):
Thanks for this PR, and apologies for all the questions - I'm still ramping up on the conventions and the domain.

controllers/eventhandlers/targetgrouppolicy.go
solmonk: in that case we assume that the targetRef has the same namespace.

controllers/eventhandlers/targetgrouppolicy.go
solmonk: per [gwapi spec](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.BackendObjectReference) - Kind defaults to Service when unspecified

controllers/eventhandlers/targetgrouppolicy.go
solmonk: will add check for it, thanks

examples/target-group-policy.yaml
solmonk: it will simply set group to empty string, btw empty group means "core" in k8s which is a common convention

controllers/serviceexport_controller.go
```diff
@@ -179,12 +182,23 @@ func (r *ServiceExportReconciler) buildAndDeployModel(ctx context.Context, srvEx
...
+func (r *ServiceExportReconciler) SetupWithManager(log gwlog.Logger, mgr ctrl.Manager) error {
 	svcEventsHandler := eventhandlers.NewEqueueRequestServiceEvent(r.Client)
-	return ctrl.NewControllerManagedBy(mgr).
+	tgpEventHandler := eventhandlers.NewTargetGroupPolicyEventHandler(log, r.Client)
+	builder := ctrl.NewControllerManagedBy(mgr).
 		// Uncomment the following line adding a pointer to an instance of the controlled resource as an argument
 		For(&mcs_api.ServiceExport{}).
 		Watches(&source.Kind{Type: &corev1.Service{}}, svcEventsHandler).
-		Complete(r)
+		Watches(&source.Kind{Type: &v1alpha1.TargetGroupPolicy{}}, handler.EnqueueRequestsFromMapFunc(tgpEventHandler.MapToServiceExport))
```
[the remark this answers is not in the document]
solmonk: ouch, I should remove this

controllers/serviceexport_controller.go
```diff
@@ -179,12 +182,23 @@ func (r *ServiceExportReconciler) buildAndDeployModel(ctx context.Context, srvEx
...
+
+	if ok, err := k8s.IsGVKSupported(mgr, "application-networking.k8s.aws/v1alpha1", "TargetGroupPolicy"); ok {
```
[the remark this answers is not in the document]
solmonk: Makes sense, I think I could just codegen those constants

controllers/eventhandlers/targetgrouppolicy.go
solmonk: yep it is worth logging here. will fix

controllers/eventhandlers/targetgrouppolicy.go
solmonk: makes sense, will try to improve this

controllers/eventhandlers/targetgrouppolicy.go
solmonk: I don't think it is possible from what k8s provides, this applies for any other event handlers we have. I think our best bet is to log properly and users will hopefully notice and recreate resources

controllers/eventhandlers/targetgrouppolicy.go
solmonk: want to be clear on that we want to support this particular resource groupkind. the struct is from gwapi so we cannot enforce this by yaml validation

controllers/eventhandlers/targetgrouppolicy.go
erikfuller: The wording is kind of ambiguous. Does it mean if it's not specified the value will be set to "Service", or does it mean if it's not specified it should be interepreted as "Service". I would expect the former, but I can't find code that says so. I tried applying some config that didn't have the field specified, and it *looks* like it's being set to "Service".

My interpretation I think is that it actually should never be nil. I think for the purposes of our code, we should ensure it exists and equals "Service"

solmonk: yup it seems you're right. it should never be nil as it is defined as a default value in CRD. Fixed the logic and test case.

controllers/eventhandlers/targetgrouppolicy.go
solmonk: Yes, but we intentionally swallow error here so I prefer not returning it (just to swallow later). we have more information inside and we can see targetRef related logs put in the same place

erikfuller (approved):
Changes look good, thanks for all the explanations. Will also give @mikhail-aws a chance to review before merging.

PR #326, opened 2023-08-11
feat: create targets if backendRef svc port match

rmullinnix461332:
**What type of PR is this?**
bug

**Which issue does this PR fix**:
#241 

**What does this PR do / Why do we need it**:
it only creates targets if the backend ref port or service export port matches the service port.  Matches all ports if no port specified

**If an issue # is not available please add repro steps and logs from aws-gateway-controller showing the issue**:

**Testing done on this change**:

test cases.  no access to e2e test automation.  Added new test cases
**Automation added to e2e**:

no
**Will this PR introduce any new dependencies?**:

no
**Will this break upgrades or downgrades. Has updating a running cluster been tested?**:
no
**Does this PR introduce any user-facing change?**:

no
```release-note
fix: only creates targets if the backend ref port or service export port matches the service port
```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #350, opened 2023-08-23
Enable GRPCRoute Traffic + Remaining GRPCRoute Event Handling

xWink:
**What type of PR is this?**
feature, cleanup

#25 

- Previously, Lattice target groups were always created with an HTTP1 protocol version
  - Now, we check what type of k8s route is used (HTTPRoute or GRPCRoute), and conditionally set the protocol version to HTTP1 or gRPC
- Previously, TargetGroupSynthesizer only checked if target groups were referenced by HTTPRoutes, which resulted in target groups referenced by only GRPCRoutes being marked as stale, and the controller would try to delete them
  - Now, TargetGroupSynthesizer knows to check either HTTPRoutes or GRPCRoutes and only marks target groups stale if they aren't referenced at all
- Previously, Gateway Reconciler, Gateway Event Handler, Service Event Handler, and ServiceImport Event Handler would only update impacted HTTPRoutes
  - Now, they also update impacted GRPCRoutes
-  Previously, `grpc-route.yaml` had an incorrectly configured match for the Greeter service and incorrect sectionName to allow traffic
   - Now, the match and sectionName are correctly set to allow traffic through
- Previously, all Route abstraction logic was in a single file, which was approaching 1000 lines long
  - Now, this has been split into `route.go`, `httproute.go`, and `grpcroute.go`
  - New route types can be added to new files for easier reading
- Previously, many files were logging with glog
  - Now, they log with gwlog
- General code and logging cleanup 

#### Note
ServiceExport support is not included here, as it will require a massive refactor to how we store all target groups in the cache. @solmonk is working on something like this already, so he said he will own that refactor.

1. `kubectl apply -f examples/my-hotel-gateway-multi-listeners.yaml`
1. `kubectl apply -f examples/greeter-grpc-server.yaml`
1. `kubectl apply -f examples/greeter-grpc-route.yaml`
1. `kubectl get pods -A`
1. Copy the name of a pod running the greeter-grpc-server
1. `kubectl exec -it <name-of-grpc-server-pod> -- bash`
1. Run this command to make our test client

```
cat << EOF > test.go
package main

import (
	"crypto/tls"
	"log"
	"os"

	"golang.org/x/net/context"
	"google.golang.org/grpc"
	"google.golang.org/grpc/credentials"
	pb "google.golang.org/grpc/examples/helloworld/helloworld"
)

func main() {
	if len(os.Args) < 3 {
		log.Fatalf("Usage: %s <address> <port>", os.Args[0])
	}

	address := os.Args[1] + ":" + os.Args[2]

	// Create a connection with insecure TLS (no server verification).
	creds := credentials.NewTLS(&tls.Config{
		InsecureSkipVerify: true,
	})
	conn, err := grpc.Dial(address, grpc.WithTransportCredentials(creds))
	if err != nil {
		log.Fatalf("did not connect: %v", err)
	}
	defer conn.Close()
	c := pb.NewGreeterClient(conn)

	// Contact the server and print out its response.
	name := "world"
	if len(os.Args) > 3 {
		name = os.Args[3]
	}
	r, err := c.SayHello(context.Background(), &pb.HelloRequest{Name: name})
	if err != nil {
		log.Fatalf("could not greet: %v", err)
	}
	log.Printf("Greeting: %s", r.Message)
}
EOF

```

8. `go run test.go  <SERVICE DNS> <PORT>`
  - Make sure to replace \<SERVICE DNS\> with the actual Lattice service DNS
  - Make sure to replace \<PORT\> with the port your Lattice listener uses (443 by default)
  - Expected output: `Greeting: Hello world`

- E2e tests pass and manual steps shown above also pass.
- Match for greeter-grpc-route was edited to not include the method, so it matches any method in the service (in Lattice, this should show up as a rule with Prefix on `/helloworld.Greeter/`)
- Rules for greeter-grpc-route were edited to look like below (in Lattice, this should show up as 2 rules, each matching against the corresponding service and method)
```
rules:
    - matches:
        - method:
            service: helloworld.Greeter
            method: SayHello
      backendRefs:
        - name: grpc-server
          kind: Service
          port: 50051
          weight: 10
    - matches:
        - method:
            service: helloworld.Greeter2
            method: SayHello2
      backendRefs:
        - name: grpc-server
          kind: Service
          port: 50051
          weight: 10
```

**Automation added to e2e**:

This will be done separately

No

No

Yes

```release-note
Enabled traffic for GRPCRoutes
Logging cleanup
```


cmd/aws-application-networking-k8s/main.go
```diff
@@ -151,20 +151,14 @@ func main() {
...
 	}
 
-	serviceImportReconciler := controllers.NewServceImportReconciler(mgr.GetClient(), mgr.GetScheme(),
-		mgr.GetEventRecorderFor("ServiceImport"), finalizerManager, latticeDataStore)
-
-	if err = serviceImportReconciler.SetupWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create controller", "controller", "ServiceImport")
-		os.Exit(1)
+	err = controllers.RegisterServiceImportReconciler(ctrlLog.Named("serviceimport"), mgr, latticeDataStore, finalizerManager)
+	if err != nil {
```
zijun726911, line 155: Should we unify the name  `Reconciler` and `Controller` to only use name `Controller`? (Maybe in a separate PR)

controllers/eventhandlers/service.go
```diff
@@ -14,43 +13,44 @@ import (
...
 }
 
-func (h *enqueueRequetsForServiceEvent) Create(e event.CreateEvent, queue workqueue.RateLimitingInterface) {
-	glog.V(6).Info("Event: service create")
+func (h *enqueueRequestForServiceWithExportEvent) Create(e event.CreateEvent, queue workqueue.RateLimitingInterface) {
+	h.log.Info("Event: service create")
 	service := e.Object.(*corev1.Service)
 	h.enqueueImpactedService(queue, service)
 	h.enqueueImpactedServiceExport(queue, service)
 }
```
zijun726911, line 36: Question for old code and not related to current PR,

we did this in serviceexport_controller.go:
```
svcExportEventsHandler := eventhandlers.NewEqueueRequestServiceWithExportEvent(log, r.client)

	return ctrl.NewControllerManagedBy(mgr).
		For(&mcs_api.ServiceExport{}).
		Watches(&source.Kind{Type: &corev1.Service{}}, svcExportEventsHandler).
		Complete(r)
```

if the serviceexport_controller saw a k8sService created, does it need to do`h.enqueueImpactedServiceExport(queue, service)`? That can cause one ServiceExport creation request to `Reconcile` twice?

controllers/eventhandlers/service.go
```diff
@@ -14,43 +13,44 @@ import (
...
-type enqueueRequetsForServiceEvent struct {
+type enqueueRequestForServiceWithExportEvent struct {
+	log    gwlog.Logger
 	client client.Client
 }
 
-func NewEqueueRequestServiceEvent(client client.Client) handler.EventHandler {
-	return &enqueueRequetsForServiceEvent{
+func NewEqueueRequestServiceWithExportEvent(log gwlog.Logger, client client.Client) handler.EventHandler {
+	return &enqueueRequestForServiceWithExportEvent{
```
zijun726911, line 25: I think  both old name and new name  are confusing... Can we change to a name likes  `NewK8sServiceEventHandlerForServiceExportController`? (or do it in a separate PR)

controllers/eventhandlers/serviceimport.go
```diff
@@ -54,53 +56,60 @@ func (h *enqueueRequestsForServiceImportEvent) Generic(e event.GenericEvent, que
...
 }
 
-func isServiceImportUsedByHTTPRoute(httpRoute gateway_api.HTTPRoute, serviceImport *mcs_api.ServiceImport) bool {
-	for _, httpRule := range httpRoute.Spec.Rules {
-		for _, httpBackendRef := range httpRule.BackendRefs {
-			if string(*httpBackendRef.BackendObjectReference.Kind) != "serviceimport" {
+func isServiceImportUsedByRoute(route core.Route, serviceImport *mcs_api.ServiceImport) bool {
+	for _, rule := range route.Spec().Rules() {
+		for _, backendRef := range rule.BackendRefs() {
+			if string(*backendRef.Kind()) != "serviceimport" {
```
zijun726911, line 94: Should be
`if string(*backendRef.Kind()) != "Serviceimport" {`

Someone mentioned this bug, could you fix in passing?

https://github.com/aws/aws-application-networking-k8s/pull/326/files

cmd/aws-application-networking-k8s/main.go
xWink: This is my bad, will fix in next commit!

liwenwu-amazon:
Can you add a  grpc_xxx.md in the doc directory?   You can put similar information you had in the CR description for now.   So other folks can just follow that grpc_xxx.md to manually test e2e grpc functionalities.  thanks

controllers/gateway_controller.go
```diff
@@ -476,5 +447,33 @@ func UpdateGWListenerStatus(ctx context.Context, k8sclient client.Client, gw *ga
...
+		} else {
+			validRoute = false
+		}
+	}
+
+	if validRoute {
+		return true, supportedKinds
+	} else {
+		return false, supportedKinds
+	}
```
mikhail-aws, line 478: - Guessing from code expected behavior - if list is empty or contains unsupported type return false, am I right? It would be helpful if you document intent in the function description
- Please add unit test
- nit: return arguments order: result comes first and bool-ok second (flip them)

pkg/model/core/httproute.go
```diff
@@ -0,0 +1,317 @@
...
+
+type HTTPRoute struct {
+	r gateway_api_v1beta1.HTTPRoute
+}
+
+func NewHTTPRoute(route gateway_api_v1beta1.HTTPRoute) *HTTPRoute {
+	return &HTTPRoute{r: route}
+}
+
+func GetHTTPRoute(client client.Client, ctx context.Context, routeNamespacedName types.NamespacedName) (Route, error) {
```
mikhail-aws, line 24: context always first argument

pkg/deploy/lattice/target_group_synthesizer.go
```diff
@@ -229,71 +237,71 @@ func (t *targetGroupSynthesizer) SynthesizeSDKTargetGroups(ctx context.Context)
...
 
-			httpRoute := &gateway_api.HTTPRoute{}
-
-			tgName := latticestore.TargetGroupName(*srvName, *srvNamespace)
-
-			if err := t.client.Get(ctx, httprouteName, httpRoute); err != nil {
-				glog.V(6).Infof("tgname %v is not used by httproute %v\n", tgName, httpRoute)
-
+			var route core.Route
+			if *sdkTG.getTargetGroupOutput.Config.ProtocolVersion == vpclattice.TargetGroupProtocolVersionGrpc {
```
mikhail-aws, line 275: I would use switch in places where multiple versions supported

pkg/gateway/model_build_lattice_service.go
```diff
@@ -23,29 +24,39 @@ type LatticeServiceBuilder interface {
...
-
-	cloud lattice_aws.Cloud
+	cloud       lattice_aws.Cloud
 }
 
-func NewLatticeServiceBuilder(client client.Client, datastore *latticestore.LatticeDataStore, cloud lattice_aws.Cloud) *latticeServiceModelBuilder {
-	return &latticeServiceModelBuilder{
-		Client:    client,
+func NewLatticeServiceBuilder(
+	log gwlog.Logger,
```
mikhail-aws, line 36: thank you for propagating logger

pkg/gateway/model_build_listener.go
```diff
@@ -19,38 +17,39 @@ const (
 	awsCustomCertARN = "application-networking.k8s.aws/certificate-arn"
 )
 
-func (t *latticeServiceModelBuildTask) extractListenerInfo(ctx context.Context, parentRef gateway_api.ParentReference) (int64, string, string, error) {
-
-	var protocol gateway_api.ProtocolType = gateway_api.HTTPProtocolType
+func (t *latticeServiceModelBuildTask) extractListenerInfo(
+	ctx context.Context,
+	parentRef gateway_api.ParentReference,
+) (int64, string, string, error) {
```
mikhail-aws, line 23: need to wrap return arguments into struct

pkg/gateway/model_build_targetgroup.go
```diff
@@ -28,48 +27,56 @@ type TargetGroupModelBuilder interface {
...
+	log           gwlog.Logger
+	client        client.Client
 	serviceExport *mcs_api.ServiceExport
 	targetGroup   *latticemodel.TargetGroup
 	tgByResID     map[string]*latticemodel.TargetGroup
 	stack         core.Stack
-
-	Datastore *latticestore.LatticeDataStore
-	cloud     lattice_aws.Cloud
+	Datastore     *latticestore.LatticeDataStore
```
mikhail-aws, line 61: For consistency, I like datastore, but I see places where latticeDataStore is used, also it should be private property.

If it's not a burden can you align datastores in your PR in places where you change it. Not asking to go beyond that.

pkg/gateway/model_build_targetgroup.go
```diff
@@ -350,22 +358,27 @@ func (t *latticeServiceModelBuildTask) buildTargetGroupSpec(ctx context.Context,
 		}
 	}
 
-	tgName := latticestore.TargetGroupName(string(httpBackendRef.Name()), namespace)
+	tgName := latticestore.TargetGroupName(string(backendRef.Name()), namespace)
+
+	protocolVersion := vpclattice.TargetGroupProtocolVersionHttp1
+	if _, ok := t.route.(*core.GRPCRoute); ok {
+		protocolVersion = vpclattice.TargetGroupProtocolVersionGrpc
+	}
```
mikhail-aws, line 366: probably switch on route type with default to http1

pkg/gateway/model_build_targetgroup_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/aws/aws-application-networking-k8s/pkg/utils/gwlog"
```
mikhail-aws, line 7: please add imports sort and grouping, usually order std lib, 3ed party, local

 https://www.jetbrains.com/help/go/creating-and-optimizing-imports.html#use-the-goimports-local-grouping

pkg/model/core/grpcroute.go
```diff
@@ -0,0 +1,302 @@
...
+
+type GRPCRoute struct {
+	r gateway_api_v1alpha2.GRPCRoute
+}
+
+func NewGRPCRoute(route gateway_api_v1alpha2.GRPCRoute) *GRPCRoute {
+	return &GRPCRoute{r: route}
+}
+
+func GetGRPCRoute(client client.Client, ctx context.Context, routeNamespacedName types.NamespacedName) (Route, error) {
```
mikhail-aws, line 25: context first argument, always

pkg/model/core/grpcroute.go
```diff
@@ -0,0 +1,302 @@
...
+	err := client.Get(ctx, routeNamespacedName, grpcRoute)
+	if err != nil {
+		return nil, err
+	}
+	return NewGRPCRoute(*grpcRoute), nil
+}
+
+func ListGRPCRoutes(client client.Client, context context.Context) []Route {
+	routeList := &gateway_api_v1alpha2.GRPCRouteList{}
+	client.List(context, routeList)
```
mikhail-aws, line 36: I believe list can return an error, why drop it here?

pkg/model/core/grpcroute.go
```diff
@@ -0,0 +1,302 @@
...
+
+func (r *GRPCRouteRule) Matches() []RouteMatch {
+	var routeMatches []RouteMatch
+	for _, routeMatch := range r.r.Matches {
+		routeMatches = append(routeMatches, &GRPCRouteMatch{routeMatch})
+	}
+	return routeMatches
+}
+
+func (r *GRPCRouteRule) Equals(routeRule RouteRule) bool {
```
mikhail-aws, line 156: please add unit test, same for other methods with conditional logic, for example other "Equal" methods

mikhail-aws (changes requested):

mikhail-aws:
Thank you for adding new logger and cleaning up logs.

mikhail-aws:
47 commits in diff seems not accurate, looks like your fork main branch out of sync

pkg/gateway/model_build_lattice_service.go
zijun726911: One general question, when we should use [this logger](https://github.com/aws/aws-application-networking-k8s/blob/d8b2e73263dff6356148ba4f46aa1dd35289aae7/cmd/aws-application-networking-k8s/main.go#L90), and when we should use `gwlog.FallbackLogge`r?

zijun726911 (commented):
Overall LGTM, please address @mikhail-aws's comments

pkg/gateway/model_build_lattice_service.go
mikhail-aws: fallback is a backup plan when it's hard to propagate logger and you need something for debugging. Rather than using glog.v6 it's better to use FallbackLogger. You can create an alias in package for convenience 
```
var log = gwlog.FallbackLogger

func foo() {
 - glog.V(6).Info("bla")
 + log.Debug("bla")
}
```

controllers/eventhandlers/gateway.go
```diff
@@ -64,32 +65,29 @@ func (h *enqueueRequestsForGatewayEvent) Generic(e event.GenericEvent, queue wor
 
 }
 
-func (h *enqueueRequestsForGatewayEvent) enqueueImpactedHTTPRoute(queue workqueue.RateLimitingInterface, gw *gateway_api.Gateway) {
-	httpRouteList := &gateway_api.HTTPRouteList{}
+func (h *enqueueRequestsForGatewayEvent) enqueueImpactedRoutes(queue workqueue.RateLimitingInterface) {
+	routes := core.ListAllRoutes(h.client, context.TODO())
```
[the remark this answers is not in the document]
xWink: I am keeping the logic identical to before to avoid making too many behavioural changes

xWink:
> Can you add a grpc_xxx.md in the doc directory? You can put similar information you had in the CR description for now. So other folks can just follow that grpc_xxx.md to manually test e2e grpc functionalities. thanks

Will do this in a separate PR. Current focus is purely code change to unblock others.

controllers/eventhandlers/serviceimport.go
xWink: Will do, thanks for pointing it out. Found it in target group model builder too.

controllers/eventhandlers/serviceimport.go
```diff
@@ -54,53 +56,60 @@ func (h *enqueueRequestsForServiceImportEvent) Generic(e event.GenericEvent, que
...
+	var routes []core.Route
 
-	h.client.List(context.TODO(), httpRouteList)
+	switch h.routeType {
+	case core.HttpRouteType:
+		routes = core.ListHTTPRoutes(h.client, context.TODO())
+	case core.GrpcRouteType:
+		routes = core.ListGRPCRoutes(h.client, context.TODO())
+	default:
+		h.log.Errorf("Invalid routeType %s", h.routeType)
```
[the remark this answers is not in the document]
xWink: Done next commit

pkg/model/core/httproute.go
xWink: Fixed next commit

pkg/deploy/lattice/target_group_synthesizer.go
xWink: This is on purpose, as we default to http with the sole exception of grpc

pkg/gateway/model_build_listener.go
xWink: Would like to avoid changing too much now. While 3 return values is more than what we'd like, it's also not too obtrusive to read at the moment.

pkg/gateway/model_build_targetgroup.go
xWink: Yup, done next commit

pkg/gateway/model_build_targetgroup_test.go
xWink: Done next commit, thanks for the jetbrains link!

pkg/model/core/grpcroute.go
xWink: Done next commit

pkg/model/core/grpcroute.go
xWink: Fair point, fixed next commit!

xWink:
> 47 commits in diff seems not accurate, looks like your fork main branch out of sync

Most likely, though I will squash for the merge anyways. Will rebase after this to re-sync my local.

pkg/gateway/model_build_listener.go
mikhail-aws: Multi-return purpose is not for results that are suppose to be used together, it's anti-pattern of usage. 

You wouldn't see returning more than 3 values in golang projects. Most of the time it's (result, error), (result, bool), and rare occasions with 3 values for example create input/output channels for socket - createConnection() (in, out, error), so in/out can be used separately, likely in different go-routines.

Since it's old code, we can change it after this PR.

controllers/gateway_controller.go
xWink: From my understanding, this has no impact on controller functionality, and is only for correctness of some metadata (based on Doyoon's description). For the sake of time, let's skip this one for now.

pkg/gateway/model_build_targetgroup.go
xWink: Is that not the same thing?

pkg/model/core/grpcroute.go
xWink: Added a suite of unit tests for all route abstraction Equals methods

xWink:
Not sure why make presubmit fails here. It runs without issues on my end, and I have no file changes that aren't committed

```
kapshawn@88665a240c05 aws-application-networking-k8s % make presubmit
go run sigs.k8s.io/controller-tools/cmd/controller-gen@v0.13.0 object paths=./pkg/apis/...
go run sigs.k8s.io/controller-tools/cmd/controller-gen@v0.13.0 crd paths=./pkg/apis/... output:crd:artifacts:config=config/crds/bases
go run k8s.io/code-generator/cmd/register-gen@v0.28.0 --input-dirs ./pkg/apis/applicationnetworking/v1alpha1 --output-base ./ --go-header-file hack/boilerplate.go.txt
cp config/crds/bases/application-networking.k8s.aws* helm/crds
go mod tidy
go generate ./...
go vet ./...
go fmt ./...
go test ./pkg/... ./controllers/... -coverprofile coverage.out
?       github.com/aws/aws-application-networking-k8s/pkg/apis/applicationnetworking/v1alpha1   [no test files]
ok      github.com/aws/aws-application-networking-k8s/pkg/aws/services  0.649s  coverage: 4.4% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/config        0.323s  coverage: 38.3% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/deploy        0.737s  coverage: 20.6% of statements
?       github.com/aws/aws-application-networking-k8s/pkg/utils/gwlog   [no test files]
ok      github.com/aws/aws-application-networking-k8s/pkg/deploy/externaldns    1.176s  coverage: 77.8% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/deploy/lattice        0.981s  coverage: 86.3% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/gateway       1.457s  coverage: 79.1% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/latticestore  1.472s  coverage: 73.4% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/model/core    1.903s  coverage: 61.7% of statements
ok      github.com/aws/aws-application-networking-k8s/pkg/model/core/graph      1.908s  coverage: 17.2% of statements
ok      github.com/aws/aws-application-networking-k8s/controllers       0.668s  coverage: 0.0% of statements
ok      github.com/aws/aws-application-networking-k8s/controllers/eventhandlers 0.952s  coverage: 18.0% of statements
kapshawn@88665a240c05 aws-application-networking-k8s % git status
On branch 25
Your branch is up to date with 'origin/25'.
```

solmonk:
try removing changes of `client_mocks.go` (looks like you don't need it), it seems breaking CI

controllers/eventhandlers/service.go
```diff
@@ -88,85 +90,81 @@ func (h *enqueueRequetsForServiceEvent) enqueueImpactedServiceExport(queue workq
...
-	for _, httpRule := range httpRoute.Spec.Rules {
-		for _, httpBackendRef := range httpRule.BackendRefs {
-			//glog.V(6).Infof("isServiceUsedByHTTPRoute httpBackendRef %v, %v\n", httpBackendRef.BackendObjectReference, ep.Name)
-			if string(*httpBackendRef.BackendObjectReference.Kind) != "service" {
-				glog.V(6).Infof("isServiceUsedByHTTPRoute: kind %v\n", string(*httpBackendRef.BackendObjectReference.Kind))
+func isServiceUsedByRoute(route core.Route, ep *corev1.Service) bool {
+	for _, rule := range route.Spec().Rules() {
+		for _, backendRef := range rule.BackendRefs() {
+			if string(*backendRef.Kind()) != "service" {
 				continue
```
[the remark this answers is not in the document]
xWink: What kind of bugs would we be expecting if this is already wrong?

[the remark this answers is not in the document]
xWink: I'd like to keep this PR focused on the GRPC changes, as this is a blocker for other upcoming work. If this is a bug, let's create an issue to address and test it.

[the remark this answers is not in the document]
zijun726911: Ok, make sense, just keep it

mikhail-aws (approved):

solmonk (approved):

PR #351, opened 2023-08-24
Use right port on backendRef port match

solmonk:
**What type of PR is this?**

bug

#241 

Interpret backendRef port "correctly", it should be service port, not target port per [gwapi documentation](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.BackendObjectReference).

To do this, we need to figure out what is the right set of endpoint ports only based on the service port. We can deduce this by the following logic:

* [A service port MUST have a name if there are multiple ports exposed from a service](https://kubernetes.io/docs/concepts/services-networking/service/#multi-port-services). Therefore, if a port is named, endpoint port is only relevant if it has the same name.
* If a service port is unnamed, it MUST be the only port that is exposed from a service. In this case, as long as the service port is matching with backendRef/annotations, we can consider all endpoints valid

This is also commented in the code.

unit tests and e2e tests.

**Automation added to e2e**:
Fixed existing e2e test case, I just simplified it into two test scenarios that seems relevant to this use case.

```release-note

```

### DIFF
--- a/pkg/gateway/model_build_targets.go
+++ b/pkg/gateway/model_build_targets.go
@@ -24,7 +24,7 @@ import (
 const (
 	resourceIDLatticeTargets = "LatticeTargets"
 	portAnnotationsKey       = "multicluster.x-k8s.io/port"
-	undefinedPort            = int64(0)
+	undefinedPort            = int32(0)
 )
 
 type LatticeTargetsBuilder interface {
@@ -110,7 +110,7 @@ func (t *latticeTargetsModelBuildTask) buildLatticeTargets(ctx context.Context) 
 		return errors.New(errmsg)
 	}
 
-	definedPorts := make([]int64, 0)
+	definedPorts := make(map[int32]struct{})
 	if tg.ByServiceExport {
 		serviceExport := &mcs_api.ServiceExport{}
 		err = t.Client.Get(ctx, namespacedName, serviceExport)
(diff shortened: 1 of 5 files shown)


pkg/gateway/model_build_targets_test.go
```diff
@@ -65,25 +66,71 @@ func Test_Targets(t *testing.T) {
...
+			wantErrIsNil: true,
+			expectedTargetList: []latticemodel.Target{
+				{
+					TargetIP: "10.10.1.1",
 					Port:     8675,
 				},
 				{
 					TargetIP: "10.10.2.2",
-					Port:     309,
+					Port:     8675,
```
mikhail-aws, line 133: Can you explain please why it's 8675?

As I understand from code that if port is not named then it's the only port. Here we have 2 named ports "a" and "b" and both targets are generated with "a" port, where second port go?

solmonk: in line 87 - we pick the port 80, this is part of target model build task which is supposed to be filled in from route backendRef

pkg/gateway/model_build_targets.go
```diff
@@ -189,7 +206,7 @@ type latticeTargetsModelBuildTask struct {
 	tgName         string
 	tgNamespace    string
 	routename      string
-	port           int64
+	port           int32
 	latticeTargets *latticemodel.Targets
```
zijun726911, line 210: Could you change it's name to routeBackendRefPort ?

pkg/gateway/model_build_targets.go
```diff
@@ -156,14 +179,8 @@ func (t *latticeTargetsModelBuildTask) buildLatticeTargets(ctx context.Context)
...
 						Port:     int64(port.Port),
 					}
-
-					for _, definedPort := range definedPorts {
-						if target.Port == definedPort || definedPort == undefinedPort {
-							// target port matches httproute
-							targetList = append(targetList, target)
-						} else {
-							glog.V(6).Infof("Port does not match the target - port:%d, definedPort:%d, target:%v", target.Port, definedPort, target)
-						}
```
zijun726911, line 166: Could you write a comment here to say "endPoints' port should be corresponding k8sService's targetPort"

pkg/gateway/model_build_targets.go
```diff
@@ -120,20 +120,43 @@ func (t *latticeTargetsModelBuildTask) buildLatticeTargets(ctx context.Context)
...
+
+	for _, port := range svc.Spec.Ports {
+		if _, ok := definedPorts[port.Port]; ok {
+			if port.Name != "" {
+				servicePortNames[port.Name] = struct{}{}
+			} else {
+				// Unnamed, consider all endpoints valid
+				skipMatch = true
+			}
+		}
```
zijun726911, line 154: For httproute backendref k8sService scenario,
Say, one k8sService have only one unnamed (targetPort 8080, port 80)
And the httproute backendref have a port 81,
In that case we create a lattice target for port 8080, is that expected behavior

pkg/gateway/model_build_targets_test.go
mikhail-aws: should {Name: "b", Port: 309}} be in expected target list?

solmonk: no - by picking 80 in model build task, we know that only relevant service port is {Name: "a", Port: 80, TargetPort: 8675}. The endpoints with other service ports will be ignored.

pkg/gateway/model_build_targets.go
solmonk: Added a bit of explanation, hope it is clear now

pkg/gateway/model_build_targets.go
solmonk: Fixed, but backendRef already implies route so I prefer backendRefPort.

pkg/gateway/model_build_targets_test.go
solmonk: Not sure if you intended to put the same port, but it should not be possible by k8s spec

mikhail-aws (approved):

pkg/gateway/model_build_targets_test.go
zijun726911:
```
core1.ServicePort{ {Name: "a", Port: 80, Target: "8675"}, {Name: "b", Port: 80, Target: "309"} }

```
Seems in that scenario, only one (Port,TargetPort ) can take effect
image

zijun726911: core1.ServicePort{ {Name: "a", Port: 80, Target: "8675"}, {Name: "b", Port: 8090, Target: "309"} }

For this one it can have two (Port,TargetPort )
image

zijun726911 (approved):
LGTM

pkg/gateway/model_build_targets_test.go
solmonk: by k8s spec it is stored as a map with port as a key, probably that's why.

PR #356, opened 2023-08-25
add generic managedBy tag, fix deletion for not managed Sn-Service associations

mikhail-aws:
### Note:
- add ClusterName property to config
- create generic tag for resources that gateway should manage
- add helper methods for tags
- remove EKS sdk client, it's not used
- update logic for service_manager when we create and update service
  - when create, add default tags to association
  - when update, check managedBy tag before deleting association
 
### Issues:
- update #317 
- update #342  
- update #310 

### Tests:
- updated unit tests with tags validation
- manually tested `getClusterName` using metadata with deploying k8s pod (my cluster name is "cl")

```bash
kubectl run -it gateway --image {accountid-redacted}.dkr.ecr.us-west-2.amazonaws.com/cl-repo:gateway

{"level":"info","ts":1692932391.0858064,"logger":"setup","caller":"workspace/main.go:106","msg":"init config","VpcId":"vpc-redacted","Region":"us-west-2","AccoundId":"redacted","DefaultServiceNetwork":"","UseLongTgName":false,"ClusterName":"cl"}
```
- checked in console that tags are populated to Service and association during e2e tests
```
Key		Value
ManagedBy	{accountid-redacted}-cl-{vpcid-redacted}
```
- e2e tests, all pass except GRPC, GRPC is not fully merged yet, error is expected
```
Summarizing 1 Failure:
  [FAIL] GRPCRoute traffic test [It] Test GRPC traffic by k8s network only and don't use vpc lattice, only a demo for how to use grpc helper functions, this test case will be removed later
  /Users/mhlbrz/workplace/aws-application-networking-k8s/test/suites/integration/grpcroute_traffic_test.go:74

Ran 13 of 13 Specs in 1892.216 seconds
FAIL! -- 12 Passed | 1 Failed | 0 Pending | 0 Skipped
--- FAIL: TestIntegration (1893.30s)
```

### DIFF
--- a/pkg/aws/cloud.go
+++ b/pkg/aws/cloud.go
@@ -1,7 +1,7 @@
 package aws
 
 import (
-	"strings"
+	"fmt"
 
 	"github.com/aws/aws-application-networking-k8s/pkg/aws/services"
 	"github.com/aws/aws-application-networking-k8s/pkg/utils/gwlog"
@@ -11,7 +11,8 @@ import (
 )
 
 const (
-	TagManagedBy = "ManagedBy"
+	TagBase      = "application-networking.k8s.aws/"
+	TagManagedBy = TagBase + "ManagedBy"
 )
 
 type Tags = map[string]*string
@@ -36,7 +37,7 @@ type Cloud interface {
 	IsArnManaged(arn string) (bool, error)
 
 	// check if tags map has managedBy tag
-	IsManagedByTagSet(tags Tags) bool
+	ContainsManagedBy(tags Tags) bool
 }
 
 // NewCloud constructs new Cloud implementation.
@@ -97,7 +98,7 @@ func (c *defaultCloud) DefaultTags() Tags {
 	return tags
 }
 
-func (c *defaultCloud) IsManagedByTagSet(tags Tags) bool {
+func (c *defaultCloud) ContainsManagedBy(tags Tags) bool {
 	tag, ok := tags[TagManagedBy]
 	if !ok || tag == nil {
 		return false
@@ -111,20 +112,10 @@ func (c *defaultCloud) IsArnManaged(arn string) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	isManaged := c.IsManagedByTagSet(resp.Tags)
+	isManaged := c.ContainsManagedBy(resp.Tags)
 	return isManaged, nil
 }
 
 func getManagedByTag(cfg CloudConfig) string {
-	tagKeys := []string{}
-	if cfg.AccountId != "" {
-		tagKeys = append(tagKeys, cfg.AccountId)
-	}
-	if cfg.ClusterName != "" {
-		tagKeys = append(tagKeys, cfg.ClusterName)
-	}
-	if cfg.VpcId != "" {
-		tagKeys = append(tagKeys, cfg.VpcId)
-	}
-	return strings.Join(tagKeys, "-")
+	return fmt.Sprintf("%s/%s/%s", cfg.AccountId, cfg.ClusterName, cfg.VpcId)
 }
--- a/pkg/aws/cloud.go
+++ b/pkg/aws/cloud.go
@@ -1,26 +1,42 @@
 package aws
 
 import (
+	"strings"
+
 	"github.com/aws/aws-application-networking-k8s/pkg/aws/services"
-	"github.com/aws/aws-application-networking-k8s/pkg/config"
 	"github.com/aws/aws-application-networking-k8s/pkg/utils/gwlog"
-	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
-	"github.com/aws/aws-sdk-go/service/eks"
+	"github.com/aws/aws-sdk-go/service/vpclattice"
 )
+
+const (
+	TagManagedBy = "ManagedBy"
+)
+
+type Tags = map[string]*string
 
 //go:generate mockgen -destination cloud_mocks.go -package aws github.com/aws/aws-application-networking-k8s/pkg/aws Cloud
 
 type CloudConfig struct {
-	VpcId     string
-	AccountId string
+	VpcId       string
+	AccountId   string
+	Region      string
+	ClusterName string
 }
 
 type Cloud interface {
 	Config() CloudConfig
 	Lattice() services.Lattice
-	EKS() services.EKS
+
+	// creates lattice tags with default values populated
+	DefaultTags() Tags
+
+	// check if managedBy tag set for lattice resource
+	IsArnManaged(arn string) (bool, error)
+
+	// check if tags map has managedBy tag
+	IsManagedByTagSet(tags Tags) bool
 }
 
 // NewCloud constructs new Cloud implementation.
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -45,7 +45,7 @@ help: ## Display this help.
 
 .PHONY: run
 run: ## Run in development mode
-	go run cmd/aws-application-networking-k8s/main.go --debug -v 6
+	go run cmd/aws-application-networking-k8s/main.go --debug
```
xWink, line 48: Why remove the `-v 6`? This removes logs that are still necessary for debugging until we fully migrate to zap logger

pkg/config/controller_config.go
```diff
@@ -77,6 +82,12 @@ func configInit(metadata EC2Metadata) error {
 	// CLUSTER_LOCAL_GATEWAY
 	DefaultServiceNetwork = os.Getenv(CLUSTER_LOCAL_GATEWAY)
 
+	// CLUSTER_NAME
+	ClusterName, err = getClusterName(sess)
+	if err != nil {
+		return fmt.Errorf("cannot get cluster name: %s", err)
+	}
+
```
zijun726911, line 90: Does this `configInit` also work for ECS Fargate based  k8s cluster?

zijun726911:
I think the one GRPC e2etest should also work? currently it didn't test our controller logic or grpcRoute... let me check the mainline code to figure out the reason

pkg/config/controller_config.go
mikhail-aws: env or metadata same as other variables

pkg/config/controller_config.go
```diff
@@ -88,3 +99,38 @@ func configInit(metadata EC2Metadata) error {
 
 	return nil
 }
+
+// try to find cluster name, search in env then in ec2 instance tags
+// returns empty string if not found
+func getClusterName(sess *session.Session) (string, error) {
+	cn := os.Getenv(CLUSTER_NAME)
+	if cn != "" {
+		return cn, nil
```
mohakkohli, line 108: probably a good idea to add some logging here that name is being fetched from the env variable

pkg/config/controller_config.go
```diff
@@ -88,3 +99,38 @@ func configInit(metadata EC2Metadata) error {
...
+	tagRes, err := ec2Client.DescribeTags(tagReq)
+	if err != nil {
+		return "", err
+	}
+	for _, tag := range tagRes.Tags {
+		if *tag.Key == "aws:eks:cluster-name" {
+			return *tag.Value, nil
+		}
+	}
+	return "", nil
```
mohakkohli, line 135: should we instead want to fail here? It seems like we will not have the cluster name and user wouldn't know that now there are these edge cases they will run into

pkg/deploy/lattice/service_manager.go
```diff
@@ -187,7 +189,16 @@ func (m *defaultServiceManager) updateAssociations(ctx context.Context, svc *Ser
 	}
 
 	for _, assoc := range toDelete {
-		m.deleteAssociation(ctx, assoc.Arn)
+		isManaged, err := m.cloud.IsArnManaged(*assoc.Arn)
```
mohakkohli, line 192: not for now, but in future we can also optimize by avoiding making a call to get tags when createdBy is different account for association.

pkg/aws/cloud.go
```diff
@@ -47,44 +63,68 @@ func NewCloud(log gwlog.Logger, cfg CloudConfig) (Cloud, error) {
...
+	resp, err := c.lattice.ListTagsForResource(tagsReq)
 	if err != nil {
-		return ""
+		return false, err
+	}
+	isManaged := c.IsManagedByTagSet(resp.Tags)
+	return isManaged, nil
+}
+
+func getManagedByTag(cfg CloudConfig) string {
```
mohakkohli, line 118: if everything is empty the tag value will be empty, do we simply take the value and put in or we check for empty strings? 

seems like its allowed to have tag with value of 0 length (Value Length Constraints: Minimum length of 0. Maximum length of 256.) so maybe fine?

mohakkohli (changes requested):
overall looks good. just have questions around whether we should give more visibility to customers when we attempt fall back to empties and what edge cases they need to be aware of.

erikfuller (changes requested):
Thanks for putting this together. After reading through I think I'm fairly convicted we should be using a more unique tag name and require all values to be present for the tag value.

pkg/config/controller_config.go
mikhail-aws: I dont think logging is important here, as long as variable is resolved and resolved to correct name it's not important where it comes from. If variable is in wrong value, we still log entire config, customer can see what actual value is. We should specify variables resolution order in the documentation.

pkg/config/controller_config.go
mikhail-aws: yeah, it seems everyone is moving towards cluster-name must have variable.

pkg/config/controller_config.go
```diff
@@ -88,3 +99,38 @@ func configInit(metadata EC2Metadata) error {
...
 }
+
+// try to find cluster name, search in env then in ec2 instance tags
+// returns empty string if not found
+func getClusterName(sess *session.Session) (string, error) {
+	cn := os.Getenv(CLUSTER_NAME)
+	if cn != "" {
+		return cn, nil
+	}
+	// fallback to ec2 instance tags
```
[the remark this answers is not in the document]
mikhail-aws: Need to dig into fargate-eks, for now lets expect CLUSTER_NAME variable will be set by customer. Will create backlog issue

pkg/aws/cloud.go
```diff
@@ -1,26 +1,42 @@
...
 	"github.com/aws/aws-application-networking-k8s/pkg/utils/gwlog"
-	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
-	"github.com/aws/aws-sdk-go/service/eks"
+	"github.com/aws/aws-sdk-go/service/vpclattice"
+)
+
+const (
+	TagManagedBy = "ManagedBy"
```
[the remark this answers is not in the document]
mikhail-aws: will use `application-networking.k8s.aws/ManagedBy`

pkg/deploy/lattice/service_manager.go
```diff
@@ -102,6 +103,7 @@ func (m *defaultServiceManager) newCreateSvcReq(svc *Service) *CreateSvcReq {
 	svcName := svc.LatticeName()
 	req := &vpclattice.CreateServiceInput{
 		Name: &svcName,
+		Tags: m.cloud.DefaultTags(),
```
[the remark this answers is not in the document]
mikhail-aws: Will remove it from this PR, and leave it for the larger tagging change.

pkg/aws/cloud.go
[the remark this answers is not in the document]
mikhail-aws: Will add validation on the controller-config that all variables are set, we will assume they all correct here. I'm ok with any, we should treat this id as single unique tag, may be even hash it, so customer also dont build anything around exact pattern matching, making it harder for us to change in future.

mikhail-aws:
updated PR:
- added error if cluster name is not found
- naming changes
- removed tagging from service, left only association

mohakkohli (approved):

pkg/config/controller_config.go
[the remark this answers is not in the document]
mikhail-aws: https://github.com/aws/aws-application-networking-k8s/issues/358

pkg/aws/cloud_test.go
```diff
@@ -0,0 +1,81 @@
...
+	t.Run("arn sent", func(t *testing.T) {
+		arn := "arn"
+		lat.EXPECT().ListTagsForResource(gomock.Any()).
+			DoAndReturn(
+				func(req *vpclattice.ListTagsForResourceInput) (*vpclattice.ListTagsForResourceOutput, error) {
+					assert.Equal(t, arn, *req.ResourceArn)
+					return &vpclattice.ListTagsForResourceOutput{}, nil
+				})
+		cl.IsArnManaged(arn)
+	})
```
erikfuller, line 54: Non blocking, but I generally recommend avoiding tests like this that verify implementation rather than outputs. The other tests effectivley do this same work, and a customer argument matcher would help ensure, for example, we aren't always sending through the value ```"arn"```.

erikfuller (approved):

pkg/aws/cloud_test.go
mikhail-aws: Intention here is to test side effect of function, make sure that when arn passed into function we do call lattice with that arn, dont drop it or change it. I might overthink this.

Issue #315, opened 2023-08-09
feature request: be able to customize target group names in k8s


shulin-sq:
When creating multiple target groups I noticed that their naming were cut off/not very helpful given my organization's naming scenes of services and namespaces in k8s.

As per the documentation https://github.com/aws/aws-application-networking-k8s/blob/main/docs/enviroment.md 

the current target group names is 

k8s-(k8s service name)-(k8s service namespace)

and the long version includes a vpc id which we do not need since it's the same for all of our workloads right now. 

Currently my organization names services and namespaces the same (each service gets its own namespace) so the naming of the target group becomes k8s-foo-foo. Any uniqueness comes for a suffix behind the service name, so k8s-foobar-foo. However if the service name is too long the unique suffix is cut off.

Since we have this weird naming scheme, I'm wondering how feasible it would be to introduce a new naming pattern or make that configurable. For example some ideas may be
- an env variable (or perhaps a better input location, perhaps a configmap for the controller?) where we can define a target group naming format 
- a different mode that does not cut off characters of the service name (with a longer character limit for the target group)

solmonk:
Hi, thanks for spotting this issue. We are working on increasing the size limit of target group names in the next release.

solmonk:
TG name size limit got increased with recent naming changes, targeting next release. For full customization capability, we need to figure out direction and details.

erikfuller:
@shulin-sq, can you please have a read through #425 and see if the proposed approach would address your use case?

graehren:
@shulin-sq we released EKS controller version [v0.0.18](https://github.com/aws/aws-application-networking-k8s/releases/tag/v0.0.18) yesterday which contains a revised TG naming strategy. This update should resolve your issues with the TG naming conflict. Please let us know if you have any other questions or feature requests.

PR #357, opened 2023-08-25
Update target group info using TargetGroupPolicy

solmonk:
**What type of PR is this?**

feature

#304 #315 (partial) #69 

Now supporting configuring target group protocols and health checks by attaching TargetGroupPolicy CRD to k8s Service.

This PR includes breaking change on Lattice target group naming scheme. Now TG names on VPC Lattice include protocol and protocol version - this is necessary for avoiding conflicts and supporting a few service export scenarios.

I also have increased the name size to allow longer names.

Whenever a TargetGroupPolicy is applied for existing target group, if protocol / protocolVersion is changed:
* New target group will be created by TG synthesizer, since now the TG name is different.
* Once new target group is active, it will be picked up by Rule synthesizer
* TG synthesizer will clean up old TG

Since we have this order it should be happening quickly without downtime.

If only HC is changed, none of these happens, it will be just calling UpdateTargetGroup API.

WIP things:
* unit tests
* GRPC ServiceExport support (will revisit after GRPC PR merge)

Did manual tests, unit test changes pending

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/deploy/lattice/target_group_manager.go
+++ b/pkg/deploy/lattice/target_group_manager.go
@@ -10,10 +10,12 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/vpclattice"
 
+	"fmt"
 	lattice_aws "github.com/aws/aws-application-networking-k8s/pkg/aws"
 	"github.com/aws/aws-application-networking-k8s/pkg/config"
 	"github.com/aws/aws-application-networking-k8s/pkg/latticestore"
 	latticemodel "github.com/aws/aws-application-networking-k8s/pkg/model/lattice"
+	"strings"
 )
 
 type TargetGroupManager interface {
@@ -33,16 +35,18 @@ func NewTargetGroupManager(cloud lattice_aws.Cloud) *defaultTargetGroupManager {
 	}
 }
 
+// Determines the "actual" target group name used in VPC Lattice.
 func getLatticeTGName(targetGroup *latticemodel.TargetGroup) string {
-	var tgName string
+	var (
+		namePrefix      = targetGroup.Spec.Name
+		protocol        = strings.ToLower(targetGroup.Spec.Config.Protocol)
+		protocolVersion = strings.ToLower(targetGroup.Spec.Config.ProtocolVersion)
+	)
 	if config.UseLongTGName {
-		tgName = latticestore.TargetGroupLongName(targetGroup.Spec.Name,
+		namePrefix = latticestore.TargetGroupLongName(namePrefix,
 			targetGroup.Spec.Config.K8SHTTPRouteName, config.VpcID)
-	} else {
-		tgName = targetGroup.Spec.Name
 	}
-
-	return tgName
+	return fmt.Sprintf("%s-%s-%s", namePrefix, protocol, protocolVersion)
 }
 
 // Create will try to create a target group
@@ -68,12 +72,23 @@ func (s *defaultTargetGroupManager) Create(ctx context.Context, targetGroup *lat
 
 	latticeTGName := getLatticeTGName(targetGroup)
 	// check if exists
-	tgSummary, err := s.findTGByName(ctx, latticeTGName)
+	tgSummary, err := s.findTargetGroup(ctx, targetGroup)
 	if err != nil {
 		return latticemodel.TargetGroupStatus{TargetGroupARN: "", TargetGroupID: ""}, err
 	}
+
+	vpcLatticeSess := s.cloud.Lattice()
 	if tgSummary != nil {
-		return latticemodel.TargetGroupStatus{TargetGroupARN: aws.StringValue(tgSummary.Arn), TargetGroupID: aws.StringValue(tgSummary.Id)}, err
+		if targetGroup.Spec.Config.HealthCheckConfig != nil {
+			_, err := vpcLatticeSess.UpdateTargetGroupWithContext(ctx, &vpclattice.UpdateTargetGroupInput{
+				HealthCheck:           targetGroup.Spec.Config.HealthCheckConfig,
+				TargetGroupIdentifier: tgSummary.Id,
+			})
+			if err != nil {
+				return latticemodel.TargetGroupStatus{TargetGroupARN: "", TargetGroupID: ""}, err
+			}
+		}
+		return latticemodel.TargetGroupStatus{TargetGroupARN: aws.StringValue(tgSummary.Arn), TargetGroupID: aws.StringValue(tgSummary.Id)}, nil
 	}
 
 	glog.V(6).Infof("create targetgroup API here %v\n", targetGroup)
@@ -92,6 +107,7 @@ func (s *defaultTargetGroupManager) Create(ctx context.Context, targetGroup *lat
 		ProtocolVersion: &targetGroup.Spec.Config.ProtocolVersion,
 		VpcIdentifier:   &targetGroup.Spec.Config.VpcID,
 		IpAddressType:   ipAddressType,
+		HealthCheck:     targetGroup.Spec.Config.HealthCheckConfig,
 	}
 
 	targetGroupType := string(targetGroup.Spec.Type)
@@ -114,7 +130,6 @@ func (s *defaultTargetGroupManager) Create(ctx context.Context, targetGroup *lat
 		createTargetGroupInput.Tags[latticemodel.K8SHTTPRouteNamespaceKey] = &targetGroup.Spec.Config.K8SHTTPRouteNamespace
 	}
 
-	vpcLatticeSess := s.cloud.Lattice()
 	resp, err := vpcLatticeSess.CreateTargetGroupWithContext(ctx, &createTargetGroupInput)
 	glog.V(2).Infof("create target group >>>> req [%v], resp[%v] err[%v]\n", createTargetGroupInput, resp, err)
 
@@ -145,7 +160,7 @@ func (s *defaultTargetGroupManager) Get(ctx context.Context, targetGroup *lattic
 	glog.V(6).Infof("Create Lattice Target Group API call for name %s \n", targetGroup.Spec.Name)
 
 	// check if exists
-	tgSummary, err := s.findTGByName(ctx, getLatticeTGName(targetGroup))
+	tgSummary, err := s.findTargetGroup(ctx, targetGroup)
 	if err != nil {
 		return latticemodel.TargetGroupStatus{TargetGroupARN: "", TargetGroupID: ""}, err
 	}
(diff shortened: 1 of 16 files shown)


pkg/deploy/lattice/target_group_manager.go
```diff
@@ -68,12 +72,23 @@ func (s *defaultTargetGroupManager) Create(ctx context.Context, targetGroup *lat
...
+	vpcLatticeSess := s.cloud.Lattice()
 	if tgSummary != nil {
-		return latticemodel.TargetGroupStatus{TargetGroupARN: aws.StringValue(tgSummary.Arn), TargetGroupID: aws.StringValue(tgSummary.Id)}, err
+		if targetGroup.Spec.Config.HealthCheckConfig != nil {
+			_, err := vpcLatticeSess.UpdateTargetGroupWithContext(ctx, &vpclattice.UpdateTargetGroupInput{
+				HealthCheck:           targetGroup.Spec.Config.HealthCheckConfig,
+				TargetGroupIdentifier: tgSummary.Id,
+			})
+			if err != nil {
+				return latticemodel.TargetGroupStatus{TargetGroupARN: "", TargetGroupID: ""}, err
```
erikfuller, line 88: Any reason not to use?
```golang
return latticemodel.TargetGroupStatus{}, err
```

pkg/deploy/lattice/target_group_manager_test.go
```diff
@@ -68,7 +71,7 @@ func Test_CreateTargetGroup_TGNotExist_Active(t *testing.T) {
 		mockVpcLatticeSess := mocks.NewMockLattice(c)
 		arn := "12345678912345678912"
 		id := "12345678912345678912"
-		name := "test"
+		name := "test-http-http1"
```
erikfuller, line 74: Isn't this a breaking change?

pkg/deploy/lattice/target_group_manager_test.go
```diff
@@ -77,13 +80,12 @@ func Test_CreateTargetGroup_TGNotExist_Active(t *testing.T) {
...
 		p := int64(8080)
-		prot := "HTTP"
 		emptystring := ""
 		config := &vpclattice.TargetGroupConfig{
 			Port:            &p,
-			Protocol:        &prot,
+			Protocol:        &tgSpec.Config.Protocol,
 			VpcIdentifier:   &config.VpcID,
-			ProtocolVersion: &emptystring,
+			ProtocolVersion: &tgSpec.Config.ProtocolVersion,
```
erikfuller, line 88: Are Protocol and ProtocolVersion required now when they weren't before or are empty values just an artifact of the unit tests?

pkg/deploy/lattice/target_group_synthesizer.go
```diff
@@ -271,8 +271,10 @@ func (t *targetGroupSynthesizer) SynthesizeSDKTargetGroups(ctx context.Context)
 				glog.V(6).Infof("tgname %v is not used by httproute %v\n", tgName, httpRoute)
 
 			} else {
-
-				isUsed := t.isTargetGroupUsedByaHTTPRoute(ctx, tgName, httpRoute)
+				// We have finished rule reconciliation at this point.
+				// If a target group under HTTPRoute does not have any service, it is stale.
+				isUsed := t.isTargetGroupUsedByaHTTPRoute(ctx, tgName, httpRoute) &&
+					len(sdkTG.getTargetGroupOutput.ServiceArns) > 0
```
erikfuller, line 277: I'm having a hard time understanding why lines 270-288 matter, looks like we're just using it for logging or is there some side effect happening I can't see?

Maybe don't worry about it, this super long function could really use some refactoring.

pkg/deploy/lattice/target_group_synthesizer_test.go
```diff
@@ -371,14 +374,29 @@ func Test_SynthesizeSDKTargetGroups(t *testing.T) {
 			wantDataStoreError:   nil,
 			wantDataStoreStatus:  "",
 		},
+		{
+			name: "Delete SDK TargetGroup since it is dangling with no service associated",
```
erikfuller, line 378: When would this case happen under normal circumstances? Want to make sure we're not being over eager with our deletes.

pkg/gateway/model_build_targetgroup.go
```diff
@@ -404,6 +441,57 @@ func (t *latticeServiceModelBuildTask) buildTargetGroupName(_ context.Context, b
...
+func getAttachedTargetGroupPolicy(ctx context.Context, k8sClient client.Client, svcName, svcNamespace string) (*v1alpha1.TargetGroupPolicy, error) {
+	policyList := &v1alpha1.TargetGroupPolicyList{}
+	err := k8sClient.List(ctx, policyList)
+	if err != nil {
+		return nil, err
+	}
+	for _, policy := range policyList.Items {
+		targetRef := policy.Spec.TargetRef
+		if targetRef == nil {
+			continue
```
erikfuller, line 453: Should we log or is this expected?

pkg/gateway/model_build_targetgroup.go
```diff
@@ -372,6 +391,23 @@ func (t *latticeServiceModelBuildTask) buildTargetGroupSpec(ctx context.Context,
 
 	tgName := latticestore.TargetGroupName(string(httpBackendRef.Name()), namespace)
 
+	tgp, err := getAttachedTargetGroupPolicy(ctx, client, string(httpBackendRef.Name()), namespace)
+	if err != nil {
+		return latticemodel.TargetGroupSpec{}, err
+	}
+	protocol := "HTTP"
+	protocolVersion := vpclattice.TargetGroupProtocolVersionHttp1
```
erikfuller, line 399: Is this from the spec?

pkg/gateway/model_build_targetgroup.go
```diff
@@ -372,6 +391,23 @@ func (t *latticeServiceModelBuildTask) buildTargetGroupSpec(ctx context.Context,
...
+	healthCheckConfig := &vpclattice.HealthCheckConfig{}
+	if tgp != nil {
+		if tgp.Spec.Protocol != nil {
+			protocol = *tgp.Spec.Protocol
+		}
+		if tgp.Spec.ProtocolVersion != nil {
+			protocolVersion = *tgp.Spec.ProtocolVersion
```
erikfuller, line 406: do we need to do any translation on these for the API, like ```ToUpper()``` or anything?

erikfuller (commented):
Mostly just questions from me. Biggest one is around whether or not we should be changing default target group names. Maybe there's already been discussion on one of the issues that I missed?

pkg/deploy/lattice/target_group_synthesizer_test.go
solmonk: This happens when changing target group protocol - it should create `tg-https-http1` and remove `tg-http-http1` for example. Due to how we have detected `inUse` target group before, I need to make a change in behavior here (also related to the comment above)

pkg/gateway/model_build_targetgroup.go
solmonk: yes those are default values

pkg/gateway/model_build_targetgroup.go
solmonk: I can double check but I don't think so. we already have validations on YAML file.

pkg/latticestore/latticestore.go
```diff
@@ -297,12 +297,21 @@ func (ds *LatticeDataStore) GetLatticeService(name string, namespace string) (La
...
+		utils.Truncate(namespace, 50),
+	)
+}
+
+// worst case - (70)-(20)-(21)-https-http2 (125 chars)
+func TargetGroupLongName(defaultName, routeName, vpcId string) string {
+	return fmt.Sprintf("%s-%s-%s",
+		utils.Truncate(defaultName, 70),
+		utils.Truncate(routeName, 20),
+		utils.Truncate(vpcId, 21),
```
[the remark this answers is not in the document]
solmonk: will add descriptions to this, but this is mostly for just fitting in 128b limit. Wanted to put most spaces for the name while preserving some minimum length for routes

This should be eventually changed to some configurable format, as discussed in #315

pkg/gateway/model_build_targetgroup.go
solmonk: I think we can ignore this, my ground rule here is to only care about TargetGroupPolicy that is relevant.

pkg/gateway/model_build_targetgroup.go
```diff
@@ -372,6 +391,23 @@ func (t *latticeServiceModelBuildTask) buildTargetGroupSpec(ctx context.Context,
 
 	tgName := latticestore.TargetGroupName(string(httpBackendRef.Name()), namespace)
 
+	tgp, err := getAttachedTargetGroupPolicy(ctx, client, string(httpBackendRef.Name()), namespace)
```
mikhail-aws, line 394: following code block looks identical to line 198-213

pkg/model/lattice/targetgroup.go
```diff
@@ -29,13 +30,15 @@ type TargetGroupSpec struct {
...
-	EKSClusterName  string `json:"eksclustername"`
-	IsServiceImport bool   `json:"serviceimport"`
+	Port              int32                         `json:"port"`
+	Protocol          string                        `json:"protocol"`
+	ProtocolVersion   string                        `json:"protocolversion"`
+	VpcID             string                        `json:"vpcid"`
+	IpAddressType     string                        `json:"ipaddresstype"`
+	EKSClusterName    string                        `json:"eksclustername"`
+	IsServiceImport   bool                          `json:"serviceimport"`
+	HealthCheckConfig *vpclattice.HealthCheckConfig `json:"healthCheckConfig"`
```
mikhail-aws, line 40: It looks convenient to reuse generated aws-sdk health check config, but it does not align with models package. I'm not sure what the approach should be, my current understanding:
k8s spec -> lattice models -> aws-sdk models -> lattice calls

So lattice models dont really use aws-sdk, we do transform to aws-sdk models in the deploy package, in managers. It might cause code bloat in terms of simple 1:1 models mapping, but we can do something smart about it.

pkg/gateway/model_build_targetgroup.go
solmonk: it is right now, but will be quite different with GRPC serviceexport follow-up change

pkg/model/lattice/targetgroup.go
solmonk: I intend the structure of health check CRD spec to be exactly the same as lattice health check config, to remove any confusion and exactly follow lattice behavior on default values, etc. With this in mind, I still think it is best mapped to the lattice health check config itself. Since lattice models define how we want resources to look like in lattice, I think it kinda aligns.

pkg/gateway/model_build_targetgroup.go
```diff
@@ -420,6 +458,57 @@ func (t *latticeServiceModelBuildTask) buildTargetGroupName(_ context.Context, b
 	}
 }
 
+func getAttachedTargetGroupPolicy(ctx context.Context, k8sClient client.Client, svcName, svcNamespace string) (*v1alpha1.TargetGroupPolicy, error) {
+	policyList := &v1alpha1.TargetGroupPolicyList{}
+	err := k8sClient.List(ctx, policyList)
```
zijun726911, line 463: Any chance to use field selector for `List()` here to aviod for loop?

mikhail-aws (approved):

Issue #359, opened 2023-08-27
[Bug] Http/GRPCRoute's backendref port didn't take effect when updating HttpRoute/GRPCRoute/k8Service


zijun726911:
For following scenarios, Http/GRPCRoute's backendref ports didn't take effect and the controller will create lattice targets with unexpected ports:

- Update k8Service
- Update  Http/GRPCRoute (unexpected targets ports can be created in a moment and then be de-registered immediately with draining, the reason is service_controller [`Watches` HTTPRoute and GRPCRoute](https://github.com/aws/aws-application-networking-k8s/blob/138c1126540432673cd0628ed401c07bcdf96ea2/controllers/service_controller.go#L92) and do [enqueueImpactedService()](https://github.com/aws/aws-application-networking-k8s/blob/1c36c1e706256a24cefd6af63c0d47124dc1310a/controllers/eventhandlers/route.go#L41) )

Root cause:
In this PR #351, we only assign  `&latticeTargetsModelBuildTask{ backendRefPort: port}` in `(t *latticeServiceModelBuildTask) buildTargets()`, but miss to do that in  [`(b *LatticeTargetsModelBuilder) Build()`](https://github.com/aws/aws-application-networking-k8s/blob/138c1126540432673cd0628ed401c07bcdf96ea2/pkg/gateway/model_build_targets.go#L52).

Adding ` backendRefPort: port` logic in the [`(b *LatticeTargetsModelBuilder) Build()`](https://github.com/aws/aws-application-networking-k8s/blob/138c1126540432673cd0628ed401c07bcdf96ea2/pkg/gateway/model_build_targets.go#L52)` is quite likely could fix it.

Furthermore, build targets logic could diverge triggering from k8sService/Route/ServiceExport, should we consider to refactor and combine them?

solmonk:
For ServiceExport, instead of using backendRef ports we are relying on annotations for determining ports - because at the moment of creating TGs we may not have HTTPRoute in the same cluster. Please try setting up `"multicluster.x-k8s.io/port"` annotation on ServiceExport.

zijun726911:
Ok, I think only happen for updating HttpRoute/GRPCRoute/Service scenarios, no  ServiceExport

solmonk:
So your explanation is:
* k8s Service change does trigger all these - ServiceReconciler, HTTPRouteReconciler, and ServiceExportReconciler.
[...]

PR #360, opened 2023-08-27
Refactor event handling and remove service reconciler

solmonk:
**What type of PR is this?**

refactoring/enhancement

#359

We have multiple reconcilers that do the same thing which causes bug #359, and it is really hard to understand what is going on event handler. I did a refactoring to help clarify which resource events are mapped to which other resources.

* (Almost) deleted service reconciler, now it only removes finalizer for keeping compatibility. Its main purpose was watching endpoint event, so now it is watched on serviceexport / route reconcilers instead.
* Removed and unified event handlers. Now service event handler covers all service related events such as policy and endpoints.
* Added `mapper.go` under event handler package dedicated to resource mapping using k8s client api calls 
* Added a few unit tests on event handler

e2etest, also manually tested #359 case by updating existing service endpoints.

**Automation added to e2e**:

```release-note

```


controllers/eventhandlers/service.go
```diff
@@ -2,169 +2,78 @@ package eventhandlers
...
-}
-
-func (h *enqueueRequestForServiceWithExportEvent) Delete(e event.DeleteEvent, queue workqueue.RateLimitingInterface) {
-	h.log.Info("Event: service delete")
-	service := e.Object.(*corev1.Service)
-	h.enqueueImpactedService(queue, service)
-	h.enqueueImpactedServiceExport(queue, service)
+func NewServiceEventHandler(log gwlog.Logger, client client.Client) *serviceEventHandler {
+	return &serviceEventHandler{log: log, client: client,
+		mapper: &resourceMapper{log: log, client: client}}
```
mikhail-aws, line 23: should we create mapper in the main func and pass mapper into controllers?

controllers/eventhandlers/mapper.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	svc := &corev1.Service{}
+	if err := r.client.Get(ctx, svcName, svc); err != nil {
+		if errors.IsNotFound(err) {
+			r.log.Debugw("TargetGroupPolicy is referring to non-existent service, skipping",
+				"policyName", policyName, "serviceName", svcName.String())
+		} else {
+			// Still gracefully skipping the event but errors other than NotFound are bad sign.
+			r.log.Errorw("Failed to query targetRef of TargetGroupPolicy",
+				"policyName", policyName, "serviceName", svcName.String(), "reason", err.Error())
```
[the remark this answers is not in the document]
solmonk: I don't really prefer to panic and shut down whole controller because of this. I think logging error here is pretty much equivalent to panic-recover.

controllers/eventhandlers/mapper_test.go
```diff
@@ -176,59 +101,96 @@ func TestMapToHTTPRoute(t *testing.T) {
...
+			targetKind:      "Service",
+			targetNamespace: nil,
+			serviceFound:    true,
+			success:         true,
+		},
+	}
+
+	for _, tt := range testCases {
+		mockClient := mock_client.NewMockClient(c)
+		mapper := &resourceMapper{log: gwlog.NewLogger(true), client: mockClient}
```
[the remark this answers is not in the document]
solmonk: will update.

controllers/eventhandlers/service.go
solmonk: This is worth considering, but what would we gain from such dependency injection? I did this way because mapper struct is defined mostly for easier unit testing and code reuse, and I don't think we will be configuring anything on the mapper.

controllers/eventhandlers/service.go
```diff
@@ -2,169 +2,78 @@ package eventhandlers
...
+	routes := h.mapper.ServiceToRoutes(ctx, svc, routeType)
 
-			if namespace != ep.Namespace {
-				continue
-			}
-
-			return true
-		}
+	var requests []reconcile.Request
+	for _, route := range routes {
```
mikhail-aws, line 72: nil check on routes

controllers/eventhandlers/serviceimport.go
```diff
@@ -3,121 +3,47 @@ package eventhandlers
...
-
-	if err != nil {
-		h.log.Errorf("Error while listing routes, %s", err)
-	}
+func (h *serviceImportEventHandler) mapToRoute(obj client.Object, routeType core.RouteType) []reconcile.Request {
+	ctx := context.Background()
+	routes := h.mapper.ServiceImportToRoutes(ctx, obj.(*mcs_api.ServiceImport), routeType)
 
+	var requests []reconcile.Request
 	for _, route := range routes {
```
mikhail-aws, line 42: nil check on routes or change mapper to return empty lists when nothing to return?

mikhail-aws (changes requested):

controllers/eventhandlers/serviceimport.go
solmonk: I prefer nil slice over empty slice though, we can still iterate over it. https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

mikhail-aws (approved):

controllers/eventhandlers/serviceimport.go
mikhail-aws: nvm, I forgot that nil is also empty list, so range works fine

PR #370, opened 2023-08-30
Add more grpc yaml examples

zijun726911:
**What type of PR is this?**

- Add more grpc yaml examples
-  Shows more fields in TragetGroupPolicy example

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/examples/target-group-policy.yaml
+++ b/examples/target-group-policy.yaml
@@ -16,7 +16,7 @@ spec:
         healthyThresholdCount: 3
         unhealthyThresholdCount: 2
         path: "/"
-        port: 80
+        port: 8090
         protocol: HTTP
         protocolVersion: HTTP1
         statusMatch: "200"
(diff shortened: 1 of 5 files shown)


examples/target-group-policy.yaml
```diff
@@ -12,3 +12,11 @@ spec:
     healthCheck:
         enabled: true
         intervalSeconds: 10
+        timeoutSeconds: 1
+        healthyThresholdCount: 3
+        unhealthyThresholdCount: 2
+        path: "/"
+        port: 80
```
solmonk, line 19: The port should be 8090 instead in the default setting since it refers to the target port

zijun726911: Changed, BTW if  grpcBin's TLS cert cipher is a concern, I could later create my own cert and build my own grpcBin image

solmonk (approved):

Issue #331, opened 2023-08-16
[Bug] Delete k8sService do not trigger de-registering vpc lattice targets


zijun726911:
For example, if I have a full connectivity setup (set up gateway, httproute, services and deployment) 
if I do `kubectl delete service my-svc` ( my-svc is referenced in the httproute ), the controller didn't trigger de-registering vpc lattice target (I think expected behavior is it should)

We should investigate the root cause

mikhail-aws:
is it fixed in #332 ?

zijun726911:
I think that is a seperate issue, will double confirm

zijun726911:
Just tried this test case https://github.com/aws/aws-application-networking-k8s/blob/9add0cfa73bfca7fb823d47c620b661db531c224/test/suites/integration/deregister_targets_test.go#L69

Not been fix, this bug still exist

zijun726911:
https://github.com/aws/aws-application-networking-k8s/issues/331

Resaon this bug happen:  

Following controller could handle delete k8sService event:
- route_controller.go (build model failed will not invoke lattice api to de-registering targets)
- serviceexport_controller.go (build model failed will not invoke lattice api to de-registering targets)
- service_controller.go (do nothing to handle delete k8sService event )


Potential fix:
 in the service_controller.go, add finalizer when create k8sService, 
when handling k8sService deletion requestion, get all related routes name by `resourceMapper.ServiceToRoutes()`, use k8sService Name and route name to get all its targetGroup by datastore. de-registering targets all targets for these targetGroup

zijun726911:
We had a refactoring that deleted the [service_controller.go](https://github.com/aws/aws-application-networking-k8s/pull/360/files#diff-00e81e75f6d3a8e448c8698e8838e914e508f7bb66e569da6fdd3d394baf96e5) deregistering targets logic and seems intend to use route_controller to do deregistering, but route_controller code path build target model failed

zijun726911:
But I have a impression that even before this refactoring, the K8S service deletion already didn't work

PR #255, opened 2023-05-19
Fix service delete event

zmingxi:
**What type of PR is this?**
bug

**Which issue does this PR fix**:
https://github.com/aws/aws-application-networking-k8s/issues/237

**What does this PR do / Why do we need it**:
Fix

**If an issue # is not available please add repro steps and logs from aws-gateway-controller showing the issue**:

**Testing done on this change**:
make e2etest
Manually run https://github.com/aws/aws-application-networking-k8s/issues/257

**Automation added to e2e**:

**Will this PR introduce any new dependencies?**:
No

**Will this break upgrades or downgrades. Has updating a running cluster been tested?**:
No
**Does this PR introduce any user-facing change?**:
No

```release-note

```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #382, opened 2023-09-02
Bug fix, make sure delete k8sService de-registering targets

zijun726911:
**What type of PR is this?** bug fix for #331 

**Which issue does this PR fix**:  bug fix for #331 

- make sure model_build_targetgroup.go and model_build_targets.go don't return too early so that when handing k8sService deletion event, route_controller.go could finally hit `targetsSynthesizer.Synthesize(ctx)` to deregister the targets
- Add a new field `K8sServiceExists bool` in TargetGroupSpec. Do `targetGroupManager.Create()` in `SynthesizeTriggeredTargetGroupsCreation` only if corresponding K8sServices exist

Use  `my-hotel-gateway.yaml`, `inventory-route-path.yaml`(modified version, have 2 Kind:Service backendrefs), `inventory-ver1.yaml`, `inventory-ver2.yaml` to test various 2 k8sServices and httproute creation and deletion orders, all looks good

**Automation added to e2e**:
- un-commented one previous failed e2etest, whole e2etest suite can pass
```
Ran 29 of 29 Specs in 1928.783 seconds
SUCCESS! -- 29 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestIntegration (1929.29s)
PASS
```

```release-note

```

### DIFF
--- a/pkg/deploy/lattice/target_group_synthesizer.go
+++ b/pkg/deploy/lattice/target_group_synthesizer.go
@@ -3,9 +3,11 @@ package lattice
 import (
 	"context"
 	"errors"
-	"github.com/aws/aws-application-networking-k8s/pkg/utils/gwlog"
-	"github.com/aws/aws-sdk-go/service/vpclattice"
 	"strings"
+
+	"github.com/aws/aws-sdk-go/service/vpclattice"
+
+	"github.com/aws/aws-application-networking-k8s/pkg/utils/gwlog"
 
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -375,6 +377,8 @@ func (t *TargetGroupSynthesizer) PostSynthesize(ctx context.Context) error {
 	return nil
 }
 
+// SynthesizeTriggeredTargetGroupsCreation currently, is only used  by route creation,
+// serviceExport creation uses another method SynthesizeTriggeredTargetGroup()
 func (t *TargetGroupSynthesizer) SynthesizeTriggeredTargetGroupsCreation(ctx context.Context) error {
 	var resTargetGroups []*latticemodel.TargetGroup
 	var returnErr = false
@@ -398,6 +402,13 @@ func (t *TargetGroupSynthesizer) SynthesizeTriggeredTargetGroupsCreation(ctx con
 				resTargetGroup.Spec.Config.IsServiceImport, "")
 			t.log.Infof("targetGroup Synthesized successfully for %s: %v", resTargetGroup.Spec.Name, tgStatus)
 		} else { // handle TargetGroup creation request that triggered by httproute with backendref k8sService creation or serviceExport creation
+			t.log.Debugf("Only create targetGroup if corresponding K8sService exists, resTargetGroup.Spec.K8sServiceExists: %v", resTargetGroup.Spec.K8sServiceExists)
+			if !resTargetGroup.Spec.K8sServiceExists {
+				// In here we just do `continue` and don't do `returnErr = true` , because we hope to handle this scenario:
+				// given the Route and its backendref k8sServices exist, then if the user try to do the deleting k8sService,
+				// we hope the code could reach the later `targetsSynthesizer.Synthesize()` function to deregister lattice targets
+				continue
+			}
 			resTargetGroup.Spec.Config.VpcID = config.VpcID
 			tgStatus, err := t.targetGroupManager.Create(ctx, resTargetGroup)
 			if err != nil {
--- a/test/suites/integration/deregister_targets_test.go
+++ b/test/suites/integration/deregister_targets_test.go
@@ -15,66 +15,74 @@ import (
 	"github.com/aws/aws-application-networking-k8s/test/pkg/test"
 )
 
-var _ = Describe("Deregister Targets", func() {
+var _ = Describe("Deregister Targets", Ordered, func() {
 	var (
-		deployment         *appsv1.Deployment
-		service            *v1.Service
+		deployments        = make([]*appsv1.Deployment, 2)
+		services           = make([]*v1.Service, 2)
+		targetGroups       = make([]*vpclattice.TargetGroupSummary, 2)
 		pathMatchHttpRoute *v1beta1.HTTPRoute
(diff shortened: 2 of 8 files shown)


test/suites/integration/deregister_targets_test.go
```diff
@@ -15,67 +15,75 @@ import (
...
+			services[1],
+			deployments[1],
 		)
 	})
-
-	//It("Kubernetes Service deletion deregisters targets", func() {
-	//	Fail("Currently controller have a bug, service deletion does NOT trigger targets de-registering, need to further investigate the root cause")
-	//	testFramework.ExpectDeleted(ctx, service)
-	//  verifyNoTargetsForTargetGroup(targetGroup)
-	//})
```
zijun726911, line 73: Un-commented this test case and do a refactoring of this test file deregister_targets_test.go, the only purpose of refactoring is to make it run faster i.e., just wait for one 5min draining time in stead of two

mikhail-aws: no commented code please, remove or refactor

pkg/deploy/lattice/target_group_synthesizer.go
```diff
@@ -398,6 +402,13 @@ func (t *TargetGroupSynthesizer) SynthesizeTriggeredTargetGroupsCreation(ctx con
 				resTargetGroup.Spec.Config.IsServiceImport, "")
 			t.log.Infof("targetGroup Synthesized successfully for %s: %v", resTargetGroup.Spec.Name, tgStatus)
 		} else { // handle TargetGroup creation request that triggered by httproute with backendref k8sService creation or serviceExport creation
+			t.log.Debugf("Only create targetGroup if corresponding K8sService exists, resTargetGroup.Spec.K8sServiceExists: %v", resTargetGroup.Spec.K8sServiceExists)
+			if !resTargetGroup.Spec.K8sServiceExists {
+				// In here we just do `continue` and don't do `returnErr = true` , because we hope to handle this scenario:
+				// given the Route and its backendref k8sServices exist, then if the user try to do the deleting k8sService,
+				// we hope the code could reach the later `targetsSynthesizer.Synthesize()` function to deregister lattice targets
+				continue
```
mikhail-aws, line 410: sounds magical, "we hope that it will work". What stop us from making it "for sure will be deleted/de-registered"?

mikhail-aws (changes requested):
It's a tricky fix. Not a fan of this solution. Some indirect, side-effect that we hope will happen.
De-registering targets is not something unusual, so our code should be straight-forward with this use-case. We have delete event, we delete service, we de-register targets. It should be clear from code.

test/suites/integration/deregister_targets_test.go
zijun726911: Already Un-commented, you can see the test case at line 66

zijun726911:
> 

 I once discussed with @solmonk before I send this PR, I think we have 2 main directions to fix this bug:
1.1 Use `route_controller.go` and its underlying code path to de-registering targets. While keep k8sService don't having finalizer. For all `client.Get(ctx, svc)`, keep them return empty but still go ahead to build mode latticeTargets mode (Current PR implimentation)
1.2 Use `route_controller.go` and its underlying code path to de-registering targets. While add k8sService finalizer in service_controller.go and remove k8sService finalizer at the end of route_controller.go, in that case `client.Get(ctx, svc)` can return k8sService when trying to de-registering targets.
3. Use `service_controller.go` to de-registering targets. Add and remove serviceFinalizer in service_controller.go, when handling deleting service, do `mapToAllRoute()` and deregister all related TargetGroups' targets, and keep route_controller.go code don't change  (This is the  straight-forward way that @mikhail-aws mentioned) 

My understanding is option1.1 is the best way from my discussion with @solmonk? Any comment from @solmonk?

pkg/deploy/lattice/target_group_synthesizer.go
```diff
@@ -375,6 +377,8 @@ func (t *TargetGroupSynthesizer) PostSynthesize(ctx context.Context) error {
 	return nil
 }
 
+// Currently, SynthesizeTriggeredTargetGroupsCreation() is used only by httproute creation with backendrefs k8sServices,
+// serviceExport creation uses another method SynthesizeTriggeredTargetGroup()
```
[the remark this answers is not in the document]
zijun726911: Will change

solmonk:
If a backend service is not found, the thing we should do is:
* to set Route's ResolvedRef status as false and gracefully move on.
* Deregister targets as a part of regular endpoint sync logic.

I'm yet to review the details but I believe at least "skip errors when backend service is not found" is a right approach to take.

pkg/deploy/lattice/target_group_synthesizer.go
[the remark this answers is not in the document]
solmonk: another way of dealing with this is updating function name to clearly state what it is for. Looking at the function names..
* SynthesizeTriggeredTargetGroup
* SynthesizeTriggeredTargetGroupsCreation

liwenwu-amazon:
PR #255. should fix de-register targets of deleted service.  Please find out why there is a regression and fix that regression .

zijun726911:
> 

Thanks, will take a look how to set ResolvedRefs status https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPBackendRef

zijun726911:
Discussed with solmonk, that can be a regression, as when we did this refactoring, we did not consider "delete k8sService de-registering targets" use case. 
 [service_controller.go](https://github.com/aws/aws-application-networking-k8s/pull/360/files#diff-00e81e75f6d3a8e448c8698e8838e914e508f7bb66e569da6fdd3d394baf96e5)

Will revivie part of `(r *serviceReconciler) reconcileTargetsResource()` code as the straight-forward fix. 

Will create a backlog task for this idea as a long term improvement:
```
If a backend service is not found, the thing we should do is:
    to set Route's ResolvedRef status as false and gracefully move on. (status part could be addressed later)
    Deregister targets as a part of regular endpoint sync logic.
```

zijun726911:
Will use this way to solve this bug  #387

PR #390, opened 2023-09-07
update go version in docs

mikhail-aws:
Note:
Go v1.21 is out and there are problems with unit tests. Stating explicit version in docs.


mohakkohli (approved):
lets also create an issue to look into why its not working with v1.21.1

PR #391, opened 2023-09-08
Removal of caching for service networks

erikfuller:
**What type of PR is this?**
cleanup

n/a

This is the first of what will be several refactors to eliminate the local ```LatticeDataStore``` cache. Currently, use of the data store results in inconsistent behaviour across the codebase and encourages side-effect programming as many decisions are made based on the presence of or state of the data in the cache rather than using the Lattice APIs as the source of truth. Longer term, this approach should reduce the surface area for consistency bugs or race conditions.

This PR eliminates use of the data store for *service network* operations, instead calling the VPC Lattice API directly. This increases call volume to the ```ListServiceNetworks``` API, but centralizes this logic in a way that should be more reliably cacheable in future.

Additional changes include minor refactoring for readability, typos, or consistency in naming.

Unit tests added and some modified or removed as appropriate.
```
ok  	github.com/aws/aws-application-networking-k8s/pkg/aws	0.976s	coverage: 24.2% of statements
ok  	github.com/aws/aws-application-networking-k8s/pkg/aws/services	0.643s	coverage: 5.8% of statements
ok  	github.com/aws/aws-application-networking-k8s/pkg/config	0.488s	coverage: 32.9% of statements
ok  	github.com/aws/aws-application-networking-k8s/pkg/deploy	1.134s	coverage: 20.6% of statements
ok  	github.com/aws/aws-application-networking-k8s/pkg/deploy/externaldns	2.572s	coverage: 77.1% of statements
ok  	github.com/aws/aws-application-networking-k8s/pkg/deploy/lattice	1.599s	coverage: 85.5% of statements
ok  	github.com/aws/aws-application-networking-k8s/pkg/gateway	2.248s	coverage: 74.7% of statements
ok  	github.com/aws/aws-application-networking-k8s/pkg/latticestore	2.583s	coverage: 70.3% of statements
ok  	github.com/aws/aws-application-networking-k8s/pkg/model/core	2.047s	coverage: 61.7% of statements
ok  	github.com/aws/aws-application-networking-k8s/pkg/model/core/graph	1.997s	coverage: 17.2% of statements
ok  	github.com/aws/aws-application-networking-k8s/pkg/utils	1.735s	coverage: 18.8% of statements
ok  	github.com/aws/aws-application-networking-k8s/controllers	0.661s	coverage: 0.0% of statements
ok  	github.com/aws/aws-application-networking-k8s/controllers/eventhandlers	0.970s	coverage: 54.8% of statements
```

 Ran e2e tests.
```
Ran 29 of 29 Specs in 1881.602 seconds
SUCCESS! -- 29 Passed | 0 Failed | 0 Pending | 0 Skipped
```

**Automation added to e2e**:
No new e2e tests

No

n/a

No


pkg/aws/services/vpclattice.go
```diff
@@ -13,6 +14,12 @@ import (
 
 //go:generate mockgen -destination vpclattice_mocks.go -package services github.com/aws/aws-application-networking-k8s/pkg/aws/services Lattice
 
+type Tags = map[string]*string
```
mikhail-aws, line 17: re-declaration from cloud.go, we can remove it then from cloud.go

pkg/aws/services/vpclattice.go
```diff
@@ -13,6 +14,12 @@ import (
...
+
+type ServiceNetworkInfo struct {
+	SvcNetwork *vpclattice.ServiceNetworkSummary
```
mikhail-aws, line 20: why reference? ServiceNetworkInfo returned below as reference too, I would assume that when we return ServiceNetworkInfo content should be valid, and there is no need for extra nullity check.

pkg/aws/services/vpclattice.go
```diff
@@ -169,3 +177,65 @@ func (d *defaultLattice) ListServiceNetworkServiceAssociationsAsList(ctx context
 
 	return result, nil
 }
+
+func (d *defaultLattice) FindServiceNetwork(ctx context.Context, name string, optionalAccountId string) (*ServiceNetworkInfo, error) {
+	input := vpclattice.ListServiceNetworksInput{}
+
+	for {
+		resp, err := d.ListServiceNetworksWithContext(ctx, &input)
```
mikhail-aws, line 185: there is implementation for paginated view in sdk: ListListenersPagesWithContext. That will replace sharp-edged infinite for loop with closure function.

https://github.com/aws/aws-sdk-go/blob/v1.45.6/service/vpclattice/vpclatticeiface/interface.go#L195

pkg/aws/services/vpclattice.go
```diff
@@ -169,3 +177,65 @@ func (d *defaultLattice) ListServiceNetworkServiceAssociationsAsList(ctx context
...
+		}
+
+		if resp.NextToken == nil {
+			break
+		}
+
+		input.NextToken = resp.NextToken
+	}
+
+	return nil, nil
```
mikhail-aws, line 227: is nil,nil = not found? lattice store returns error on not found, we might have regression, but I havnt look how do we handle errors from store, might be ok-ish?

mikhail-aws (changes requested):

pkg/aws/services/vpclattice.go
erikfuller: Great question. There are only two places in the code that weren't tests and called the data store previously.

#### gateway_controller.go#254
```
	serviceNetworkStatus, err := r.datastore.GetServiceNetworkStatus(gw.Name, config.AccountID)
	if err = r.updateGatewayStatus(ctx, &serviceNetworkStatus, gw); err != nil {
		return err
	}
```
Here, previous logic actually ignored the error, which could have potentially caused issues. Since we weren't seeing crashes here, maybe there was some other invariant that always assured this codepath would resolve to an item in the cache. Ultimately, this value is only used for a message on a condition update.

#### service_manager.go#82
```
	sn, err := m.datastore.GetServiceNetworkStatus(snName, m.cloud.Config().AccountId)
	if err != nil {
		return err
	}
	assocReq := &CreateSnSvcAssocReq{
		ServiceIdentifier:        svcId,
		ServiceNetworkIdentifier: aws.String(sn.ID),
		Tags:                     m.cloud.DefaultTags(),
	}
```
While this code checks for and returns an error, what it cares about is obtaining the ID for the service network, not branching based on availability of the object in the cache. In the new code, if for some reason we can't find the service network, we'll return an error here and retry.

pkg/aws/services/vpclattice.go
erikfuller: I'll sync up with you offline on this one. I adapted existing implementations in this file, but if they are flawed then that's no good. Basically every other method in this file has this pattern.

pkg/aws/services/vpclattice.go
mikhail-aws: I would add nullity check to service_manager.go and throw error, otherwise it might panic

pkg/utils/common.go
```diff
@@ -1,16 +1,25 @@
...
 
 type MapFunc[T any, U any] func(T) U
 type FilterFunc[T any] func(T) bool
 
-// TODO: should be check by API call (Mingxi)
-func ArntoId(arn string) string {
-	if len(arn) == 0 {
-		return ""
+// arn:<partition>:vpc-lattice:<region>:<account id>:<resource-type>/<resource-id>
+func ArnToAccountId(arn string) (string, error) {
```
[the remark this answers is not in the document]
erikfuller: Thank you! I failed in my apparently lazy attempt to find this on my own.

pkg/aws/services/vpclattice.go
erikfuller: Using the paginated methods will really simplify the methods in this file. We agreed I could postpone that to the next PR.

pkg/deploy/lattice/service_manager.go
```diff
@@ -79,13 +79,13 @@ func (m *defaultServiceManager) createServiceAndAssociate(ctx context.Context, s
 }
 
 func (m *defaultServiceManager) createAssociation(ctx context.Context, svcId *string, snName string) error {
-	sn, err := m.datastore.GetServiceNetworkStatus(snName, m.cloud.Config().AccountId)
+	snInfo, err := m.cloud.Lattice().FindServiceNetwork(ctx, snName, m.cloud.Config().AccountId)
```
liwenwu-amazon, line 82: This can be a expensive call.

pkg/utils/common.go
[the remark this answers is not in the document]
erikfuller: Fixed

pkg/aws/services/vpclattice.go
erikfuller: Done

pkg/aws/services/vpclattice.go
erikfuller: Good point. Updated.

pkg/deploy/lattice/service_manager.go
erikfuller: Absolutely. I was planning on addressing performance at the end of the refactor, but I could potentially do it now. We were already iterating over all items in ```service_manager.go```, though this will certainly be more calls.

My thinking here was to first consolidate the logic so that we were only dealing with something more like a read-through cache, then optimize the calls. Does that sound reasonable?

pkg/aws/services/vpclattice.go
erikfuller: Good catch! Updated in the two places missing it. All calls should now be checking for a nil result.

pkg/deploy/lattice/service_network_synthesizer.go
```diff
@@ -117,7 +81,39 @@ func (s *serviceNetworkSynthesizer) synthesizeTriggeredGateways(ctx context.Cont
...
 		return nil
 	}
+}
+
+func (s *serviceNetworkSynthesizer) deleteServiceNetwork(ctx context.Context, resServiceNetwork *latticemodel.ServiceNetwork) string {
+	glog.V(6).Infof("Synthesizing Gateway: Del %v\n", resServiceNetwork.Spec.Name)
+
+	// TODO need to check if service network is referenced by gateway in other namespace
+	gwList := &gateway_api.GatewayList{}
+	s.Client.List(context.TODO(), gwList)
```
mikhail-aws, line 91: nit: use context from function argument - ctx, not context.TODO()

pkg/deploy/lattice/service_network_synthesizer.go
```diff
@@ -117,7 +81,39 @@ func (s *serviceNetworkSynthesizer) synthesizeTriggeredGateways(ctx context.Cont
...
+	if snUsedByGateway {
+		glog.V(6).Infof("Skipping deleting gw: %v since it is still used by gateway(s)",
+			resServiceNetwork.Spec.Name)
+
+		return ""
+	}
+
+	err := s.serviceNetworkManager.Delete(ctx, resServiceNetwork.Spec.Name)
+	if err != nil {
+		return LATTICE_RETRY
```
mikhail-aws, line 111: nit: return or log original error. For example:
```go
		return fmt.Errorf("%w: %w", RetryErr, err)
```
https://github.com/aws/aws-application-networking-k8s/blob/main/pkg/deploy/lattice/error.go#L13

Example: https://github.com/aws/aws-application-networking-k8s/blob/main/pkg/deploy/lattice/service_manager.go#L207-L214

pkg/deploy/lattice/service_network_synthesizer.go
```diff
@@ -117,7 +81,39 @@ func (s *serviceNetworkSynthesizer) synthesizeTriggeredGateways(ctx context.Cont
 	} else {
 		return nil
 	}
+}
+
+func (s *serviceNetworkSynthesizer) deleteServiceNetwork(ctx context.Context, resServiceNetwork *latticemodel.ServiceNetwork) string {
```
mikhail-aws, line 86: should return error

mikhail-aws (approved):
You can address comments in following PR. The deleteServiceNetwork function that return string stands out.

pkg/deploy/lattice/service_manager.go
```diff
@@ -79,13 +79,17 @@ func (m *defaultServiceManager) createServiceAndAssociate(ctx context.Context, s
...
 	if err != nil {
 		return err
 	}
+	if snInfo == nil {
+		return fmt.Errorf("Service network %s for account %s not found", snName, m.cloud.Config().AccountId)
+	}
+
```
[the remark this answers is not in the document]
erikfuller: I agree. Currently there were only two places, but as I do more resource types I'll see which pattern makes more sense.

pkg/deploy/lattice/service_network_synthesizer.go
erikfuller: This was a straight refactor-to-method, but happy to update when I see these.

pkg/deploy/lattice/service_network_synthesizer.go
erikfuller: This was also a straight copy/paste/new method refactor, so for now I'll just add a log line. I'm hesitant to change too much of the flow logic at once.

pkg/deploy/lattice/service_network_synthesizer.go
erikfuller: I don't disagree, but I was trying to minimize logic changes in the refactor. Previously, this code only set a return string.

PR #394, opened 2023-09-13
Removal of caching for services

erikfuller:
**What type of PR is this?**
cleanup

n/a

This is the second refactor to ultimately eliminate the ```LatticeDataStore```. This PR eliminates caching for services. Prev PR: https://github.com/aws/aws-application-networking-k8s/pull/391 

This PR eliminates use of the data store for *service* operations and instead calls the VPC Lattice API directly. This increases call volume to the ```ListServices``` API, but centralizes this logic in a way that should be more reliably cacheable in future.

Additional changes include minor refactoring for readability, typos, or consistency in naming. Also, I had some follow up actions from the previous PR addressed here. These include:

1. Returning an error when resource not found by name
2. Using auto-paginated SDK methods (```List*PagesWithContext```)
3. Removed unused cleanup logic from e2e tests

Automated tests (unit/integration).

make test
```
go test ./pkg/... ./controllers/... -coverprofile coverage.out
?   	github.com/aws/aws-application-networking-k8s/pkg/apis/applicationnetworking/v1alpha1	[no test files]
ok  	github.com/aws/aws-application-networking-k8s/pkg/aws	0.503s	coverage: 24.2% of statements
ok  	github.com/aws/aws-application-networking-k8s/pkg/aws/services	0.808s	coverage: 5.8% of statements
ok  	github.com/aws/aws-application-networking-k8s/pkg/config	0.898s	coverage: 32.9% of statements
?   	github.com/aws/aws-application-networking-k8s/pkg/k8s	[no test files]
?   	github.com/aws/aws-application-networking-k8s/pkg/model/lattice	[no test files]
?   	github.com/aws/aws-application-networking-k8s/pkg/runtime	[no test files]
?   	github.com/aws/aws-application-networking-k8s/pkg/utils	[no test files]
?   	github.com/aws/aws-application-networking-k8s/pkg/utils/gwlog	[no test files]
?   	github.com/aws/aws-application-networking-k8s/pkg/utils/retry	[no test files]
?   	github.com/aws/aws-application-networking-k8s/pkg/utils/ttime	[no test files]
ok  	github.com/aws/aws-application-networking-k8s/pkg/deploy	0.872s	coverage: 20.6% of statements
ok  	github.com/aws/aws-application-networking-k8s/pkg/deploy/externaldns	1.367s	coverage: 77.1% of statements
ok  	github.com/aws/aws-application-networking-k8s/pkg/deploy/lattice	2.268s	coverage: 85.4% of statements
ok  	github.com/aws/aws-application-networking-k8s/pkg/gateway	2.794s	coverage: 74.9% of statements
ok  	github.com/aws/aws-application-networking-k8s/pkg/latticestore	1.053s	coverage: 64.1% of statements
ok  	github.com/aws/aws-application-networking-k8s/pkg/model/core	2.963s	coverage: 61.7% of statements
ok  	github.com/aws/aws-application-networking-k8s/pkg/model/core/graph	2.903s	coverage: 17.2% of statements
ok  	github.com/aws/aws-application-networking-k8s/controllers	0.772s	coverage: 0.0% of statements
ok  	github.com/aws/aws-application-networking-k8s/controllers/eventhandlers	1.111s	coverage: 54.8% of statements

```

e2e tests
```
Ran 31 of 31 Specs in 2050.026 seconds
SUCCESS! -- 31 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestIntegration (2050.59s)
```

No

n/a

No


controllers/gateway_controller.go
```diff
@@ -254,9 +250,6 @@ func (r *GatewayReconciler) reconcileGatewayResources(ctx context.Context, gw *g
 	if err != nil {
 		return err
 	}
-	if snInfo == nil {
-		return fmt.Errorf("Service network %s for account %s not found", gw.Name, config.AccountID)
-	}
```
erikfuller, line 259: With the change on ```FindX``` logic to raise a "not found" error, these additional nil checks on the return value are no longer required.

pkg/aws/services/vpclattice.go
```diff
@@ -20,6 +22,29 @@ type ServiceNetworkInfo struct {
...
 }
+
+type LatticeNameProvider interface {
+	LatticeName() string
+}
+
+type NotFoundError struct {
+	ResourceType string
+	Name         string
+}
```
erikfuller, line 33: Getting something that looked and behaved sensibly for error handling and identifying errors of this type took me a while. For folks more familiar with Go please feel free to suggest an alternate approach if you think there's a better option.

pkg/aws/services/vpclattice.go
```diff
@@ -54,125 +80,101 @@ func NewDefaultLattice(sess *session.Session, region string) *defaultLattice {
...
-		if resp.NextToken == nil {
-			break
+	err := d.ListServiceNetworksPagesWithContext(ctx, input, func(page *vpclattice.ListServiceNetworksOutput, lastPage bool) bool {
+		for _, sn := range page.Items {
+			result = append(result, sn)
 		}
-		input.NextToken = resp.NextToken
-		resp, err = d.ListServiceNetworksWithContext(ctx, input)
+		return true
+	})
```
erikfuller, line 89: I've updated all the methods here to use automatic pagination as per @mikhail-aws suggestion in previous PR. The way these methods work is you provide a function for them to call for each page found and you return ```true``` if you want the call to keep fetching pages.

test/pkg/test/framework.go
```diff
@@ -200,37 +189,11 @@ func (env *Framework) ExpectToBeClean(ctx context.Context) {
...
 					continue
 				}
-				g.Expect(env.TestCasesCreatedServiceNames).To(Not(ContainElements(BeKeyOf(*tg.Name))))
 			}
 		}
 	}).Should(Succeed())
 }
 
-// if we still want this method, we should tag the things we create with the test suite
-func (env *Framework) CleanTestEnvironment(ctx context.Context) {
```
erikfuller, line 210: IMO the "best" way to clean up the test environment is via the new e2e-clean target. Tracking these resources as side-effects of various creation calls seems unreliable, similar to the reasons we're eliminating the ```LatticeDataStore``` cache.

Given we're not using the cleanup methods today, I think removing them is better for now than continuing to maintain this unused code.

pkg/aws/services/vpclattice.go
mikhail-aws: compact version:

```go
 7 func (d *defaultLattice) ListServiceNetworksAsList(ctx context.Context, input *vpclattice.ListServiceNetworksInput) ([]*vpclattice.ServiceNetworkSummary, error) {
  8         result := []*vpclattice.ServiceNetworkSummary{}
  9         err := d.ListServiceNetworksPagesWithContext(ctx, input, func(page *vpclattice.ListServiceNetworksOutput, lastPage bool) bool {
 10                 result = append(result, page.Items...)
 11         }
 12         return result, err
 13 }
```

Generic get all pages function. Since every page is a struct, getting Items from page can be done through reflection, or explicitly written as getItemsFn, that return items from page.

```go
 func ListAll[REQ any, ITEM any, PAGE any](
         ctx context.Context,    req REQ,
         listPages func(context.Context, REQ, func(PAGE, bool) bool, ...request.Option) error,
         getPageItems func(PAGE) []ITEM)([]ITEM, error) {
                 out := []ITEM{}
                 err := listPages(ctx, req, func(page PAGE, _ bool) bool {
                         out = append(out, getPageItems(page)...)
                         return true
                 })
                 return out, err
         }

 func (l *Lattice) ListServiceNetworksAsList(ctx context.Context, in *vpclattice.ListServiceNetworksInput) ([]*vpclattice.ServiceNetworkSummary, error) {
         return ListAll(
                 ctx, in, l.c.ListServiceNetworksPagesWithContext,
                 func(page *vpclattice.ListServiceNetworksOutput) []*vpclattice.ServiceNetworkSummary {
                         return page.Items
                 })
 }
```

mikhail-aws: I dont see any value in unit-tests with mocks for these Lattice wrapper functions. As you mentioned, it's painful and kind of self-fulfilling prophecy.

mikhail-aws (commented):
Didnt have a chance review everything. Left minor comments.

test/pkg/test/framework.go
mikhail-aws: Interesting idea, do you want to remove cleanup after each test and do only one at the very end of all tests?

pkg/aws/services/vpclattice.go
zijun726911: I am curious what the benefit of  `ListServiceNetworksPagesWithContext(ctx, input, fn)`  compared to the old `NextToken` way? (Still learning this part)

pkg/deploy/lattice/listener_manager.go
```diff
@@ -35,24 +37,37 @@ func NewListenerManager(cloud lattice_aws.Cloud, latticeDataStore *latticestore.
...
+}
 
-	if err != nil {
-		errmsg := fmt.Sprintf("Service %v not found during listener creation", listener.Spec)
-		glog.V(6).Infof("Error during create listner %s \n", errmsg)
-		return latticemodel.ListenerStatus{}, errors.New(errmsg)
-	}
+func (r *ListenerNameProvider) LatticeName() string {
+	return utils.LatticeServiceName(r.l.Spec.Name, r.l.Spec.Namespace)
+}
```
zijun726911, line 50: Does this method use lattice listener name?

When create listener, we use that format?  i.e., `listener.Spec.Name` should follow this format? "routeName-routeNS-port-protocol"

https://github.com/aws/aws-application-networking-k8s/blob/138c1126540432673cd0628ed401c07bcdf96ea2/pkg/gateway/model_build_listener.go#L149

test/pkg/test/framework.go
erikfuller: Removing resources as each test finishes is what it does today. We don't call anything at the end of all tests. These ```Clean*``` methods are currently unused.

pkg/aws/services/vpclattice.go
erikfuller: Honestly I'm not convinced it's that big a win. The logic is a little harder to wrap your head around, but it means we can't get into an infinite loop or accidently omit pages by mishandling ```NextToken```.

pkg/deploy/lattice/listener_manager.go
erikfuller: The ```LatticeName()``` is intended to return the Lattice service name, based off the original route values for the listener. I can see here in this context ```LatticeName()``` is causing confusion. I'll update this to be ```LatticeServiceName()``` or similar instead.

zijun726911: Oh, Seems your code is correct, I confused by the  second param`listenerResourceName` and fifth param ` t.route.Name()` here

https://github.com/aws/aws-application-networking-k8s/blob/138c1126540432673cd0628ed401c07bcdf96ea2/pkg/gateway/model_build_listener.go#L152

pkg/aws/services/vpclattice.go
```diff
@@ -20,6 +22,29 @@ type ServiceNetworkInfo struct {
...
+}
+
+type NotFoundError struct {
+	ResourceType string
+	Name         string
+}
+
+func (e *NotFoundError) Error() string {
+	return fmt.Sprintf("%s %s not found", e.ResourceType, e.Name)
+}
```
zijun726911, line 37: Do we need to change to:
```
return fmt.Sprintf("Lattice %s %s not found", e.ResourceType, e.Name)
```

So that log can show  "Lattice service xxxxxx not found" and "Lattice service network xxxxxx not found"

pkg/aws/services/vpclattice.go
```diff
@@ -181,51 +183,81 @@ func (d *defaultLattice) ListServiceNetworkServiceAssociationsAsList(ctx context
...
+
+	return snMatch, nil
+}
+func (d *defaultLattice) FindService(ctx context.Context, nameProvider LatticeServiceNameProvider) (*vpclattice.ServiceSummary, error) {
+	serviceName := nameProvider.LatticeServiceName()
+	input := vpclattice.ListServicesInput{}
+
+	var svcMatch *vpclattice.ServiceSummary
+	err := d.ListServicesPagesWithContext(ctx, &input, func(page *vpclattice.ListServicesOutput, lastPage bool) bool {
+		for _, svc := range page.Items {
```
zijun726911, line 243: Does FindService() also need to check:

```
acctIdMatches, err1 := accountIdMatches(optionalAccountId, *r.Arn)

if !acctIdMatches {
    continue
}
``` 

in case in the future we support RAM share lattice Service in the controller and one current account Service and one foreign Service have the same name. (check accountId at least not harmful?)

zijun726911 (approved):
Overall, LGTM,  both current `IsNotFoundError()` and the error handling way the @mikhail-aws  mentions are ok for me

pkg/aws/services/vpclattice.go
erikfuller: I'll update this for next PR

pkg/deploy/lattice/listener_manager.go
erikfuller: Yeah these naming bits get confusing

PR #396, opened 2023-09-14
Add VpcAssociationPolicy CRD

zijun726911:
Which issue does this PR fix: Add VpcAssociationPolicy CRD to support SNVA security Group API

What does this PR do / Why do we need it: Add VpcAssociationPolicy CRD to support SNVA security Group API

If an issue # is not available please add repro steps and logs from aws-gateway-controller showing the issue:

Testing done on this change:

Installed this CRD in my own cluster succeeded by kubectl apply -f config/crds/bases/application-networking.k8s.aws_vpcassociationpolicies.yaml

**Automation added to e2e**:

```release-note

```


liwenwu-amazon:
Shall we use a different name, e.g. GatewayAWSPolicy?   And we can add other parameters to them, e.g.
* enable/disable access log

zijun726911:
> Shall we use a different name, e.g. GatewayAWSPolicy? And we can add other parameters to them, e.g.
> 
>     * enable/disable access log

I think we decide to do them in 2 CRDs, 
- one CRD for access log and authpolicy that can apply to both Gateway and Route
- Another one VpcAssociationPolicy CRD that can only apply to Gateway


For the name of VpcAssociationPolicy we could discuss, GatewayAWSPolicy can be a option

pkg/apis/applicationnetworking/v1alpha1/vpcassociationpolicy_types.go
```diff
@@ -0,0 +1,70 @@
...
+type VpcAssociationPolicySpec struct {
+
+	// SecurityGroupIds defines the security groups enforced on the VpcServiceNetworkAssociation.
+	// For more details, please check the VPC Lattice documentation https://docs.aws.amazon.com/vpc-lattice/latest/ug/security-groups.html
+	//
+	// +optional
+	// +kubebuilder:validation:MinItems=1
+	SecurityGroupIds []SecurityGroupId `json:"securityGroupIds,omitempty"`
+
+	// AssociateWithVpc indicates whether the VpcServiceNetworkAssociation should be created for current the VPC of EKS cluster.
```
solmonk, line 50: typo: `current the` -> `the current`

I would avoid saying "EKS", I think this product is supposed to work for non-eks k8s clusters

pkg/model/core/policy.go
```diff
@@ -0,0 +1,11 @@
+package core
+
+import (
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/gateway-api/apis/v1alpha2"
+)
+
+type Policy interface {
+	GetNamespacedName() types.NamespacedName
+	GetTargetRef() *v1alpha2.PolicyTargetReference
```
solmonk, line 10: I'm okay with having this interface but not sure how we could benefit from it

pkg/apis/applicationnetworking/v1alpha1/vpcassociationpolicy_types.go
```diff
@@ -0,0 +1,70 @@
...
+	//
+	// Both this flag and Gateway annotation "application-networking.k8s.aws/lattice-vpc-association" are reserved tentatively for backward compatibility.
+	// Either one of them set to true or both of them undefined will result in the VpcServiceNetworkAssociation created.
+	// +optional
+	// +kubebuilder:default=true
+	AssociateWithVpc *bool `json:"associateWithVpc,omitempty"`
```
solmonk, line 56: looking at this I'm thinking it is quite not straightforward to have optional value true by default. Could we have an inverse of this option?

pkg/apis/applicationnetworking/v1alpha1/vpcassociationpolicy_types.go
```diff
@@ -0,0 +1,70 @@
...
+// +kubebuilder:validation:MaxLength=32
+// +kubebuilder:validation:MinLength=3
+// +kubebuilder:validation:Pattern=`^sg-[0-9a-z]+$`
+type SecurityGroupId string
+
+// VpcAssociationPolicySpec defines the desired state of VpcAssociationPolicy.
+type VpcAssociationPolicySpec struct {
+
+	// SecurityGroupIds defines the security groups enforced on the VpcServiceNetworkAssociation.
+	// For more details, please check the VPC Lattice documentation https://docs.aws.amazon.com/vpc-lattice/latest/ug/security-groups.html
```
solmonk, line 44: Worth mentioning that it does not take any effect when it is not associated with vpc

pkg/apis/applicationnetworking/v1alpha1/vpcassociationpolicy_types.go
zijun726911: Ok.. but k8s cluster in other cloud provider don't have vpc lattice?

pkg/apis/applicationnetworking/v1alpha1/vpcassociationpolicy_types.go
zijun726911: I can remove  the `// +kubebuilder:default=true` here


I will do the in the controller code logic: ` Either one of them set to true or both of them undefined will result in the VpcServiceNetworkAssociation created.`

pkg/model/core/policy.go
zijun726911: Actually my latter code change will use that:

```
func (r *resourceMapper) TargetGroupPolicyToService(ctx context.Context, tgp *v1alpha1.TargetGroupPolicy) *corev1.Service {
	if obj := r.policyToTargetRefObj(ctx, tgp, &corev1.Service{}); obj != nil {
		return obj.(*corev1.Service)
	} else {
		return nil
	}
}

func (r *resourceMapper) VpcAssociationPolicyToGateway(ctx context.Context, vap *v1alpha1.VpcAssociationPolicy) *gateway_api.Gateway {
	if obj := r.policyToTargetRefObj(ctx, vap, &gateway_api.Gateway{}); obj != nil {
		return obj.(*gateway_api.Gateway)
	} else {
		return nil
	}
}

func (r *resourceMapper) policyToTargetRefObj(ctx context.Context, policy core.Policy, retObj client.Object) client.Object { }
	
```

pkg/apis/applicationnetworking/v1alpha1/vpcassociationpolicy_types.go
solmonk: could be plain, un-managed k8s cluster installed on ec2 instance

solmonk (approved):

Issue #386, opened 2023-09-05
v0.0.16 Controller fails to create/update a Lattice Service when no endpoints in any of the corresponding K8S Services


klyubin:
Steps to reproduce:
1. Create a K8S HTTPRoute pointed at a Lattice Gateway as the parent and, as the backend, a K8S Service whose selectors don't match any endpoints.
2. Observe that the Controller is unable to create (or update) the Lattice Service, and keeps spewing the following:
```json
{"level":"info","ts":1693939293.0518966,"logger":"controller.route","caller":"controllers/route_controller.go:312","msg":"RouteReconciler: Failed deploy traffic-exemplar due to err TargetSynthesize: Failed to create targets :&{{0xc012a2d000 AWS:VPCServiceNetwork::Targets k8s-traffic-exemplar-traffic-exemplar} {traffic-exemplar traffic-exemplar traffic-exemplar  []}} , err:InvalidParameter: 1 validation error(s) found.\n- missing required field, RegisterTargetsInput.Targets.\n\n"}
```

It appears that the Controller is calling some API which does not let it provide an empty list of Targets for a Target Group. Strangely, the AWS Console does let me create/update Target Groups without any targets. So, it seems the Controller has a bug.

solmonk:
Thanks for reporting this issue.

Relevant: https://github.com/aws/aws-application-networking-k8s/blob/main/pkg/deploy/stack_deployer.go#L120-L133

I had a quick look, I feel like this is not a v0.0.16 regression and the controller has been working like this since the beginning. The controller cannot proceed to service reconciliation if it cannot finish target registration, and I don't recall any validation logic that prevents empty RegisterTargets request.

mikhail-aws:
Hints for fix:
- https://github.com/aws/aws-application-networking-k8s/blob/main/pkg/deploy/lattice/targets_synthesizer.go#L43-L44
- check if list is empty

klyubin:
Thank you very much!

PR #406, opened 2023-09-20
fix: create lattice resources when no endpoints in K8s services (#386)

scottlaiaws:
**What type of PR is this?**

bug

#386 

This PR allows the controller to continue to create Lattice service, and listeners, and rules when there are no targets to register

Added unit tests, and e2e tests.

`make e2e-test` output:

```
Ran 32 of 32 Specs in 2332.975 seconds
--- PASS: TestIntegration (2334.12s)
PASS
ok      github.com/aws/aws-application-networking-k8s/test/suites/integration   2334.147s
```

**Automation added to e2e**:

Added test case `httproute_creation_test.go`

### DIFF
--- a/pkg/deploy/lattice/targets_manager.go
+++ b/pkg/deploy/lattice/targets_manager.go
@@ -3,6 +3,7 @@ package lattice
 import (
 	"context"
 	"errors"
+
 	"github.com/golang/glog"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -94,6 +95,11 @@ func (s *defaultTargetsManager) Create(ctx context.Context, targets *latticemode
 			Port: &port,
 		}
 		targetList = append(targetList, &t)
+	}
+
+	// No targets to register
+	if len(targetList) == 0 {
+		return nil
 	}
 
 	registerRouteInput := vpclattice.RegisterTargetsInput{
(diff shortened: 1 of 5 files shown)


pkg/deploy/lattice/targets_manager.go
```diff
@@ -96,6 +97,11 @@ func (s *defaultTargetsManager) Create(ctx context.Context, targets *latticemode
 		targetList = append(targetList, &t)
 	}
 
+	if len(targetList) == 0 {
+		glog.Warningln("There are no available targets to regiser. Skipping targets registration")
```
solmonk, line 101: IMO I don't think we need this log line

solmonk (approved):
approving with minor comment

test/suites/integration/httproute_creation_test.go
```diff
@@ -82,6 +83,25 @@ var _ = Describe("HTTPRoute Creation", func() {
 		})
 	})
 
+	Context("Order #4: Service, HttpRoute", func() {
+		It("Create lattice resources successfully when no Targets available", func() {
+			httpRoute = testFramework.NewHttpRoute(testGateway, service, "Service")
+
+			// Override port
```
zijun726911, line 90: Do you need to add more comments here to say that you override the port to cause no port match so that it will cause no Targets in the TG? Or use a more explicit way to construct a no targetPort k8sService?

scottlaiaws: I can add more comments. I doubt creating "no targetPort" K8s service will be used in other place so it will be over-kill to create a method to do that.

zijun726911 (approved):

Issue #199, opened 2023-04-03
allow K8S user to specify security group for the service network (gateway) it associate with


liwenwu-amazon:
Today, lattice service allow a user to specify security group during service network to VPC association.  K8S needs similar capability to express that intent

zijun726911:
https://github.com/aws/aws-application-networking-k8s/pull/396
https://github.com/aws/aws-application-networking-k8s/pull/398

PR #409, opened 2023-09-22
service_network_manager changes for SNVA security group API support

zijun726911:
**What type of PR is this?**

**Which issue does this PR fix**: feature, #199 service_network_manager changes for SNVA security group API support 

- Finished following manual test steps:
1. `kubectl apply -f examples/my-hotel-gateway.yaml`, create SN and SNVA success
2. apply:
```
apiVersion: application-networking.k8s.aws/v1alpha1
kind: VpcAssociationPolicy
metadata:
    name: test-vpc-association-policy
spec:
    targetRef:
        group: "gateway.networking.k8s.io"
        kind: Gateway
        name: my-hotel
    securityGroupIds:
        - sg-0063d957ef47d3020
    associateWithVpc: true

```
Update SNVA security group success
3. apply 
```
apiVersion: application-networking.k8s.aws/v1alpha1
kind: VpcAssociationPolicy
metadata:
    name: test-vpc-association-policy
spec:
    targetRef:
        group: "gateway.networking.k8s.io"
        kind: Gateway
        name: my-hotel
    associateWithVpc: true
```
lattice.UpdateServiceNetworkVpcAssociationWithContext() return validation error and the controller keep LATTICE_RETRY as expected

4. apply 
```
apiVersion: application-networking.k8s.aws/v1alpha1
kind: VpcAssociationPolicy
metadata:
    name: test-vpc-association-policy
spec:
    targetRef:
        group: "gateway.networking.k8s.io"
        kind: Gateway
        name: my-hotel
    associateWithVpc: false
```
SNVA deleted as expected

5. apply 
```
apiVersion: application-networking.k8s.aws/v1alpha1
kind: VpcAssociationPolicy
metadata:
    name: test-vpc-association-policy
spec:
    targetRef:
        group: "gateway.networking.k8s.io"
        kind: Gateway
        name: my-hotel
```
SNVA is created, (as the controller by default will create SNVA, if the policy don't have `associateWithVpc` field)

- Create vpc-association-policy first, then create the my-hotel gateway. vpc-association-policy  can take effect, and all behavior as expected 
-  All existing E2E test pass:

```

Ran 32 of 32 Specs in 2074.314 seconds
--- PASS: TestIntegration (2074.81s)
PASS
```

**Automation added to e2e**: Will add new E2E test cases in a separate PR 

**Does this PR introduce any user-facing change?**: add new feature, no existing behavour changes

```release-note

```


pkg/deploy/lattice/service_network_manager_test.go
```diff
@@ -284,18 +287,267 @@ func Test_CreateServiceNetwork_MeshAlreadyExist_ServiceNetworkVpcAssociationStat
...
 	meshManager := NewDefaultServiceNetworkManager(mockCloud)
-	resp, err := meshManager.Create(ctx, &meshCreateInput)
+	resp, err := meshManager.CreateOrUpdate(ctx, &meshCreateInput)
 
 	assert.Nil(t, err)
 	assert.Equal(t, resp.ServiceNetworkARN, meshArn)
 	assert.Equal(t, resp.ServiceNetworkID, meshId)
 }
 
-func Test_CreateServiceNetwork_MeshAlreadyExist_AssociateToNotAssociate(t *testing.T) {
```
zijun726911, line 298: Did the unit test name changes: from ` _CreateServiceNetwork_` to ` _CreateOrUpdateServiceNetwork_` i.e, from `Test_CreateServiceNetwork_MeshAlreadyExist_AssociateToNotAssociate ` to `Test_CreateOrUpdateServiceNetwork_MeshAlreadyExist_AssociateToNotAssociate`

pkg/deploy/lattice/service_network_manager.go
```diff
@@ -53,66 +57,72 @@ type defaultServiceNetworkManager struct {
 // return errors.New(LATTICE_RETRY) when:
 //
 //	CreateServiceNetworkVpcAssociationInput returns ServiceNetworkVpcAssociationStatusFailed/ServiceNetworkVpcAssociationStatusCreateInProgress/MeshVpcAssociationStatusDeleteInProgress
-func (m *defaultServiceNetworkManager) Create(ctx context.Context, service_network *latticemodel.ServiceNetwork) (latticemodel.ServiceNetworkStatus, error) {
+func (m *defaultServiceNetworkManager) CreateOrUpdate(ctx context.Context, stackSn *latticemodel.ServiceNetwork) (latticemodel.ServiceNetworkStatus, error) {
 	// check if exists
-	service_networkSummary, err := m.cloud.Lattice().FindServiceNetwork(ctx, service_network.Spec.Name, "")
+	service_networkSummary, err := m.cloud.Lattice().FindServiceNetwork(ctx, stackSn.Spec.Name, "")
```
mikhail-aws, line 62: please use camel case

pkg/deploy/lattice/service_network_manager.go
```diff
@@ -53,66 +57,72 @@ type defaultServiceNetworkManager struct {
 // return errors.New(LATTICE_RETRY) when:
 //
 //	CreateServiceNetworkVpcAssociationInput returns ServiceNetworkVpcAssociationStatusFailed/ServiceNetworkVpcAssociationStatusCreateInProgress/MeshVpcAssociationStatusDeleteInProgress
-func (m *defaultServiceNetworkManager) Create(ctx context.Context, service_network *latticemodel.ServiceNetwork) (latticemodel.ServiceNetworkStatus, error) {
+func (m *defaultServiceNetworkManager) CreateOrUpdate(ctx context.Context, stackSn *latticemodel.ServiceNetwork) (latticemodel.ServiceNetworkStatus, error) {
```
mikhail-aws, line 60: why stackSn, not just sn?

pkg/deploy/lattice/service_network_manager.go
```diff
@@ -53,66 +57,72 @@ type defaultServiceNetworkManager struct {
...
 		return latticemodel.ServiceNetworkStatus{ServiceNetworkARN: "", ServiceNetworkID: ""}, err
 	}
 
 	// pre declaration
 	var service_networkID string
 	var service_networkArn string
-	var isServiceNetworkAssociatedWithVPC bool
-	var service_networkAssociatedWithCurrentVPCId *string
+	var isServiceNetworkAlreadyAssociatedWithCurrentVpc bool
+	var snvaAssociatedWithCurrentVPC *vpclattice.ServiceNetworkVpcAssociationSummary
```
mikhail-aws, line 71: you can simplify names for local variables

pkg/utils/common.go
```diff
@@ -36,3 +37,29 @@ func SliceFilter[T any](in []T, f FilterFunc[T]) []T {
...
+		} else {
+			map1["<nil>"]++
+		}
+	}
+
+	for _, str := range arr2 {
+		if str != nil {
+			map2[*str]++
+		} else {
+			map2["<nil>"]++
```
scottlaiaws, line 61: is this necessary if the `str` is `nil`? Since you compared the length of the 2 slices anyway. You can just not populate `map1` and `map2` if `str` is nil

pkg/deploy/lattice/service_network_manager.go
```diff
@@ -53,66 +57,72 @@ type defaultServiceNetworkManager struct {
...
 		}
-		service_networkInput.Tags[latticemodel.K8SServiceNetworkOwnedByVPC] = &config.VpcID
+		serviceNetworkInput.Tags[latticemodel.K8SServiceNetworkOwnedByVPC] = &config.VpcID
 
-		glog.V(2).Infof("Create service_network >>>> req[%v]", service_networkInput)
-		resp, err := vpcLatticeSess.CreateServiceNetworkWithContext(ctx, &service_networkInput)
-		glog.V(2).Infof("Create service_network >>>> resp[%v], err : %v", resp, err)
+		glog.V(2).Infof("CreateOrUpdate stackSn >>>> req[%v]", serviceNetworkInput)
+		resp, err := vpcLatticeSess.CreateServiceNetworkWithContext(ctx, &serviceNetworkInput)
+		glog.V(2).Infof("CreateOrUpdate stackSn >>>> resp[%v], err : %v", resp, err)
```
mikhail-aws, line 86: please cleanup logs

pkg/deploy/lattice/service_network_manager.go
```diff
@@ -53,66 +57,72 @@ type defaultServiceNetworkManager struct {
...
 		service_networkID = aws.StringValue(resp.Id)
 		service_networkArn = aws.StringValue(resp.Arn)
-		isServiceNetworkAssociatedWithVPC = false
-		glog.V(6).Infof(" ServiceNetwork Create API resp [%v]\n", resp)
+		isServiceNetworkAlreadyAssociatedWithCurrentVpc = false
+		glog.V(6).Infof(" ServiceNetwork CreateOrUpdate API resp [%v]\n", resp)
 
-	} else {
-		glog.V(6).Infof("service_network[%v] exists, further check association", service_network)
+	} else { //service_networkSummary != nil
```
scottlaiaws, line 96: nit: unnecessary comment

mikhail-aws, line 96: redundant comment

pkg/deploy/lattice/service_network_manager.go
```diff
@@ -274,22 +285,78 @@ func (m *defaultServiceNetworkManager) isServiceNetworkAssociatedWithVPC(ctx con
...
-					return true, r.Id, resp, errors.New(LATTICE_RETRY)
+					return true, r, resp, errors.New(LATTICE_RETRY)
 				}
 			}
 		}
 	}
 	return false, nil, resp, err
 }
+
+func (m *defaultServiceNetworkManager) UpdateServiceNetworkVpcAssociationIfNeeded(ctx context.Context, snId string, snArn string, desiredSN *latticemodel.ServiceNetwork, existingSnvaId *string) (latticemodel.ServiceNetworkStatus, error) {
```
mikhail-aws, line 308: please simplify names. What's the difference between desiredSn and snId, snArn, are they different SN's? Function arguments are confusing

pkg/deploy/lattice/service_network_manager.go
```diff
@@ -274,22 +285,78 @@ func (m *defaultServiceNetworkManager) isServiceNetworkAssociatedWithVPC(ctx con
...
+		ServiceNetworkVpcAssociationIdentifier: existingSnvaId,
+	})
+	if err != nil {
+		return latticemodel.ServiceNetworkStatus{}, err
+	}
+
+	if len(desiredSN.Spec.SecurityGroupIds) == 0 && retrievedSnva.SecurityGroupIds != nil && len(retrievedSnva.SecurityGroupIds) > 0 {
+		// VPC latiice API have a limitation that the user can't remove all security groups if current snva already has security groups.
+		// If that case happen, the controller will not delete snva. User should manually delete it.
+		//https://docs.aws.amazon.com/vpc-lattice/latest/ug/security-groups.html#security-groups-rules
```
mikhail-aws, line 319: this use-case should print error log, so operator can read log and do action

pkg/deploy/lattice/service_network_manager.go
```diff
@@ -53,66 +57,72 @@ type defaultServiceNetworkManager struct {
...
+	} else { //service_networkSummary != nil
+		glog.V(6).Infof("stackSn[%v] exists, further check association", stackSn)
 		service_networkID = aws.StringValue(service_networkSummary.SvcNetwork.Id)
 		service_networkArn = aws.StringValue(service_networkSummary.SvcNetwork.Arn)
-		isServiceNetworkAssociatedWithVPC, service_networkAssociatedWithCurrentVPCId, _, err = m.isServiceNetworkAssociatedWithVPC(ctx, service_networkID)
+		isServiceNetworkAlreadyAssociatedWithCurrentVpc, snvaAssociatedWithCurrentVPC, _, err = m.isServiceNetworkAlreadyAssociatedWithVPC(ctx, service_networkID)
 		if err != nil {
 			return latticemodel.ServiceNetworkStatus{ServiceNetworkARN: "", ServiceNetworkID: ""}, err
 		}
+		if stackSn.Spec.AssociateToVPC == true && isServiceNetworkAlreadyAssociatedWithCurrentVpc == true &&
```
scottlaiaws, line 104: maybe you can refactor this out into a separate variable to make the condition more clear

pkg/deploy/lattice/service_network_manager.go
```diff
@@ -53,66 +57,72 @@ type defaultServiceNetworkManager struct {
...
+			snvaAssociatedWithCurrentVPC.Status != nil && aws.StringValue(snvaAssociatedWithCurrentVPC.Status) == vpclattice.ServiceNetworkVpcAssociationStatusActive {
+			return m.UpdateServiceNetworkVpcAssociationIfNeeded(ctx, service_networkID, service_networkArn, stackSn, snvaAssociatedWithCurrentVPC.Id)
+
+		}
 	}
 
-	if service_network.Spec.AssociateToVPC == true {
-		if isServiceNetworkAssociatedWithVPC == false {
+	if stackSn.Spec.AssociateToVPC == true {
+		if isServiceNetworkAlreadyAssociatedWithCurrentVpc == false {
```
scottlaiaws, line 112: nit: you can use the short version of the condition in if statements. like `if condition` or `if !condition`

pkg/utils/common.go
scottlaiaws: Also add unit tests for this method

pkg/deploy/lattice/service_network_manager.go
```diff
@@ -274,22 +285,78 @@ func (m *defaultServiceNetworkManager) isServiceNetworkAssociatedWithVPC(ctx con
...
+		// If that case happen, the controller will not delete snva. User should manually delete it.
+		//https://docs.aws.amazon.com/vpc-lattice/latest/ug/security-groups.html#security-groups-rules
+
+		// TODO: Add validation webhook or update gateway (or VpcAssociationPolicy) status to indicate that error.
+		//  https://gateway-api.sigs.k8s.io/geps/gep-713/#standard-status-condition
+		return latticemodel.ServiceNetworkStatus{
+			ServiceNetworkID:     snId,
+			ServiceNetworkARN:    snArn,
+			SnvaSecurityGroupIds: nil,
+		}, errors.New(LATTICE_RETRY)
```
mikhail-aws, line 327: when return error, dont populate fields for struct

pkg/deploy/lattice/service_network_manager.go
```diff
@@ -274,22 +285,78 @@ func (m *defaultServiceNetworkManager) isServiceNetworkAssociatedWithVPC(ctx con
...
+		SecurityGroupIds:                       desiredSN.Spec.SecurityGroupIds,
+	})
+	retSnStatus := latticemodel.ServiceNetworkStatus{
+		ServiceNetworkID:     snId,
+		ServiceNetworkARN:    snArn,
+		SnvaSecurityGroupIds: updateSnvaResp.SecurityGroupIds,
+	}
+	if err != nil {
+		return retSnStatus, err
+	}
```
mikhail-aws, line 352: wrong sequence for error handling

pkg/deploy/lattice/service_network_manager.go
```diff
@@ -274,22 +285,78 @@ func (m *defaultServiceNetworkManager) isServiceNetworkAssociatedWithVPC(ctx con
...
+	switch updateSnvaResp.Status {
+	case aws.String(vpclattice.ServiceNetworkVpcAssociationStatusActive):
+		return retSnStatus, nil
+	case aws.String(vpclattice.ServiceNetworkVpcAssociationStatusUpdateInProgress):
```
mikhail-aws, line 356: declared case, but no handling, remove or add implementation

pkg/deploy/lattice/service_network_manager.go
```diff
@@ -274,22 +285,78 @@ func (m *defaultServiceNetworkManager) isServiceNetworkAssociatedWithVPC(ctx con
...
+	retSnStatus := latticemodel.ServiceNetworkStatus{
+		ServiceNetworkID:     snId,
+		ServiceNetworkARN:    snArn,
+		SnvaSecurityGroupIds: updateSnvaResp.SecurityGroupIds,
+	}
+	if err != nil {
+		return retSnStatus, err
+	}
+	switch updateSnvaResp.Status {
+	case aws.String(vpclattice.ServiceNetworkVpcAssociationStatusActive):
```
mikhail-aws, line 354: I would reverse condition. If status is not active return error, and return normal response at the end of funciton

pkg/utils/common.go
```diff
@@ -36,3 +37,29 @@ func SliceFilter[T any](in []T, f FilterFunc[T]) []T {
...
+		}
+	}
+	return reflect.DeepEqual(map1, map2)
```
mikhail-aws, line 64: dont get it, reflect.DeepEqual(arr1, arr2) should return correct result

mikhail-aws (changes requested):
Code looks more complex than problem. Please simplify names, use normal error handling order, cleanup logs. Also it's good time to refactor create/update function into separate create and update. For refactoring please submit PR with refactoring first, and then another PR with new feature.

pkg/utils/common.go
scottlaiaws: I think this method ignore the ordering of the 2 slices (like a set)

mikhail-aws: It's even more confusing, we dont care about order, but care about duplicates and nil's

pkg/deploy/lattice/service_network_manager.go
zijun726911: Will remove it

pkg/deploy/lattice/service_network_manager.go
zijun726911: Hope to to represent a meaning that this a the desired SN that user intend to create (refering the name `stackDeployer`)

pkg/deploy/lattice/service_network_manager.go
zijun726911: Will change to error log

pkg/deploy/lattice/service_network_manager.go
zijun726911: Will, retrun value `latticemodel.ServiceNetworkStatus{}` only used by unit test. I intend to do assertion for this return value

pkg/deploy/lattice/service_network_manager.go
zijun726911: What it mean by wrong sequence?  it should just return `latticemodel.ServiceNetworkStatus{}, err `?

pkg/deploy/lattice/service_network_manager.go
mikhail-aws: when you do unit test for error, you dont need to do assertion on ServiceNetworkStatus struct, it has no meaning when error returned - garbage value

pkg/deploy/lattice/service_network_manager.go
mikhail-aws: you are trying to use updateSnvaResp.SecurityGroupIds even if error returned, please handle error before that
```
updateSnvaResp, err := UpdateServiceNetworkVpcAssociationWithContext(...)
if err != nil { ... }
retSnStatus := latticemodel.ServiceNetworkStatus{ ... }
```

pkg/deploy/lattice/service_network_manager.go
mikhail-aws: looks cryptic, sn or serviceNetwork was ok

pkg/utils/common.go
mikhail-aws: I dont think it's a generic function, please move it to the place where you do comparison for security groupIds

There is a library for generic slices operations, like equality and sort. https://pkg.go.dev/golang.org/x/exp@v0.0.0-20230905200255-921286631fa9/slices

you can combine with SliceMap from common.go to convert pointers to strings, for example:

```go
func securityGroupIdsEqual(arr1, arr2 []*string) bool {
  ids1 := slices.Sort(SliceMap(arr1, aws.String))
  ids2 := slices.Sort(SliceMap(arr2, aws.String))
  return slices.Equal(ids1, ids2)
}
```

I dont get intention for comparison, how "equal" id's should be. I would guess we compare only non empty strings, and non-nil pointers. Might be worth to filter them out, then sort or put into set. But using set here seems redundant since IDs suppose to be unique, isnt it?

pkg/deploy/lattice/service_network_manager.go
zijun726911: I found it have to differenate desiredSN and m.cloud.Lattice().FindServiceNetwork() retrieved SN, plan to change the name `stackSn` to `desiredSN`

mikhail-aws: I believe difference is apparent, Create function accept model of service network and tries to create or update lattice resources. There is no ambiguity what kind of service network it is.

pkg/utils/common.go
zijun726911: Thanks, changed to securityGroupIdsEqual()

pkg/deploy/lattice/service_network_manager.go
```diff
@@ -274,22 +287,84 @@ func (m *defaultServiceNetworkManager) isServiceNetworkAssociatedWithVPC(ctx con
...
+		}, nil
+	}
+
+	updateSnvaResp, err := m.cloud.Lattice().UpdateServiceNetworkVpcAssociationWithContext(ctx, &vpclattice.UpdateServiceNetworkVpcAssociationInput{
+		ServiceNetworkVpcAssociationIdentifier: existingSnvaId,
+		SecurityGroupIds:                       desiredSN.Spec.SecurityGroupIds,
+	})
+	if err != nil {
+		return latticemodel.ServiceNetworkStatus{}, err
+	}
```
mikhail-aws, line 346: few questions:
1. Why do we need to handle "remove all security groups". I believe lattice API will throw error if we try to do so. I would let it go through and fail. If one day we remove this limitation from Lattice this code will fail for no reason.
2. If there is no diff in security group ids, does lattice do anything if still make a update call? If not I will also remove diff check here and call lattice, and let lattice figure-out diff.

zijun726911: Will do some manual test for 1 and 2. For 2, I think the short circuit logic may needed since UpdateSNVA api has a pretty low throttling limit

mikhail-aws: Cant relate low throughput and duplicating validation logic. We should not duplicate logic/code that exists in Lattice.

pkg/deploy/lattice/service_network_manager.go
```diff
@@ -54,61 +58,61 @@ type defaultServiceNetworkManager struct {
...
 
 		serviceNetworkId = aws.StringValue(resp.Id)
 		serviceNetworkArn = aws.StringValue(resp.Arn)
-		isServiceNetworkAssociatedWithVpc = false
 	} else {
-		glog.V(6).Infof("serviceNetwork %s exists, checking its VPC association", serviceNetwork.Spec.Name)
-		serviceNetworkId = aws.StringValue(serviceNetworkSummary.SvcNetwork.Id)
-		serviceNetworkArn = aws.StringValue(serviceNetworkSummary.SvcNetwork.Arn)
-		isServiceNetworkAssociatedWithVpc, serviceNetworkAssociatedWithCurrentVpcId, _, err = m.isServiceNetworkAssociatedWithVPC(ctx, serviceNetworkId)
+		m.log.Debugf("desiredSn %s exists, checking its VPC association", desiredSn.Spec.Name)
```
xWink, line 95: Seems to have overwritten "service network" with "desiredSn"

pkg/deploy/lattice/service_network_manager.go
```diff
@@ -131,23 +135,24 @@ func (m *defaultServiceNetworkManager) Create(
...
 				m.log.Errorf("Failed to delete association for %s, with response %s and err %s",
-					serviceNetwork.Spec.Name, resp, err)
+					desiredSn.Spec.Name, resp, err)
 			}
 
 			// return retry and check later if disassociation workflow finishes
 			return latticemodel.ServiceNetworkStatus{ServiceNetworkARN: "", ServiceNetworkID: ""}, errors.New(LATTICE_RETRY)
 		}
-		m.log.Debugf("Created serviceNetwork %s without VPC association", serviceNetwork.Spec.Name)
+		m.log.Debugf("Created desiredSn %s without VPC association", desiredSn.Spec.Name)
```
xWink, line 155: Seems to have overwritten "service network" with "desiredSn"

pkg/deploy/lattice/service_network_manager.go
zijun726911: Will remove the `if len(desiredSN.Spec.SecurityGroupIds) == 0 && retrievedSnva.SecurityGroupIds != nil && len(retrievedSnva.SecurityGroupIds) > 0 {}`  block in the next revision, from my test, lattice api already have a validation to prevent to update to a empty sgIds

pkg/deploy/lattice/service_network_manager.go
zijun726911: It did a variable rename from `serviceNetwork` to `desiredSn `? do you also think it's not a good renaming?

pkg/deploy/lattice/service_network_manager.go
zijun726911: Changed to `serviceNetwork`

pkg/deploy/lattice/service_network_manager.go
zijun726911: Changed to `isSnAlreadyAssociatedWithCurrentVpc` and `snvaAssociatedWithCurrentVPC `

pkg/deploy/lattice/service_network_manager.go
zijun726911: changed to only assert err for unhappy path unit test

mikhail-aws (approved):

Issue #126, opened 2023-03-13
Logging Improvements


ellistarn:
### Log Clarity and Spam
This is the first thing we should address. Right now, it's challenging for an unfamiliar operator to understand the logs. Broadly, there's too much being emitted and too many raw objects exposed. Combing through the logs requires ignoring the vast majority of data that's written to the log.

Take for example the random snippit
```
I0312 22:30:13.317466   73652 rule_manager.go:348] findMatchingRule, rule prefix /ver1 does not match rule value /ver2
I0312 22:30:13.317481   73652 rule_manager.go:172] Rule-Create: need to BatchUpdate priority
I0312 22:30:13.317488   73652 rule_manager.go:175] Rule--Create, found existing matching rule &{{0xc000ca1340 AWS::VPCServiceNetwork::Rule rule-2} {volepond-6-wqmveaex5q default 80 HTTP HTTPRouteMatch /ver2 rule-2 {[0xc000eb8480]} 2023-03-12 22:30:11.105277 -0700 PDT m=+3037.917560676} <nil>} rulsStatus {arn:aws:vpc-lattice:us-west-2:767520670908:service/svc-040db2d76ea402e95/listener/listener-040c6949803c3a581/rule/rule-063f85bccb787bc45 rule-063f85bccb787bc45 1   true false}
I0312 22:30:13.317544   73652 rule_synthesizer.go:112] getSDKRules service: [0xc00075e580] err: <nil>
I0312 22:30:13.317550   73652 rule_synthesizer.go:116] getSDKRules, listener: [0xc0003799a0] err: <nil>
```

As an operator, there are a number of things that jump out:
* These are all INFO level logs -- do I need to know this?
* Why are there err: <nil> being printed? I don't need to know about errors unless they happen
* What is SDKRules and why do I need to be aware of this? 
* Why are naked pointers exposed `0xc0003799a0`
* Why are out of context variables exposed `true false`
* How does this relate to the lattice controller? Do I need to know about this? `BatchUpdate priority`

It's worth spending some time being thoughtful about the relationship between controller and operator, and making sure the story being told is comprehensible without knowing every line of code -- it's best if you assume zero implementation knowledge.

### Formatting 

Some objects are posted in expanded form 
```
I0312 23:11:43.320527   73652 target_group_manager.go:229] tagsOutput {
  Tags: {
    K8SServiceName: "servantshimmer-5-el6wlgodww",
    K8SHTTPRouteNamespace: "default",
[...]

PR #410, opened 2023-09-24
Logging Cleanup - Part 1

xWink:
**What type of PR is this?**

cleanup

#126 partially

- Removes use of glog in event handlers, listener and rule manager, and listener synthesizer
- Removes unnecessary logs, such as in the case of log n throw
- Cleans output of logs so that they contain less clutter, are grammatically correct, and specify more useful information
- General code cleanliness
- Re-adds `e2e-test` and `e2e-clean` to `make help` output, which appear to have been inadvertently removed?

```

Ran 32 of 32 Specs in 2139.930 seconds
--- PASS: TestIntegration (2140.87s)
PASS
ok      github.com/aws/aws-application-networking-k8s/test/suites/integration   2141.628s
```

**Automation added to e2e**:

No

No

No, only logging changes

```release-note

```

### DIFF
--- a/pkg/deploy/lattice/listener_manager.go
+++ b/pkg/deploy/lattice/listener_manager.go
@@ -4,14 +4,16 @@ import (
 	"context"
 	"errors"
 	"fmt"
+
 	"github.com/aws/aws-application-networking-k8s/pkg/aws/services"
+	"github.com/aws/aws-application-networking-k8s/pkg/utils/gwlog"
+
 	"strings"
 
-	"github.com/golang/glog"
-
-	"github.com/aws/aws-application-networking-k8s/pkg/utils"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/vpclattice"
+
+	"github.com/aws/aws-application-networking-k8s/pkg/utils"
 
 	lattice_aws "github.com/aws/aws-application-networking-k8s/pkg/aws"
 	"github.com/aws/aws-application-networking-k8s/pkg/latticestore"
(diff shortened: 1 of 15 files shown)


pkg/deploy/lattice/listener_manager.go
```diff
@@ -99,53 +109,46 @@ func (d *defaultListenerManager) Create(ctx context.Context, listener *latticemo
 	}
 
 	resp, err := d.cloud.Lattice().CreateListener(&listenerInput)
+	if err != nil {
+		d.log.Errorf("Error during creation of Listener %s-%s", listenerSpec.Name, listenerSpec.Namespace)
```
mikhail-aws, line 113: I would return error here

pkg/deploy/lattice/listener_manager.go
```diff
@@ -154,52 +157,42 @@ func (d *defaultListenerManager) List(ctx context.Context, serviceID string) ([]
...
+		if aws.Int64Value(r.Port) == port {
+			d.log.Debugf("Port %d already in use by listener %s for service %s", port, *r.Arn, serviceId)
+			return r, nil
 		}
-	} else {
-		return nil, err
 	}
 
-	return nil, errors.New("Listener does not exist")
+	return nil, errors.New("listener does not exist")
```
mikhail-aws, line 186: good to have listener details in error message, unless caller of this function wrap error with these details

pkg/deploy/lattice/listener_synthesizer.go
```diff
@@ -95,19 +102,18 @@ func (l *listenerSynthesizer) getSDKListeners(ctx context.Context) ([]*latticemo
 	var resService []*latticemodel.Service
 
 	err := l.stack.ListResources(&resService)
-	glog.V(6).Infof("service: %v ignore err: %v \n", resService, err)
+	l.log.Debugf("Ignoring error when listing services %s", err)
```
mikhail-aws, line 105: if err then log error

mikhail-aws (approved):

pkg/deploy/lattice/listener_manager.go
xWink: Would like to avoid behavioural changes in this PR, just improving logging/readability of code.

pkg/deploy/lattice/listener_synthesizer.go
xWink: Will do in part 2 PR, thanks for catching this!

pkg/deploy/lattice/listener_manager.go
mikhail-aws: in general for refactoring you are right, but in this particular example resp is used as non-nil value few lines below, it will panic on nil.

xWink: Fair point, fixed next PR!

pkg/deploy/lattice/listener_manager.go
xWink: Fixed next PR!

PR #413, opened 2023-09-26
Logging Cleanup - Part 2

xWink:
**What type of PR is this?**

cleanup

#126 partially

- Removes use of glog in dnsendpoint, listener, and service network managers, and listener synthesizer
- Removes unnecessary logs, such as in the case of log and throw
- Add checks for errors before logging errors
- Cleans output of logs so that they contain less clutter, are grammatically correct, and specify more useful information
- General code cleanliness

```
Ran 32 of 32 Specs in 2170.862 seconds
--- PASS: TestIntegration (2172.09s)
PASS
ok      github.com/aws/aws-application-networking-k8s/test/suites/integration   2172.843s
```

**Automation added to e2e**:

No

No

Only logging changes

```release-note

```

### DIFF
--- a/pkg/aws/gomock_reflect_3295326204/prog.go
+++ b/pkg/aws/gomock_reflect_3295326204/prog.go
@@ -1,0 +1,64 @@
+package main
+
+import (
+	"encoding/gob"
+	"flag"
+	"fmt"
+	"os"
+	"path"
+	"reflect"
+
+	"github.com/golang/mock/mockgen/model"
+
+	pkg_ "github.com/aws/aws-application-networking-k8s/pkg/aws"
+)
+
+var output = flag.String("output", "", "The output file name, or empty to use stdout.")
+
+func main() {
+	flag.Parse()
+
+	its := []struct {
+		sym string
+		typ reflect.Type
+	}{
+
+		{"Cloud", reflect.TypeOf((*pkg_.Cloud)(nil)).Elem()},
+	}
+	pkg := &model.Package{
+		// NOTE: This behaves contrary to documented behaviour if the
+		// package name is not the final component of the import path.
+		// The reflect package doesn't expose the package name, though.
+		Name: path.Base("github.com/aws/aws-application-networking-k8s/pkg/aws"),
+	}
+
+	for _, it := range its {
+		intf, err := model.InterfaceFromInterfaceType(it.typ)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Reflection: %v\n", err)
+			os.Exit(1)
+		}
+		intf.Name = it.sym
+		pkg.Interfaces = append(pkg.Interfaces, intf)
+	}
+
+	outfile := os.Stdout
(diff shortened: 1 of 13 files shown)


pkg/aws/gomock_reflect_3295326204/prog.go
xWink, line 1: Not entirely sure what generated these files, are we meant to add them to .gitignore?

mikhail-aws: it seems code generation was interrupted and left temp files, you need to remove them

pkg/deploy/externaldns/dnsendpoint_manager.go
```diff
@@ -36,11 +39,11 @@ func (s *defaultDnsEndpointManager) Create(ctx context.Context, service *lattice
 		Name:      service.Spec.Name + "-dns",
 	}
 	if service.Spec.CustomerDomainName == "" {
-		glog.V(2).Infof("Skipping creation of %s: detected no custom domain", namespacedName.String())
+		s.log.Debugf("Skipping creation of %s: detected no custom domain", namespacedName.String())
```
mikhail-aws, line 42: %s will use Stringer interface and call String(), no need to explicitly call String()

pkg/deploy/lattice/service_network_manager.go
```diff
@@ -53,209 +54,201 @@ type defaultServiceNetworkManager struct {
...
 		deleteInput := vpclattice.DeleteServiceNetworkInput{
-			ServiceNetworkIdentifier: &service_networkID,
+			ServiceNetworkIdentifier: &serviceNetworkId,
 		}
-		glog.V(2).Infof("DeleteServiceNetworkWithContext: service_network %v", snName)
 		resp, err := vpcLatticeSess.DeleteServiceNetworkWithContext(ctx, &deleteInput)
-		glog.V(2).Infof("DeleteServiceNetworkWithContext: service_network %v , resp %v, err %v \n", snName, resp, err)
 		if err != nil {
+			m.log.Debugf("Failed to delete service network %s due to %s", snName, resp)
 			return errors.New(LATTICE_RETRY)
```
mikhail-aws, line 233: you can return wrapped retry error, it will trigger retry also will print error cause to log in Retry handler
```go
return fmt.Errorf("%w: failed to delete service network: %s: %s", RetryErr, snName, resp.Status)
```

mikhail-aws (approved):

pkg/deploy/lattice/service_network_manager.go
mikhail-aws: RetryErr is a real thing btw, https://github.com/aws/aws-application-networking-k8s/blob/main/pkg/deploy/lattice/error.go

pkg/deploy/externaldns/dnsendpoint_manager.go
xWink: Thanks, will include this in part 3!

Issue #200, opened 2023-04-03
allow K8S user to specify whether to enable access log 


liwenwu-amazon:
Today, K8S lattice controller does `NOT` enable access log.  This needs to be enhanced to allow K8S user to specify their intend

bmihaescu:
@liwenwu-amazon  will you consider adding support for logs prefix as well in release v0.0.18?
The same as we do with ALB ingress: 

`alb.ingress.kubernetes.io/load-balancer-attributes: access_logs.s3.enabled=true,access_logs.s3.bucket={{.Values.bucketName}},access_logs.s3.prefix={{.Values.prefix}},idle_timeout.timeout_seconds=30,routing.http.drop_invalid_header_fields.enabled=true`

PR #424, opened 2023-10-04
Access Log Policy Controller

xWink:
**What type of PR is this?**

Feature

#200 partially

- Adds new controller for AccessLogPolicy
  - This controller enables creating and deleting AccessLogPolicy resources, and appropriately sets statuses and finalizers. It does not yet support creating VPC Lattice Access Log Subscriptions, and thus has no real functionality for customers, but that will come in bite-sized followup PRs.
- Adds example AccessLogPolicy yaml

#### Output of `kubectl describe alp test-policy` after creation:
```
Name:         test-policy
Namespace:    default
Labels:       <none>
Annotations:  <none>
API Version:  application-networking.k8s.aws/v1alpha1
Kind:         AccessLogPolicy
Metadata:
  Creation Timestamp:  2023-10-05T02:14:37Z
  Finalizers:
    accesslogpolicy.k8s.aws/resources
  Generation:        1
  Resource Version:  24456853
  UID:               a67bb8b1-7744-4380-a22c-f35bdee92f3f
Spec:
  Destination Arn:  arn:aws:s3:::my-bucket
  Target Ref:
    Group:  gateway.networking.k8s.io
    Kind:   Gateway
    Name:   my-hotel
Status:
  Conditions:
    Last Transition Time:  2023-10-05T02:14:37Z
    Message:               application-networking.k8s.aws/gateway-api-controller
    Observed Generation:   1
    Reason:                Accepted
    Status:                True
Events:                    <none>
```

#### Output after updating the destinationArn
```
Name:         test-policy
Namespace:    default
Labels:       <none>
Annotations:  <none>
API Version:  application-networking.k8s.aws/v1alpha1
Kind:         AccessLogPolicy
Metadata:
  Creation Timestamp:  2023-10-05T02:14:37Z
  Finalizers:
    accesslogpolicy.k8s.aws/resources
  Generation:        2
  Resource Version:  24584481
  UID:               a67bb8b1-7744-4380-a22c-f35bdee92f3f
Spec:
  Destination Arn:  arn:aws:s3:::my-bucket-1
  Target Ref:
    Group:  gateway.networking.k8s.io
    Kind:   Gateway
    Name:   my-hotel
Status:
  Conditions:
    Last Transition Time:  2023-10-05T02:14:37Z
    Message:               application-networking.k8s.aws/gateway-api-controller
    Observed Generation:   2
    Reason:                Accepted
    Status:                True
Events:                    <none>
```

```
Ran 32 of 32 Specs in 2227.128 seconds
--- PASS: TestIntegration (2228.35s)
PASS
ok      github.com/aws/aws-application-networking-k8s/test/suites/integration   2229.062s
```

**Automation added to e2e**:

No

No

Not yet, since use of the AccessLogPolicy doesn't achieve anything

```release-note

```


controllers/accesslogpolicy_controller.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	if !alp.DeletionTimestamp.IsZero() {
+		return r.reconcileDelete(ctx, alp)
+	} else {
+		return r.reconcileUpsert(ctx, alp)
+	}
+}
+
+func (r *accessLogPolicyReconciler) reconcileDelete(ctx context.Context, alp *anv1alpha1.AccessLogPolicy) error {
```
scottlaiaws, line 115: qq here: when you implement actual Access Log Subscription deletion in this method, how would you get the identifier to delete the access log?

xWink: When we receive a delete, we can see the resource being deleted, which will allow us to see the target reference of Gateway or Route resource.

We can list the customer's VPC Lattice Service Networks or Services and find the one whose name matches the Kubernetes resource being created/deleted.

Once we have the id of the VPC Lattice resource, we can use it to call ListAccessLogSubscriptions, and delete the one whose destinationArn matches the deleted AccessLogPolicy destinationArn.

zijun726911:
Before, we handle `TargetGroupPolicy` (model build and stack deploy) as part of route_controller.go and handle VpcAssociation as part of gateway_controller.go. How we decide to handle XXXXXPolicy in the `xxxxx_policy_controller` or in the existing `xxx_k8s_resource_controller`?

xWink:
> Before, we handle `TargetGroupPolicy` (model build and stack deploy) as part of route_controller.go and handle VpcAssociation as part of gateway_controller.go. How we decide to handle XXXXXPolicy in the `xxxxx_policy_controller` or in the existing `xxx_k8s_resource_controller`?

We opted to handle AccessLogPolicy as its own standalone resource in the Kubernetes setting because it's its own standalone resource in the VPC Lattice setting. Since Access Log Subscription has its own lifecycle, so to should Access Log Policy.

This simplifies handling some of the create/delete/update cases vs. only adding event handling to existing controllers, which would increase their complexities substantially.

controllers/accesslogpolicy_controller.go
scottlaiaws: Thanks

examples/access-log-policy-example.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: application-networking.k8s.aws/v1alpha1
+kind: AccessLogPolicy
+metadata:
+  name: test-policy
+spec:
+  targetRef:
+    group: gateway.networking.k8s.io
+    kind: Gateway
+    name: my-hotel
+  destinationArn: "arn:aws:s3:::my-bucket"
```
erikfuller, line 10: Just a general question on the feature. How are we planning on reporting back the actual log paths? From what I recall, logs are based on unique resource ids, not names. How will folks using the controller know which logs apply to which k8s resources?

xWink: Not 100% sure I follow, so please correct me if I'm going down the wrong path here. When a user creates a Gateway or Route, that resource is associated to 1 VPC Lattice Service Network or Service. When they create an AccessLogPolicy with a targetRef of a Gateway or Route, it's the equivalent of creating a VPC Lattice Access Log Subscription for the corresponding ServiceNetwork or Service.

So if I create Gateway `my-hotel`, which in VPC Lattice is Service Network `sn-12345678901234567`, and then create an AccessLogPolicy which refers to Gateway `my-hotel`, then it will create an AccessLogSubscription with resourceIdentifier `sn-12345678901234567`, and any access logs flowing through that Service Network will wind up in the log destination I specified in the AccessLogPolicy.

erikfuller: Right. So the path in S3 would look like [[1](https://docs.aws.amazon.com/vpc-lattice/latest/ug/monitoring-access-logs.html)]:

`[bucket]/[prefix]/AWSLogs/[accountId]/VpcLattice/AccessLogs/[region]/[YYYY/MM/DD]/[resource-id]/[accountId]_VpcLatticeAccessLogs_[region]_[resource-id]_YYYYMMDDTHHmmZ_[hash].json.gz
`

Let's assume I have a number of gateways I'm managing - how will I know the Lattice resource ID for each? For example, do we patch the gateway with the Lattice service network ID? I see it as a message here: https://github.com/aws/aws-application-networking-k8s/blob/main/controllers/gateway_controller.go#L279. Ideally, I would be able to find out the ID by doing something like ```kubectl get gateway```.

Similarly, it may be difficult to map from a k8s route to a log line, or from a log line back to the k8s objects, since these log entries also use unique IDs rather than k8s-based names. Maybe for now just enabling the logs is enough, then customers can use the Lattice API out of band to determine the ID. Longer term we should look a improving this experience.

xWink: Agreed, this would absolutely improve clarity for users. I'll be sure to include the ALS id in the AccessLogPolicy, and we can create an issue to add this behaviour to other resources.

xWink:
Are there any remaining concerns regarding this PR that need addressing?

zijun726911 (approved):
LGTM

PR #428, opened 2023-10-05
add default tags to all lattice resources

### LINKED ISSUE #392: Use managedBy tag for all resources created by gateway-api
We need to start using common tags for all resources managed by gateway-api.
It might be worth to populate tags automatically on Create/Update lattice calls, wrap lattice client.

Scope: add new tags to all resources that we create in Lattice.

ManagedBy tag:
https://github.com/aws/aws-application-networking-k8s/blob/main/pkg/aws/cloud.go#L14

mikhail-aws:
Note:
Add default tags to all lattice resources (managers). Major change in managers. Test changes are mostly layout and cloud mock removal.

closes #392

### DIFF
--- a/pkg/deploy/lattice/config_test.go
+++ b/pkg/deploy/lattice/config_test.go
@@ -1,0 +1,10 @@
+package lattice
+
+import pkg_aws "github.com/aws/aws-application-networking-k8s/pkg/aws"
+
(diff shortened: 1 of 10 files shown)


zijun726911 (approved):
LGTM, This code change is pretty straightforward

PR #430, opened 2023-10-11
Support Creating Access Log Policies

xWink:
**What type of PR is this?**

Feature

#200 partially

- Adds controller support for creating Access Log Policies
  - When an Access Log Policy is created with a Gateway targetRef, a VPC Lattice Access Log Subscription for the Gateway's corresponding Service Network is created
  - When an Access Log Policy is created with an HTTPRoute or GRPCRoute targetRef, a VPC Lattice Access Log Subscription for the Route's corresponding Service is created
  - The Access Log Policy's status is updated according to the result of the creation attempt
    - If creation succeeds, the Access Log Policy status is set to Accepted
    - If the Access Log Subscription destination does not exist, the Policy status is set to Invalid
    - If the targetRef's Group is not `gateway.networking.k8s.io`, the Policy status is set to Invalid
    - If the targetRef's Kind is not `Gateway`, `HTTPRoute`, or `GRPCRoute`, the Policy status is set to Invalid
    - If the targetRef already has an Access Log Policy for the same destination type, the Policy status is set to Conflicted
    - If the VPC Lattice resource (Service or Service Network) corresponding to the targetRef does not exist, the controller will enter a reconciliation loop until the resource exists or the Policy is updated/deleted
    - If the targetRef itself does not exist, the status of the Access Log Policy is set to TargetNotFound
- Added a suite of unit and e2e tests for the new controller, model builder, synthesizer, and manager logic
  - The new e2e test suite consists of 8 tests and adds about 10% to the total e2e test run time

Note: There was discussion about adding annotations to Access Log Policies in [this thread](https://github.com/aws/aws-application-networking-k8s/pull/424#discussion_r1347728197). I intend to add this logic in the upcoming Delete logic, as this PR is large enough as is and the Delete logic will be able to make use of the annotations.

```
Ran 41 of 41 Specs in 3060.228 seconds
SUCCESS! -- 41 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestIntegration (3061.28s)
PASS
ok      github.com/aws/aws-application-networking-k8s/test/suites/integration   3061.965s
```

**Automation added to e2e**:

E2E tests now rely on:
- S3 API (need to create/delete 1 Bucket)
- CloudWatch Logs API (need to create/delete 1 Log Group)
- Firehose API (need to create/delete 1 Delivery Stream)
- IAM API (need to create/delete 1 Role and attach an inline policy to it)

No

Yes, but this isn't ready for release yet since you can't update or delete these policies. Release note will be added as part of the final PR of this feature.

```release-note

```


pkg/gateway/model_build_access_log_subscription.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+type accessLogSubscriptionModelBuildTask struct {
+	stack                 core.Stack
+	accessLogPolicy       *anv1alpha1.AccessLogPolicy
+	accessLogSubscription *model.AccessLogSubscription
+}
+
+func (t *accessLogSubscriptionModelBuildTask) run(ctx context.Context) error {
+	sourceType := model.ServiceSourceType
```
zijun726911, line 57: One general comment, since we are doing a brand new e2e code path for `accessLogPolicy`, can we don't follow the old side-effect style code? I think old style is a little bit hard to trace the data flow (tho we already got used to it)

Can we try to use stateless functions as much as we can?

 for example:
`accessLogSubscriptionModelBuildTask` could don't have member variables `stack`, `accessLogPolicy ` , `accessLogSubscription`. In stead, use function input params and return vaule to represent them:
`func (t *accessLogSubscriptionModelBuildTask) run(ctx context.Context, accessLogPolicy       *anv1alpha1.AccessLogPolicy ) *model.AccessLogSubscription, error `
(And I think either `stack` or  `accessLogSubscription` just need one, don't need both)

Another example is `(s *accessLogSubscriptionSynthesizer) Synthesize()`, can we removed member variable `stack` of accessLogSubscriptionSynthesizer and put it in the input parameter?

`func (s *accessLogSubscriptionSynthesizer) Synthesize(ctx context.Context, accessLogSubscription *model.AccessLogSubscription) error `

xWink: While I agree a functional code path is ideal, I think we are at a point now where deviation from the existing style would require a much more substantial refactor of the codebase. I would rather keep things consistent and have predictable side-effects, than have multiple different controllers with wildly different styles of code

Perhaps we should create a refactoring ticket to migrate over if the existing code style is causing churn?

zijun726911: Ok, make sense to do code style refactoring in a seperate PR, I can create a issue for it

pkg/deploy/lattice/access_log_subscription_manager.go
```diff
@@ -0,0 +1,106 @@
...
+	for _, serviceNetworkSummary := range serviceNetworkSummaries {
+		if *serviceNetworkSummary.Name == name {
+			return serviceNetworkSummary, nil
+		}
+	}
+	return nil, services.NewNotFoundError("ServiceNetwork", name)
+}
+
+func getServiceWithName(ctx context.Context, vpcLatticeSess services.Lattice, name string) (*vpclattice.ServiceSummary, error) {
+	serviceSummaries, err := vpcLatticeSess.ListServicesAsList(ctx, &vpclattice.ListServicesInput{})
```
zijun726911, line 96: We have a FindService()  function, same with that one?
https://github.com/aws/aws-application-networking-k8s/blob/6cccaa51c8bd797a0576302bc4bb3757604c94fb/pkg/aws/services/vpclattice.go#L231

controllers/accesslogpolicy_controller.go
```diff
@@ -133,16 +140,24 @@ func (r *accessLogPolicyReconciler) reconcileUpsert(ctx context.Context, alp *an
 			alp.Spec.TargetRef.Kind, alp.Spec.TargetRef.Name)
 		err := r.updateAccessLogPolicyStatus(ctx, alp, gwv1alpha2.PolicyReasonTargetNotFound)
 		if err != nil {
```
zijun726911, line 142: Seems we have this behavior: `If the targetRef does not exist, the controller will enter a reconciliation loop until the targetRef exists or the Policy is updated/deleted`

One thing I am not sure is we return `retryErr==nil` with policy status `Accepted==false` with reason `TargetNotFound`, does the controller can keep the reconciliation loop?

pkg/deploy/lattice/access_log_subscription_manager.go
```diff
@@ -0,0 +1,106 @@
...
+		return nil, err
+	}
+
+	return &lattice.AccessLogSubscriptionStatus{
+		Arn: *createALSOutput.Arn,
+		Id:  *createALSOutput.Id,
+	}, nil
+}
+
+func getServiceNetworkWithName(ctx context.Context, vpcLatticeSess services.Lattice, name string) (*vpclattice.ServiceNetworkSummary, error) {
```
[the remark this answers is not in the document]
xWink: Thanks for pointing this out, wasn't aware it was an option! Will fix next commit

pkg/deploy/lattice/access_log_subscription_manager.go
xWink: Same as above!

pkg/deploy/lattice/access_log_subscription_manager.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func (m *defaultAccessLogSubscriptionManager) Create(
+	ctx context.Context,
+	accessLogSubscription *lattice.AccessLogSubscription,
+) (*lattice.AccessLogSubscriptionStatus, error) {
+	vpcLatticeSess := m.cloud.Lattice()
+
+	var resourceIdentifier string
+	if accessLogSubscription.Spec.SourceType == lattice.ServiceNetworkSourceType {
```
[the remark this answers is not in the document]
xWink: Sure, done next commit!
Note that I'll be using fmt.Errorf rather than NewInvalidError. This would be a case where the controller is improperly handling something, rather than the user input being invalid.

pkg/deploy/lattice/access_log_subscription_manager.go
```diff
@@ -0,0 +1,106 @@
...
+		resourceIdentifier = *service.Arn
+	}
+
+	createALSInput := &vpclattice.CreateAccessLogSubscriptionInput{
+		ResourceIdentifier: &resourceIdentifier,
+		DestinationArn:     &accessLogSubscription.Spec.DestinationArn,
+	}
+
+	createALSOutput, err := vpcLatticeSess.CreateAccessLogSubscriptionWithContext(ctx, createALSInput)
+	if err != nil {
```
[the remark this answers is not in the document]
xWink: Thanks for the suggestion, this looks much cleaner. Done next commit!

zijun726911:
Do you have a validation logic to reject (fast fail) AccessLogPolicy that it's targetRef.Kind != Gateway or HTTPRoute or GRPCRoute (also need to check targetRef.Group)?

Similar to that one: https://github.com/aws/aws-application-networking-k8s/blob/3f89fd8c39c66f54af349eb9823fc8baf1a5944c/controllers/eventhandlers/mapper.go#L98

controllers/accesslogpolicy_controller.go
xWink: My bad, I worded this poorly. Correct to

```
If the VPC Lattice resource corresponding to the targetRef does not exist, the controller will enter a reconciliation loop until the targetRef exists or the Policy is updated/deleted
```

If the targetRef itself does not exist, we just set the status to TargetNotFound. If the Service Network or Service does not exist, then we do the retry loop.

xWink:
> Do you have a validation logic to reject (fast fail) AccessLogPolicy that it's targetRef.Kind != Gateway or HTTPRoute or GRPCRoute (also need to check targetRef.Group)?
> 
> Similar to that one:
> 
> https://github.com/aws/aws-application-networking-k8s/blob/3f89fd8c39c66f54af349eb9823fc8baf1a5944c/controllers/eventhandlers/mapper.go#L98
> 
> (also, any chance to re-use `policyToTargetRefObj()` function? don't need to do it if its too complex or not fit your use case)

Good callout, I'll add logic to mark the status of the resource as Invalid if the Kind or Group is unsupported next commit.

controllers/accesslogpolicy_controller.go
zijun726911: Great, thanks for clarifying! 

For the behavior: `If the targetRef itself does not exist, the status of the Access Log Policy is set to TargetNotFound`

If the AccessLogPolicy k8s resource is created first, then later, corresponding targetRefObj (Gateway or HTTPRoute or GRPCRoute) is created, do we have a sort of `gwEventHandler.MapToAccessLogPolicy()` (`enqueueImpactedAccessLogPolicy()`) to make the controller do the `(r *accessLogPolicyReconciler) reconcile()` again? Assume you will do it in a separate task _[EKS][Access Logs] Handle Gateway and Route_ ?

xWink: For the case of reverse order creation, the reconciliation loop should handle it for us without trouble

pkg/deploy/lattice/access_log_subscription_synthesizer.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	var ret error = nil
+	for _, als := range accessLogSubscriptions {
+		if !als.Spec.IsDeleted {
+			s.log.Debugf("Started creating or updating access log subscription %s", als.ID())
+			_, err := s.accessLogSubscriptionManager.Create(ctx, als)
+			if err != nil {
+				s.log.Debugf("Synthesizing access log subscription %s failed due to %s", als.ID(), err)
+				ret = err
```
mikhail-aws, line 48: Print or return error. If you want to introduce a batch operation where each loop iteration can produce error that should not stop loop you need to collect errors. Right now in production mode only last error will be displayed even if there are more.

pkg/deploy/lattice/access_log_subscription_synthesizer_test.go
```diff
@@ -0,0 +1,73 @@
...
+	assert.Nil(t, err)
+}
+
+func Test_Synthesize_AccessLogSubscriptionSpecIsNotDeletedButErrorOccurs_ReturnsError(t *testing.T) {
+	ctx := context.TODO()
+	c := gomock.NewController(t)
+	defer c.Finish()
+	mockManager := NewMockAccessLogSubscriptionManager(c)
+	k8sClient := mockclient.NewMockClient(c)
+	builder := gateway.NewAccessLogSubscriptionModelBuilder(gwlog.FallbackLogger, k8sClient)
```
mikhail-aws, line 53: Please use sub-tests for common fixtures, it's painful to refactor. Already facing it in other places.
Also test names in double quotes can be "human" readable.

```go
func TestSynthesizeAccessLogSub( t *testing.T) {
        ctx := context.TODO()
	c := gomock.NewController(t)
	defer c.Finish()
	mockManager := NewMockAccessLogSubscriptionManager(c)
	k8sClient := mockclient.NewMockClient(c)
	builder := gateway.NewAccessLogSubscriptionModelBuilder(gwlog.FallbackLogger, k8sClient)

       t.Run("SpecIsNotDeleted_CreatesAccessLogSubscription", func(t *testing.T) {
       ...
       })

       t.Run("SpecIsNotDeletedButErrorOccurs_ReturnsError", func(t *testing.T) {
       ...
       })
}
```

pkg/deploy/stack_deployer.go
```diff
@@ -235,3 +234,29 @@ func (d *latticeTargetsStackDeployer) Deploy(ctx context.Context, stack core.Sta
 	}
 	return nil
 }
+
+type accessLogSubscriptionStackDeployer struct {
+	log                          gwlog.Logger
+	k8sClient                    client.Client
+	stack                        core.Stack
+	accessLogSubscriptionManager lattice.AccessLogSubscriptionManager
```
mikhail-aws, line 242: As long as there is single manager name can be simplified to manager, I dont see there will be multiple managers here.

pkg/gateway/model_build_access_log_subscription.go
```diff
@@ -0,0 +1,86 @@
...
+	 */
+	sourceName := string(t.accessLogPolicy.Spec.TargetRef.Name)
+	if sourceType == model.ServiceSourceType {
+		namespace := t.accessLogPolicy.Spec.TargetRef.Namespace
+		if namespace != nil {
+			sourceName = fmt.Sprintf("%s-%s", sourceName, string(*namespace))
+		} else {
+			sourceName = fmt.Sprintf("%s-default", sourceName)
+		}
+	}
```
mikhail-aws, line 74: Few questions:
1. When namespace is nil? why it's should be possible
2. Naming rules for Gateway and non-Gateway seems common problem, I dont think we should handle it here. There should be single place for this logic in the code base. I have a sense we have something like this already in other places.

pkg/gateway/model_build_access_log_subscription_test.go
```diff
@@ -0,0 +1,233 @@
...
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
mikhail-aws, line 11: apimachineryv1

pkg/gateway/model_build_access_log_subscription_test.go
```diff
@@ -0,0 +1,233 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	testclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/gateway-api/apis/v1alpha2"
```
mikhail-aws, line 15: gwv1alpha2

pkg/model/lattice/accesslogsubscription.go
```diff
@@ -0,0 +1,56 @@
...
+		Spec: AccessLogSubscriptionSpec{
+			SourceType:     sourceType,
+			SourceName:     sourceName,
+			DestinationArn: destinationArn,
+			IsDeleted:      isDeleted,
+		},
+		Status: nil,
+	}
+
+	stack.AddResource(als)
```
mikhail-aws, line 53: should not be here, please move stack mutation above

mikhail-aws (changes requested):

controllers/accesslogpolicy_controller.go
```diff
@@ -179,6 +194,33 @@ func (r *accessLogPolicyReconciler) targetRefExists(ctx context.Context, alp *an
...
+		r.eventRecorder.Event(alp, corev1.EventTypeWarning, k8s.AccessLogPolicyEventReasonFailedBuildModel,
+			fmt.Sprintf("Failed to build model due to %s", err))
+		return err
+	}
+
+	jsonStack, err := r.stackMarshaller.Marshal(stack)
+	if err != nil {
+		return err
+	}
+	r.log.Debugw("Successfully built model", "stack", jsonStack)
```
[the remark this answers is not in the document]
xWink: I've put it to use a few times here, one debug of the stack doesn't seem to cause too much noise imo

controllers/accesslogpolicy_controller.go
```diff
@@ -26,14 +26,18 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
+	builder2 "sigs.k8s.io/controller-runtime/pkg/builder"
```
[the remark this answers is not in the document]
xWink: Will fix next commit!

xWink:
Will also update the message field in the Policy's Status to include useful information during the error cases next commit

controllers/accesslogpolicy_controller.go
```diff
@@ -179,6 +194,33 @@ func (r *accessLogPolicyReconciler) targetRefExists(ctx context.Context, alp *an
...
+		return err
+	}
+	r.log.Debugw("Successfully built model", "stack", jsonStack)
+
+	if err := r.stackDeployer.Deploy(ctx, stack); err != nil {
+		return err
+	}
+	r.log.Debugw("successfully deployed model",
+		"stack", stack.StackID().Name+":"+stack.StackID().Namespace,
+	)
```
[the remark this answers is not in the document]
xWink: Fixed next commit

pkg/aws/services/vpclattice.go
```diff
@@ -44,6 +44,42 @@ func IsNotFoundError(err error) bool {
...
+	return fmt.Sprintf("%s %s had a conflict: %s", e.ResourceType, e.Name, e.Message)
+}
+
+func NewConflictError(resourceType string, name string, message string) error {
+	return &ConflictError{resourceType, name, message}
+}
+
+func IsConflictError(err error) bool {
+	conflictErr := &ConflictError{}
+	return errors.As(err, &conflictErr)
```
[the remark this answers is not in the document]
xWink: Fixed next commit!

pkg/deploy/stack_deployer.go
xWink: Fair enough, fixed next commit

controllers/accesslogpolicy_controller.go
```diff
@@ -163,8 +178,8 @@ func (r *accessLogPolicyReconciler) targetRefExists(ctx context.Context, alp *an
 
 	switch alp.Spec.TargetRef.Kind {
 	case "Gateway":
-		gateway := &gwv1beta1.Gateway{}
-		err = r.client.Get(ctx, targetRefNamespacedName, gateway)
+		gw := &gwv1beta1.Gateway{}
+		err = r.client.Get(ctx, targetRefNamespacedName, gw)
```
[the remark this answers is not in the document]
xWink: Good point, there are other errors that could come up here. Fixed next commit

pkg/gateway/model_build_access_log_subscription_test.go
xWink: Fixed next commit, thank you!

pkg/gateway/model_build_access_log_subscription_test.go
xWink: Fixed next commit, thank you!

pkg/model/lattice/accesslogsubscription.go
xWink: Fixed next commit

xWink:
Reran e2e tests with new changes

```
Ran 43 of 43 Specs in 2967.092 seconds
SUCCESS! -- 43 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestIntegration (2968.06s)
PASS
ok      github.com/aws/aws-application-networking-k8s/test/suites/integration   2968.977s
```

pkg/aws/services/vpclattice.go
```diff
@@ -26,6 +26,20 @@ type LatticeServiceNameProvider interface {
 	LatticeServiceName() string
 }
 
+type defaultLatticeServiceNameProvider struct {
+	name string
+}
+
+func NewLatticeServiceNameProvider(name string) LatticeServiceNameProvider {
```
zijun726911, line 33: Should be `NewDefaultLatticeServiceNameProvider `?  LatticeServiceNameProvider is a interface name

pkg/model/lattice/accesslogsubscription.go
```diff
@@ -0,0 +1,52 @@
...
+
+type AccessLogSubscription struct {
+	core.ResourceMeta `json:"-"`
+	Spec              AccessLogSubscriptionSpec    `json:"spec"`
+	Status            *AccessLogSubscriptionStatus `json:"status,omitempty"`
+}
+
+type AccessLogSubscriptionSpec struct {
+	SourceType     SourceType
+	SourceName     string
```
zijun726911, line 24: More cleaner to change to `SourceLatticeResourceType`, `SourceLatticeResourceName` ?

pkg/aws/services/vpclattice.go
zijun726911, line 33: Change name to `NewDefaultLatticeServiceNameProvider`? since LatticeServiceNameProvider is a function name.

pkg/model/lattice/accesslogsubscription.go
xWink: It's talking about AccessLogSubscription, which is already a Lattice resource

pkg/aws/services/vpclattice.go
xWink: Fixed next commit

zijun726911 (commented):
Overall LGTM, looking forward to the rest of Access Log code. 

Do you need to send e2etest at last in a separate PR?

Make sure to reply @mikhail-aws 's comments or make code change

test/suites/integration/access_log_policy_test.go
```diff
@@ -0,0 +1,651 @@
...
+		s3Client          *s3.S3
+		logsClient        *cloudwatchlogs.CloudWatchLogs
+		firehoseClient    *firehose.Firehose
+		iamClient         *iam.IAM
+		httpDeployment    *appsv1.Deployment
+		grpcDeployment    *appsv1.Deployment
+		httpK8sService    *corev1.Service
+		grpcK8sService    *corev1.Service
+		httpRoute         *gwv1beta1.HTTPRoute
+		grpcRoute         *gwv1alpha2.GRPCRoute
```
zijun726911, line 72: > The new e2e test suite consists of 8 tests and adds about 10% to the total e2e test run time

If you just create one of httpRoute/grpcRoute and create 2 AccessLogSubscriptions on it can it save some time? Or any other way to save e2etest time? (e.g., just test CloudWatchLogGroup type accessLog for gateway)

xWink:
> Overall LGTM, looking forward to the rest of Access Log code.
> 
> Do you need to send e2etest at last in a separate PR?
> 
> Make sure to reply @mikhail-aws 's comments or make code change

Thank you, I have one more commit on the way with the DefaultTags logic from #428 

e2e tests are in this PR

@mikhail-aws comments are all addressed in the above commits

test/suites/integration/access_log_policy_test.go
xWink: We need 1 of each source type and 1 of each destination type to confirm all the use-cases work, otherwise there is risk that our TargetRef identification or DestinationArn validation is broken without us knowing. I'll be adding Delete and Update e2e tests here in upcoming PRs, I think we will be able to squish some use-cases together into single tests to save time once that happens.

pkg/deploy/lattice/access_log_subscription_synthesizer.go
xWink: Currently, this debug logs every error, and returns the last one. The idea is that upper functions will return upon receiving an error, but we will have visibility of every error when in debug mode. Additionally, this pattern is consistent with other parts of the codebase. I can switch the log to an error level instead of debug so we have more visibility, but grouping all the errors into a slice wouldn't offer any additional benefits here

pkg/deploy/lattice/access_log_subscription_synthesizer_test.go
xWink: Thanks for showing this, looks much better. Will do next commit!

pkg/deploy/lattice/access_log_subscription_synthesizer.go
xWink: Good point, will return a combined error. Slices won't offer any additional benefits given how we use/propagate Synthesizer errors, but I can get all the different errors into a single message and return once.

mikhail-aws (approved):

Issue #379, opened 2023-09-01
[More e2etest cases]  TargetGroupPolicy test cases


zijun726911:

solmonk:
Fixed by #434

PR #434, opened 2023-10-12
test: add e2e tests for Target Group Policy CRD

scottlaiaws:
**What type of PR is this?** test

**Which issue does this PR fix**: #379

**What does this PR do / Why do we need it**: add e2e tests for Target Group Policy CRD

`make e2e-test` outputs:

```
Ran 38 of 38 Specs in 2715.256 seconds
SUCCESS! -- 38 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestIntegration (2716.14s)
PASS
ok      github.com/aws/aws-application-networking-k8s/test/suites/integration   2716.173s
```

**Automation added to e2e**: add `test/suites/integration/target_group_policy_test.go` 

**Will this break upgrades or downgrades. Has updating a running cluster been tested?**: N/A

### DIFF
--- a/test/suites/integration/target_group_policy_test.go
+++ b/test/suites/integration/target_group_policy_test.go
@@ -1,0 +1,144 @@
+package integration
+
+import (
+	"time"
+
+	"github.com/aws/aws-application-networking-k8s/pkg/apis/applicationnetworking/v1alpha1"
+	"github.com/aws/aws-application-networking-k8s/test/pkg/test"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/vpclattice"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	gwv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+)
+
+var _ = Describe("Target Group Policy Tests", Ordered, func() {
+	var (
+		deployment *appsv1.Deployment
+		service    *corev1.Service
+		httpRoute  *gwv1beta1.HTTPRoute
+		policy     *v1alpha1.TargetGroupPolicy
+	)
+
+	BeforeAll(func() {
+		deployment, service = testFramework.NewNginxApp(test.ElasticSearchOptions{
+			Name:      "target-group-policy-test",
+			Namespace: k8snamespace,
+		})
+
+		httpRoute = testFramework.NewHttpRoute(testGateway, service, service.Kind)
+
+		testFramework.ExpectCreated(ctx, deployment, service, httpRoute)
+	})
+
+	It("Update Protocol creates new Target Group", func() {
+		policy = testFramework.CreateTargetGroupPolicy(service, &test.TargetGroupPolicyConfig{
+			PolicyName: "test-policy",
+			Protocol:   aws.String(vpclattice.TargetGroupProtocolHttp),
+		})
+
+		testFramework.ExpectCreated(ctx, policy)
+
+		tg := testFramework.GetTargetGroup(ctx, service)
+
+		Expect(*tg.Protocol).To(Equal(vpclattice.TargetGroupProtocolHttp))
+
+		testFramework.UpdateTargetGroupPolicy(policy, &test.TargetGroupPolicyConfig{
+			Protocol: aws.String(vpclattice.TargetGroupProtocolHttps),
+		})
+
+		testFramework.ExpectUpdated(ctx, policy)
+
+		httpsTG := testFramework.GetTargetGroupWithProtocol(ctx, service, "https", "http1")
+
+		Expect(*httpsTG.Protocol).To(Equal(vpclattice.TargetGroupProtocolHttps))
+	})
+
+	It("Delete Target Group Policy reset health check config for HTTP and HTTP1 Target Group", func() {
+		policy = testFramework.CreateTargetGroupPolicy(service, &test.TargetGroupPolicyConfig{
+			PolicyName:      "test-policy",
+			Protocol:        aws.String(vpclattice.TargetGroupProtocolHttp),
+			ProtocolVersion: aws.String(vpclattice.TargetGroupProtocolVersionHttp1),
+			HealthCheck: &v1alpha1.HealthCheckConfig{
+				IntervalSeconds: aws.Int64(7),
+				StatusMatch:     aws.String("200,204"),
+			},
+		})
+
+		testFramework.ExpectCreated(ctx, policy)
+
+		// time.Sleep(10 * time.Second)
+
+		Eventually(func(g Gomega) {
+			tgSummary := testFramework.GetTargetGroup(ctx, service)
+
+			tg := testFramework.GetFullTargetGroupFromSummary(ctx, tgSummary)
+
+			g.Expect(*tg.Config.HealthCheck.ProtocolVersion).To(Equal(vpclattice.TargetGroupProtocolVersionHttp1))
+			g.Expect(*tg.Config.HealthCheck.Protocol).To(Equal(vpclattice.TargetGroupProtocolHttp))
+			g.Expect(*tg.Config.HealthCheck.HealthCheckIntervalSeconds).To(BeEquivalentTo(7))
+			g.Expect(*tg.Config.HealthCheck.Matcher.HttpCode).To(Equal("200,204"))
+		}).Within(10 * time.Second).WithPolling(1 * time.Second).Should(Succeed())
+
+		testFramework.ExpectDeletedThenNotFound(ctx, policy)
+
+		Eventually(func(g Gomega) {
+			tgSummary := testFramework.GetTargetGroup(ctx, service)
+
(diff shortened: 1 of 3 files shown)


test/suites/integration/target_group_policy_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+		Expect(*tg.Protocol).To(Equal(vpclattice.TargetGroupProtocolHttp))
+
+		testFramework.UpdateTargetGroupPolicy(policy, &test.TargetGroupPolicyConfig{
+			Protocol: aws.String(vpclattice.TargetGroupProtocolHttps),
+		})
+
+		testFramework.ExpectUpdated(ctx, policy)
+
+		httpsTG := testFramework.GetTargetGroupWithProtocol(ctx, service, "https", "http1")
```
zijun726911, line 62: Verify old tg is deleted?

test/suites/integration/target_group_policy_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+		Expect(*httpsTG.Protocol).To(Equal(vpclattice.TargetGroupProtocolHttps))
+	})
+
+	It("Delete Target Group Policy reset health check config for HTTP and HTTP1 Target Group", func() {
```
zijun726911, line 67: That can be a better name : "Delete Target Group Policy will reset health check config to default value" ?

test/suites/integration/target_group_policy_test.go
```diff
@@ -0,0 +1,130 @@
...
+		Expect(*updatedTG.Config.HealthCheck.HealthCheckTimeoutSeconds).To(BeEquivalentTo(5))
+		Expect(*updatedTG.Config.HealthCheck.HealthyThresholdCount).To(BeEquivalentTo(5))
+		Expect(*updatedTG.Config.HealthCheck.UnhealthyThresholdCount).To(BeEquivalentTo(2))
+		Expect(*updatedTG.Config.HealthCheck.Protocol).To(Equal(vpclattice.TargetGroupProtocolHttp))
+		Expect(*updatedTG.Config.HealthCheck.ProtocolVersion).To(Equal(vpclattice.TargetGroupProtocolVersionHttp1))
+		Expect(*updatedTG.Config.HealthCheck.Matcher.HttpCode).To(Equal("200"))
+		Expect(updatedTG.Config.HealthCheck.Port).To(BeNil())
+	})
+
+	It("Delete Target Group Policy create HTTP and HTTP1 Target Group", func() {
```
zijun726911, line 109: A better name ? "Delete Target Group Policy fallback to default  TargetGroupProtocol and TargetGroupProtocolVersion"

test/suites/integration/target_group_policy_test.go
```diff
@@ -0,0 +1,130 @@
...
+		policy = testFramework.CreateTargetGroupPolicy(service, &test.TargetGroupPolicyConfig{
+			PolicyName:      "test-policy",
+			Protocol:        aws.String(vpclattice.TargetGroupProtocolHttps),
+			ProtocolVersion: aws.String(vpclattice.TargetGroupProtocolVersionHttp2),
+		})
+
+		httpRoute = testFramework.NewHttpRoute(testGateway, service, service.Kind)
+
```
zijun726911, line 117: Do we really need to create httpRoute, deployment, service 3 times? can these 3 test cases reuse one  httpRoute, deployment, service to save some time?

test/suites/integration/target_group_policy_test.go
```diff
@@ -0,0 +1,130 @@
...
+		Expect(*tg.Protocol).To(Equal(vpclattice.TargetGroupProtocolHttp))
+
+		testFramework.UpdateTargetGroupPolicy(policy, &test.TargetGroupPolicyConfig{
+			Protocol: aws.String(vpclattice.TargetGroupProtocolHttps),
+		})
+
+		testFramework.ExpectUpdated(ctx, policy)
+
+		httpsTG := testFramework.GetTargetGroupWithProtocol(ctx, service, "https", "http1")
+
```
mikhail-aws, line 63: you can remove empty lines

test/suites/integration/target_group_policy_test.go
```diff
@@ -0,0 +1,130 @@
...
+		Expect(*updatedTG.Config.HealthCheck.Enabled).To(BeTrue())
+		Expect(*updatedTG.Config.HealthCheck.Path).To(Equal("/"))
+		Expect(*updatedTG.Config.HealthCheck.HealthCheckIntervalSeconds).To(BeEquivalentTo(30))
+		Expect(*updatedTG.Config.HealthCheck.HealthCheckTimeoutSeconds).To(BeEquivalentTo(5))
+		Expect(*updatedTG.Config.HealthCheck.HealthyThresholdCount).To(BeEquivalentTo(5))
+		Expect(*updatedTG.Config.HealthCheck.UnhealthyThresholdCount).To(BeEquivalentTo(2))
+		Expect(*updatedTG.Config.HealthCheck.Protocol).To(Equal(vpclattice.TargetGroupProtocolHttp))
+		Expect(*updatedTG.Config.HealthCheck.ProtocolVersion).To(Equal(vpclattice.TargetGroupProtocolVersionHttp1))
+		Expect(*updatedTG.Config.HealthCheck.Matcher.HttpCode).To(Equal("200"))
+		Expect(updatedTG.Config.HealthCheck.Port).To(BeNil())
```
mikhail-aws, line 106: Can you compare structs rather than each field of struct? Also when we reset health check config there should be some DefaultHealthCheckConfig struct in source code that we should compare with, otherwise we declare default behavior in 2 different places source and test.

test/suites/integration/target_group_policy_test.go
```diff
@@ -0,0 +1,130 @@
...
+		tg := testFramework.GetFullTargetGroupFromSummary(ctx, tgSummary)
+
+		Expect(*tg.Config.HealthCheck.ProtocolVersion).To(Equal(vpclattice.TargetGroupProtocolVersionHttp1))
+		Expect(*tg.Config.HealthCheck.Protocol).To(Equal(vpclattice.TargetGroupProtocolHttp))
+		Expect(*tg.Config.HealthCheck.HealthCheckIntervalSeconds).To(BeEquivalentTo(7))
+		Expect(*tg.Config.HealthCheck.Matcher.HttpCode).To(Equal("200,204"))
+
+		testFramework.ExpectDeletedThenNotFound(ctx, policy)
+
+		time.Sleep(2 * time.Second)
```
mikhail-aws, line 93: magic 2 seconds?

mikhail-aws (changes requested):

test/suites/integration/target_group_policy_test.go
scottlaiaws: Default Health Check config is different based on the Target Group Protocol (it's a private method that takes in `TargetGroupProtocol`)

mikhail-aws: ok, you can make it public, also dont compare each field of struct but compare structs

scottlaiaws: Since each field of a struct is a pointer, I can't use `Equal` to compare them.

Also for the default health check config method, it returned a struct with a "reset" value of `0` for each field instead the actual value, it's not possible to use that.

e.g. Our `UpdateTargetGroup` operation takes the following request:

```json
{
  "healthCheck": {
    "healthCheckIntervalSeconds": 0
  }
}
```

and then GetTargetGroup operation on the same target group will return

```json
{
  "healthCheck": {
    "healthCheckIntervalSeconds": 5
  }
}
```

mikhail-aws: > Since each field of a struct is a pointer, I can't use Equal to compare them.

Are you sure? assert.Equals de-reference pointers. Works fine for me

```go
type A struct {
	i int
	s string
	b *B
}

type B struct {
	i int
	s string
}

func TestA(t *testing.T) {
	a1 := &A{
		i: 10,
		s: "10",
		b: &B{
			i: 20,
			s: "20",
		},
	}
	a2 := &A{
		i: 10,
		s: "10",
		b: &B{
			i: 20,
			s: "20",
		},
	}
	assert.Equal(t, a1, a2)
}
m
```

mikhail-aws: reflect.DeepEqual de-reference pointers https://pkg.go.dev/reflect#DeepEqual

> Pointer values are deeply equal if they are equal using Go's == operator or if they point to deeply equal values. 

Code above also resolves into True, with reflect.DeepEqual(a1, a2)

test/suites/integration/target_group_policy_test.go
```diff
@@ -0,0 +1,144 @@
...
+		}).Within(10 * time.Second).WithPolling(1 * time.Second).Should(Succeed())
+
+		testFramework.ExpectDeletedThenNotFound(ctx, policy)
+
+		Eventually(func(g Gomega) {
+			tgSummary := testFramework.GetTargetGroup(ctx, service)
+
+			tg := testFramework.GetFullTargetGroupFromSummary(ctx, tgSummary)
+
+			g.Expect(*tg.Config.HealthCheck).To(Equal(vpclattice.HealthCheckConfig{
```
mikhail-aws, line 92: thank you

mikhail-aws (approved):

PR #436, opened 2023-10-13
Add ability to pass CLUSTER_NAME env variable in helm chart

### LINKED ISSUE #435: Helm chart failing to install when IMDS is unavailable
For the following configuration variables, we rely on IMDS to get the right values:
* `REGION`
* `AWS_ACCOUNT_ID`
* `CLUSTER_VPC_ID`
* `CLUSTER_NAME`

In many environments, IMDS is not available. (e.g. k8s w/o EC2 instance, self-managed clusters with IMDS access blocked) In this case, the values should be provided as environment variable through Helm chart. However, on [values.yaml](https://github.com/aws/aws-application-networking-k8s/blob/main/helm/values.yaml) we have 3 of these but `CLUSTER_NAME` is missing.

This is causing any Helm chart installation to fail in IMDS-disabled environment. One case reported is kops which [disables IMDS access for pods](https://kops.sigs.k8s.io/iam_roles/#instance-iam-roles) by default.

Path to fix:
* Add cluster name to helm chart value.
* Update documentation with proper IMDS setup instruction. Setting IMDS hops to 2 or higher is worth mentioning.

Temporary mitigation:
* If you are using IMDSv2, try increasing the hop limit to 2 or higher.
* Deploy manually. Modify `deploy-v0.0.17.yaml` and add above 4 environment variables.

zijun726911:
**What type of PR is this?** Fix #435 

Referred that PR https://github.com/aws/aws-application-networking-k8s/pull/264

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -81,9 +81,10 @@ Run through them again for a second cluster to use with the extended example sho
       oci://public.ecr.aws/aws-application-networking-k8s/aws-gateway-controller-chart\
       --version=v0.0.17 \
       --set=serviceAccount.create=false --namespace aws-application-networking-system \
-      # Region, clusterVpcId, awsAccountId are required for case where IMDS is NOT AVAILABLE, e.g Fargate
+      # awsRegion, clusterVpcId,clusterName, awsAccountId are required for case where IMDS is NOT AVAILABLE, e.g Fargate, self-managed clusters with IMDS access blocked 
       --set=awsRegion= \
       --set=clusterVpcId= \
+      --set=clusterName= \
       --set=awsAccountId= \
 
    ```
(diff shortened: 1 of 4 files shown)


docs/deploy.md
```diff
@@ -81,9 +81,10 @@ Run through them again for a second cluster to use with the extended example sho
       oci://public.ecr.aws/aws-application-networking-k8s/aws-gateway-controller-chart\
       --version=v0.0.17 \
       --set=serviceAccount.create=false --namespace aws-application-networking-system \
-      # Region, clusterVpcId, awsAccountId are required for case where IMDS is NOT AVAILABLE, e.g Fargate
+      # awsRegion, clusterVpcId,clusterName, awsAccountId are required for case where IMDS is NOT AVAILABLE, e.g Fargate, self-managed clusters with IMDS access blocked
```
solmonk, line 84: whitespace missing next to comma

solmonk (approved):

solmonk:
Thanks for working on this, but I think we should not close the referring issue until proper documentation is delivered.

mikhail-aws (approved):

PR #437, opened 2023-10-14
Added Support for Deleting Access Log Policies

xWink:
**What type of PR is this?**

feature

#200 partially

- Allows users to delete Access Log Policies (ALPs)
  - Deleting an ALP results in the underlying VPC Lattice Access Log Subscription (ALS) being deleted
  - If there is no underlying ALS, such as in the case of an ALP whose status is not Accepted, the ALP is deleted without attempting to delete an ALS
- Made `destinationArn` a mandatory property of `targetRef` in Access Log Policy CRD
- Upon successful creation of an ALP (before the status is set to Accepted), the controller now adds an annotation to the ALP
  - `VpcLatticeAccessLogSubscription`=`<ALS ARN>`
  - This annotation is used by the controller to determine whether an ALP was successfully created in the past (i.e. to differentiate between Create and Update events), and to identify the underlying ALS to update/delete without needing to call ListAccessLogSubscriptions based on the targetRef's id
- Added a tag on Access Log Subscription creation which indicates the namespaced name of the Access Log Policy it corresponds to
  - This is used to determine ownership of an ALS by the controller such that, in the event of a CreateAccessLogSubscription call being retried (i.e. the controller experiences an unexpected failure after the ALS was created, such as when updating the annotations), it will gracefully treat the creation as a success. Without this logic, the ALP status would be set to Conflicted due to an already existing ALS.

```
Ran 46 of 46 Specs in 2885.494 seconds
SUCCESS! -- 46 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestIntegration (2886.50s)
PASS
ok      github.com/aws/aws-application-networking-k8s/test/suites/integration   2887.317s
```

**Automation added to e2e**:

New ALP deletion e2e tests

No

Yes, but no release note yet since controller implementation for this new CRD is not complete

```release-note

```


pkg/apis/applicationnetworking/v1alpha1/accesslogpolicy_types.go
```diff
@@ -45,9 +46,8 @@ type AccessLogPolicySpec struct {
 	// Supported values are S3 Bucket, CloudWatch Log Group, and Firehose Delivery Stream ARNs.
 	//
 	// Changes to this value results in replacement of the VPC Lattice Access Log Subscription.
-	// +optional
 	// +kubebuilder:validation:Pattern=`^arn(:[a-z0-9]+([.-][a-z0-9]+)*){2}(:([a-z0-9]+([.-][a-z0-9]+)*)?){2}:([^/].*)?`
-	DestinationArn *string `json:"destinationArn,omitempty"`
+	DestinationArn *string `json:"destinationArn"`
```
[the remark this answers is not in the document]
xWink: Since we use pointer everywhere else, I opted to stay consistent with the other places. We also discussed offline and determined that there's no harm in keeping the pointer, as long as we remove omitempty.

pkg/deploy/lattice/access_log_subscription_manager_test.go
```diff
@@ -27,19 +29,37 @@ const (
...
+}
+
+func setup(t *testing.T) (
+	context.Context,
+	*services.MockLattice,
+	an_aws.Cloud,
+	services.Tags,
+) {
 	c := gomock.NewController(t)
 	defer c.Finish()
```
mikhail-aws, line 44: curious how test work since `defer c.Finish()` will be called after setup return but before test case starts

mikhail-aws: Also you dont need setup func, if you follow previously proposed approach with subtest. Here is [example](https://github.com/aws/aws-application-networking-k8s/blob/main/pkg/deploy/lattice/service_manager_test.go#L18-L31)

```go
func TestFoo (t *testing.T) {
  c := gomock.NewController(t)
  defer c.Finish()
  mockLattice := mocks.NewMockLattice(c)
       
  t.Run("Create", func(t *testing.T) {
    mockLattice.EXPECT()...
  })

    t.Run("Delete", func(t *testing.T) {
    mockLattice.EXPECT()...
  })
}

```

pkg/deploy/lattice/access_log_subscription_synthesizer.go
```diff
@@ -40,25 +38,33 @@ func (s *accessLogSubscriptionSynthesizer) Synthesize(ctx context.Context) error
...
+		case core.CreateEvent:
+			s.log.Debugf("Started creating Access Log Subscription %s", als.ID())
+			alsStatus, err := s.accessLogSubscriptionManager.Create(ctx, als)
 			if err != nil {
-				errMsg := fmt.Sprintf("synthesizing access log subscription %s failed due to: %s", als.ID(), err)
-				errs = append(errs, errMsg)
+				return err
+			} else {
+				als.Status = alsStatus
 			}
```
mikhail-aws, line 50: nit: not idiomatic go, dont use else on err nil check

pkg/model/lattice/accesslogsubscription.go
```diff
@@ -20,33 +25,36 @@ type AccessLogSubscription struct {
...
 
 func NewAccessLogSubscription(
 	stack core.Stack,
 	sourceType SourceType,
 	sourceName string,
 	destinationArn string,
-	isDeleted bool,
+	alpNamespacedName types.NamespacedName,
+	eventType core.EventType,
+	status *AccessLogSubscriptionStatus,
```
mikhail-aws, line 46: too many arguments, deserve a struct

pkg/model/core/resource.go
```diff
@@ -44,3 +44,11 @@ func (m *ResourceMeta) ID() string {
...
 	Visit(res Resource) error
 }
+
+type EventType string
+
+const (
+	CreateEvent EventType = "Create"
+	UpdateEvent EventType = "Update"
+	DeleteEvent EventType = "Delete"
+)
```
mikhail-aws, line 54: controller also defines event types, why we need own? https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.16.2/pkg/event

test/pkg/test/framework.go
```diff
@@ -519,3 +519,11 @@ func (env *Framework) RunGrpcurlCmd(opts RunGrpcurlCmdOptions) (string, string,
...
 	time.Sleep(30 * time.Second)
 }
+
+func (env *Framework) CopyDefaultTags() services.Tags {
+	tags := make(map[string]*string)
+	for k, v := range env.DefaultTags {
+		tags[k] = v
+	}
+	return tags
+}
```
mikhail-aws, line 529: I think following will be cleaner: change env.DefaultTags into env.Cloud, and call env.Cloud.DefaultTags() when we need tags, line 136

pkg/deploy/lattice/access_log_subscription_manager.go
```diff
@@ -59,31 +61,83 @@ func (m *defaultAccessLogSubscriptionManager) Create(
 		return nil, fmt.Errorf("unsupported source type: %s", accessLogSubscription.Spec.SourceType)
 	}
 
+	tags := m.cloud.DefaultTags()
```
mikhail-aws, line 64: I'm think we can benefit from new method in cloud that will take an argument of tags and merge with default tags. I can see more places that will need to define additional tags together with default ones, would be nice just to pass map of values.
for example:

```go
func (c *cloud) DefaultTagsMerge(tags Tags) Tags {
  return maps.Copy(c.DefalutTags(), tags)
}

m.cloud.DefaultTagsMerge({lattice.AccessLogPolicyTagKey}: aws.String(accessLogSubscription.Spec.ALPNamespacedName.String())})
```

test/suites/integration/access_log_policy_test.go
```diff
@@ -567,25 +583,140 @@ var _ = Describe("Creating Access Log Policy", Ordered, func() {
...
+				ResourceIdentifier: testServiceNetwork.Arn,
+			})
+			g.Expect(err).To(BeNil())
+			g.Expect(len(output.Items)).To(BeEquivalentTo(1))
+		}).Should(Succeed())
+
+		testFramework.ExpectDeleted(ctx, accessLogPolicy)
+
+		// Wait a moment for eventual consistency
+		time.Sleep(1)
```
mikhail-aws, line 616: interesting how 1 nanosecond changes situation

mikhail-aws (changes requested):

mikhail-aws:
> ALP
>     VpcLatticeAccessLogSubscription=<ALS ARN>
>     This annotation is used by the controller to determine whether an ALP was successfully created in the past (i.e. to differentiate between Create and Update events), and to identify the underlying ALS to update/delete without needing to call ListAccessLogSubscriptions based on the targetRef's id

I think this logic should be in the lattice manager. We always build model with latest k8s state, and let lattice manager figure-out should it update or create lattice resource.

controllers/accesslogpolicy_controller.go
```diff
@@ -226,25 +238,53 @@ func (r *accessLogPolicyReconciler) targetRefExists(ctx context.Context, alp *an
 func (r *accessLogPolicyReconciler) buildAndDeployModel(
 	ctx context.Context,
 	alp *anv1alpha1.AccessLogPolicy,
-) error {
+) (*core.Stack, error) {
```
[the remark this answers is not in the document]
xWink: Thank you, will fix next commit!

pkg/deploy/lattice/access_log_subscription_manager_test.go
xWink: Fixed next commit!

pkg/model/lattice/accesslogsubscription.go
xWink: Fixed next commit!

pkg/model/core/resource.go
xWink: The controller runtime event definitions are structs, and includes a GenericEvent. In order to use them, we'd have to create an interface for them and that would be the new EventType. Ultimately, it's more work with no benefit compared to using consts.

pkg/deploy/lattice/access_log_subscription_manager.go
xWink: Love this idea, will do next commit!

test/pkg/test/framework.go
xWink: Done next commit!

test/suites/integration/access_log_policy_test.go
xWink: Thanks for catching this, too used to Python. Fixed next commit :)

xWink:
> > ALP
> > VpcLatticeAccessLogSubscription=
> > This annotation is used by the controller to determine whether an ALP was successfully created in the past (i.e. to differentiate between Create and Update events), and to identify the underlying ALS to update/delete without needing to call ListAccessLogSubscriptions based on the targetRef's id
> 
> I think this logic should be in the lattice manager. Reconciler and Model builder does not care about lattice state. We always build model with latest k8s state, and let lattice manager figure-out should it update or create lattice resource.

The actual state of VPC Lattice is not what matters here, but the presence of an annotation in a k8s resource. I found this to be an incredible useful strategy for differentiating between Create and Update events, which then gets put to use in the Synthesizer. The ARN in the annotation is only ever used by the manager.

The overall ability to separate responsibilities between Synthesizer and Manager seems to be improved as well, though I could be wrong in my interpretation of which component is responsible for what, so please correct me if that's the case.

mikhail-aws:
> The overall ability to separate responsibilities between Synthesizer and Manager seems to be improved as well, though I could be wrong in my interpretation of which component is responsible for what, so please correct me if that's the case.

It's a convenience to implement, but it makes hard to reason about entire flow. I would like to keep each stage isolated with purpose it solves. 

Reconcile is a starting point for changes.
Synthesize or Build - build Lattice representation.
Manager call lattice API, will decide which API to call, create or update. Manager should fetch latest information from lattice about current state, does not rely on ARN from the k8s. To improve performance of fetching we will introduce cache on lattice calls later.

xWink:
> It's a good idea to backtrack lattice ARN's back to k8s for debugging and tracing purposes, but I dont think we should rely on it during reconciliation.
I'm curious as to why. If we say the recon isn't done until the annotation is set with the correct ARN, then a reconciled resource should have a reliable ARN in the annotation, why not use it? Same logic applies to using tags on VPC Lattice resources, no?

mikhail-aws:
My understanding of controller's runtime design that successful Reconcile function means it's success, having or not having ARN is not relevant here. My understanding that you use ARN from k8s to make decision about create or update, delete is not the case. I dont think we should do it at Reconciler stage.

It's a tightly coupled design (for example violates [LSP](https://en.wikipedia.org/wiki/Liskov_substitution_principle)) - means Reconciler knows something about how Lattice works, in particular we expect it will always have an ARN for this resource. It might work fine for this particular code path, but in larger scope it creates a "special case" and more special cases we have the harder to understand what controller or each component does.

xWink:
reran e2e tests with comments addressed

```
Ran 46 of 46 Specs in 2931.427 seconds
SUCCESS! -- 46 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestIntegration (2932.58s)
PASS
ok      github.com/aws/aws-application-networking-k8s/test/suites/integration   2933.415
```

xWink:
> It's a tightly coupled design (for example violates [LSP](https://en.wikipedia.org/wiki/Liskov_substitution_principle)) - means Reconciler knows something about how Lattice works, in particular we expect it will always have an ARN for this resource

The expectation from model builder is only on existence of an annotation, which is a K8s construct, not Lattice. Only the manager cares about the value of the ARN, and makes decisions based on that.

pkg/aws/cloud.go
```diff
@@ -97,6 +101,12 @@ func (c *defaultCloud) DefaultTags() services.Tags {
 	return tags
 }
 
+func (c *defaultCloud) DefaultTagsMergedWith(tags services.Tags) services.Tags {
+	newTags := c.DefaultTags()
+	maps.Copy(newTags, tags)
+	return newTags
+}
+
```
mikhail-aws, line 109: nit: does not hurt to have unit test

pkg/deploy/lattice/access_log_subscription_manager_test.go
```diff
@@ -27,372 +29,463 @@ const (
...
-			IsDeleted:      false,
-		},
-	}
-	serviceNetworkInfo := &services.ServiceNetworkInfo{
-		SvcNetwork: vpclattice.ServiceNetworkSummary{
-			Arn:  aws.String(serviceNetworkArn),
+	expectedTags := cloud.DefaultTags()
+	expectedTags[lattice.AccessLogPolicyTagKey] = aws.String(accessLogPolicyNamespacedName.String())
+
+	t.Run("Create_NewAccessLogSubscriptionForServiceNetwork_ReturnsSuccess", func(t *testing.T) {
```
mikhail-aws, line 46: neat

pkg/deploy/lattice/access_log_subscription_synthesizer.go
```diff
@@ -40,25 +38,32 @@ func (s *accessLogSubscriptionSynthesizer) Synthesize(ctx context.Context) error
...
+		case core.DeleteEvent:
+			s.log.Debugf("Started deleting Access Log Subscription %s", als.ID())
+			if als.Status != nil {
+				err := s.accessLogSubscriptionManager.Delete(ctx, als)
+				if err != nil {
+					return err
+				}
+			} else {
+				s.log.Debugf("Ignoring deletion of Access Log Subscription because als %s has no ARN", als.ID())
+			}
```
mikhail-aws, line 63: nit: try to avoid nested code

```go

s.log.Debugf("Started deleting Access Log Subscription %s", als.ID())
if als.Status == nil {
	s.log.Debugf("Ignoring deletion of Access Log Subscription because als %s has no ARN", als.ID())
	return nil
}
err := s.accessLogSubscriptionManager.Delete(ctx, als)
if err != nil {
	return err
}
```

pkg/deploy/lattice/access_log_subscription_synthesizer_test.go
```diff
@@ -78,4 +79,62 @@ func TestSynthesizeAccessLogSubscription(t *testing.T) {
 		err := synthesizer.Synthesize(ctx)
 		assert.NotNil(t, err)
 	})
+
+	t.Run("SpecIsDeleted_DeletesAccessLogSubscription", func(t *testing.T) {
+		ctx, c, mockManager, k8sClient, builder := setup()
+		defer c.Finish()
```
mikhail-aws, line 85: can you change this setup too as you did above? can be in another PR

pkg/gateway/model_build_access_log_subscription.go
```diff
@@ -72,9 +72,31 @@ func (t *accessLogSubscriptionModelBuildTask) run(ctx context.Context) error {
...
+	}
+
+	var status *model.AccessLogSubscriptionStatus
+	if eventType != core.CreateEvent {
+		value, exists := t.accessLogPolicy.Annotations[anv1alpha1.AccessLogSubscriptionAnnotationKey]
+		if exists {
+			status = &model.AccessLogSubscriptionStatus{
+				Arn: value,
+			}
+		}
```
mikhail-aws, line 89: if it's not CreateEvent the value should always exists, right? should we throw error here or log?

test/suites/integration/access_log_policy_test.go
```diff
@@ -567,25 +587,140 @@ var _ = Describe("Creating Access Log Policy", Ordered, func() {
...
+				ResourceIdentifier: testServiceNetwork.Arn,
+			})
+			g.Expect(err).To(BeNil())
+			g.Expect(len(output.Items)).To(BeEquivalentTo(1))
+		}).Should(Succeed())
+
+		testFramework.ExpectDeleted(ctx, accessLogPolicy)
+
+		// Wait a moment for eventual consistency
+		time.Sleep(1 * time.Second)
```
mikhail-aws, line 620: do you need sleep at all? it seemed to work fine in previous version with 1 nanosecond and can be removed

mikhail-aws (approved):
Still trying to digest ARN usage, not sure about it. Left minor comments.

test/suites/integration/access_log_policy_test.go
xWink: The test will be flakey if we remove this. There is a <1 second expectation for eventual consistency to catch up on the List API used here

pkg/aws/cloud.go
xWink: Good point, will add

pkg/deploy/lattice/access_log_subscription_manager_test.go
xWink: Fixed next commit

pkg/deploy/lattice/access_log_subscription_synthesizer.go
xWink: Fixed next commit!

pkg/gateway/model_build_access_log_subscription.go
xWink: Will throw here, good point

pkg/deploy/lattice/access_log_subscription_synthesizer_test.go
xWink: Yup, fixed next commit!

pkg/gateway/model_build_access_log_subscription.go
xWink: Correction, this is allowed to not exist because Creation might fail, resulting in annotation not being added, and user might try to delete the ALP. I'll add a debug log, rather than throw.

xWink:
One more e2e test run

```
Ran 46 of 46 Specs in 2767.409 seconds
SUCCESS! -- 46 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestIntegration (2768.73s)
PASS
ok      github.com/aws/aws-application-networking-k8s/test/suites/integration   2769.787s
```

PR #441, opened 2023-10-17
Generalize TargetRefExists() and UpdatePolicyStatus() methods for Policy CRDs

zijun726911:
**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/utils/policy/policy_test.go
+++ b/pkg/utils/policy/policy_test.go
@@ -1,0 +1,212 @@
+package policy
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gwv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+	gwv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+
+	mockclient "github.com/aws/aws-application-networking-k8s/mocks/controller-runtime/client"
+	anv1alpha1 "github.com/aws/aws-application-networking-k8s/pkg/apis/applicationnetworking/v1alpha1"
+	"github.com/aws/aws-application-networking-k8s/pkg/model/core"
+)
+
+func Test_checkGatewayOrRouteTargetRefExistsForPolicy(t *testing.T) {
+	testNamespace := gwv1beta1.Namespace("test-ns")
+
+	notFoundError := fmt.Errorf("k8s resource not found")
+	type args struct {
+		policy core.Policy
+	}
+	tests := []struct {
+		name                         string
+		args                         args
+		expectedK8sClientReturnedObj client.Object
+		want                         bool
+		wantErr                      error
+	}{
+		{
+			name: "AccessLogPolicy TargetRef gateway exists in the k8s, happy path",
+			args: args{
+				policy: &anv1alpha1.AccessLogPolicy{
+					TypeMeta:   metav1.TypeMeta{},
+					ObjectMeta: metav1.ObjectMeta{},
+					Spec: anv1alpha1.AccessLogPolicySpec{
+						DestinationArn: aws.String("destination-arn-123"),
+						TargetRef: &gwv1alpha2.PolicyTargetReference{
+							Group:     gwv1beta1.GroupName,
+							Kind:      "Gateway",
+							Name:      "test-gw",
+							Namespace: &testNamespace,
+						},
+					},
+				},
+			},
+			expectedK8sClientReturnedObj: &gwv1beta1.Gateway{
+				TypeMeta: metav1.TypeMeta{},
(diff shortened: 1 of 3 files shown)


pkg/utils/policy/policy_test.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	tests := []struct {
+		name                         string
+		args                         args
+		expectedK8sClientReturnedObj client.Object
+		want                         bool
+		wantErr                      error
+	}{
+		{
+			name: "AccessLogPolicy TargetRef gateway exists in the k8s, happy path",
```
xWink, line 37: nit: Name should describe expected outcome, rather than just say happy path

pkg/utils/policy/policy_test.go
```diff
@@ -0,0 +1,212 @@
...
+
+	notFoundError := fmt.Errorf("k8s resource not found")
+	type args struct {
+		policy core.Policy
+	}
+	tests := []struct {
+		name                         string
+		args                         args
+		expectedK8sClientReturnedObj client.Object
+		want                         bool
```
xWink, line 33: want what?

xWink: If this is just checking if we want the expected object to not be nil, shouldnt we be safe to assume that if the expectedK8sClientReturnedObj is nil, then an error is returned?

Also, expectedK8sClientReturnedObj is a bit long, can we shorten this to expectedObject?

pkg/utils/policy/policy_test.go
xWink: Same for all other happy path cases

mikhail-aws:
I dont like the direction of this PR. Not clear description, not clear why moving code into different place makes it cleaner/better/easier to understand. Dont see path forward.

mikhail-aws (changes requested):
Dont see improvements so far, except unit test coverage. I dont mind having better unit test coverage but please refactor this PR.

pkg/utils/policy/policy_test.go
zijun726911:
```
got, err := checkGatewayOrRouteTargetRefExistsForPolicy(ctx, k8sClient, tt.args.policy)
assert.Equalf(t, tt.want, got, "checkGatewayOrRouteTargetRefExistsForPolicy(%v, %v, %v)", ctx, k8sClient, tt.args.policy)
```
`want` is expected return value and `got` is actual return value, it's generated by goland idea any better name?

pkg/utils/policy/policy_test.go
zijun726911: Should I do "AccessLogPolicy TargetRef gateway exists in the k8s,  expect return true"? but 
```
want:    true,
wantErr: nil,
```
already indicate the expected outcome?

xWink: Yes, the name of the test should be indicative of what you want from the test, regardless of params. Reason is that if a test fails, I want to read the name of the test in the output and know why it failed without having to read a bunch of code

pkg/utils/policy/policy_test.go
xWink: I see, it's comparing expected/actual values. If this is a common pattern, then that's fine, I was just not familiar with it

mikhail-aws:
There might be some generic parts in two functions, it would be easier to seem them when we actually implement both of them, and common patterns emerge. Right now I dont see these common patterns in original code and this PR to make decision. These utility functions looks to specific to me, to be in generic utils.

zijun726911:
> There might be some generic parts in two: AccessLog and Auth, it would be easier to see them when we actually implement both of them, and common patterns emerge. Right now I dont see these common patterns in original code and this PR to make decision. These utility functions looks to specific to me, to be in generic utils.

My later PR will include that code in iamauthpolicy_controller.go:

```
func (r *authPolicyReconciler) reconcileUpsert(ctx context.Context, iap *anv1alpha1.IAMAuthPolicy) error {
	if err := r.finalizerManager.AddFinalizers(ctx, iap, authPolicyFinalizer); err != nil {
		r.eventRecorder.Event(iap, corev1.EventTypeWarning,
			k8s.IAMAuthPolicyEventReasonFailedAddFinalizer, fmt.Sprintf("Failed to add finalizer due to %s", err))
		return err
	}

	if !core.AuthPolicyTargetRefExists(ctx, r.client, r.log, iap) {
		r.log.Infof("Could not find Auth Log Policy targetRef %s %s",
			iap.Spec.TargetRef.Kind, iap.Spec.TargetRef.Name)
		err := utils_policy.UpdatePolicyStatus(ctx, r.client, iap, gwv1alpha2.PolicyReasonTargetNotFound)
		if err != nil {
			return fmt.Errorf("failed to update Auth Log Policy status, %w", err)
		}
		return nil
	}

	err := r.buildAndDeployModel(ctx, iap)
	if err != nil {
		// TODO: Error handling
		return utils_policy.UpdatePolicyStatus(ctx, r.client, iap, gwv1alpha2.PolicyReasonInvalid)
	}
	err = utils_policy.UpdatePolicyStatus(ctx, r.client, iap, gwv1alpha2.PolicyReasonAccepted)
	if err != nil {
		return fmt.Errorf("failed to update Access Log Policy status, %w", err)
	}
	return nil
}
```

That use `UpdatePolicyStatus` `AuthPolicyTargetRefExists`, should I include that in this PR?

PR #442, opened 2023-10-19
Added Support for Updating Access Log Policies

xWink:
**What type of PR is this?**

feature

#200 partially

- Enables support for updating Access Log Policies (ALPs)
  - When an ALP is updated to have a different `destinationArn` of the same destination type (i.e. S3 Bucket, CloudWatch Log Group, or Firehose Delivery Stream), the underlying Access Log Subscription (ALS) is updated
  - When an ALP is updated to have a different `destinationArn` of a different destination type, a new ALS is created to replace it, then the previous one is deleted
    - When an ALS is replaced, the ALP's `VpcLatticeAccessLogSubscription` annotation is updated with the new ARN
  - When an ALP is updated to have a different `targetRef`, a new ALS is created to replace the existing one, then the old one is deleted and the ALP's `VpcLatticeAccessLogSubscription` annotation is updated
  - When an ALP is updated to have a `destinationArn` for a destination that does not exist, the status of the ALP is set to Invalid
  - When an ALP is updated to have a `targetRef` for a non-existent target, the status of the ALP is set to TargetNotFound
  - When an ALP is updated to have a `destinationArn` for a destination type that is already in use for that target, the status of the ALP is set to Conflicted
  - Whenever an ALP update fails, the underlying ALS is not modified
    - This decision was made to reduce blast radius of mistaken updates, users should not lose their monitoring because of a typo in ALP spec
    - Status will still indicate the state of the current ALP configuration, regardless of underlying ALS
    - User can update the ALP again to fix it, or delete/recreate if they prefer

```
Ran 47 of 47 Specs in 3079.116 seconds
SUCCESS! -- 47 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestIntegration (3080.10s)
PASS
ok      github.com/aws/aws-application-networking-k8s/test/suites/integration   3081.006s
```

**Automation added to e2e**:

New test: `update properly changes or replaces Access Log Subscription and sets Access Log Policy status`
- This runs all the documented update scenarios and verifies that the controller behaves as expected for each of them (as described above)

No new dependencies, but one new CloudWatch Log Group is created/deleted as part of the e2e test. This is needed since we need two different destinations of the same type to verify the conflict scenario.

No.

Yes, but release note will be added in the final PR, which will include the API spec

```release-note

```

### DIFF
--- a/pkg/deploy/lattice/access_log_subscription_manager.go
+++ b/pkg/deploy/lattice/access_log_subscription_manager.go
@@ -18,6 +18,7 @@ import (
 
 type AccessLogSubscriptionManager interface {
 	Create(ctx context.Context, accessLogSubscription *lattice.AccessLogSubscription) (*lattice.AccessLogSubscriptionStatus, error)
+	Update(ctx context.Context, accessLogSubscription *lattice.AccessLogSubscription) (*lattice.AccessLogSubscriptionStatus, error)
 	Delete(ctx context.Context, accessLogSubscription *lattice.AccessLogSubscription) error
 }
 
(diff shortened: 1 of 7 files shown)


pkg/deploy/lattice/access_log_subscription_synthesizer.go
```diff
@@ -49,8 +49,12 @@ func (s *accessLogSubscriptionSynthesizer) Synthesize(ctx context.Context) error
 			als.Status = alsStatus
 		case core.UpdateEvent:
 			s.log.Debugf("Started updating Access Log Subscription %s", als.ID())
-			// TODO
-			return nil
+			alsStatus, err := s.accessLogSubscriptionManager.Update(ctx, als)
+			if err != nil {
+				return err
+			} else {
+				als.Status = alsStatus
```
xWink, line 56: Just noticed this could be denested, will fix next commit alongside any other feedback

pkg/deploy/lattice/access_log_subscription_manager.go
```diff
@@ -126,6 +127,47 @@ func (m *defaultAccessLogSubscriptionManager) Create(
...
+			if err != nil {
+				return nil, err
+			}
+			return alsStatus, nil
+		}
+	}
+
+	return &lattice.AccessLogSubscriptionStatus{
+		Arn: *updateALSOutput.Arn,
+	}, nil
```
mikhail-aws, line 168: Since you are getting into de-nesting code. You can change this one too.

```go
if err == nil {
     return &lattice.AccessLogSubscriptionStatus{
         Arn: *updateALSOutput.Arn,
     }, nil
}
swtich ...
```

xWink: Good point, will do

xWink:
Had to fixed some git issues, but the file changes should be good now. Addressed above comments.

pkg/deploy/lattice/access_log_subscription_manager_test.go
```diff
@@ -438,6 +435,200 @@ func TestAccessLogSubscriptionManager(t *testing.T) {
...
+		}
+		updateALSInput := &vpclattice.UpdateAccessLogSubscriptionInput{
+			AccessLogSubscriptionIdentifier: aws.String(accessLogSubscriptionArn),
+			DestinationArn:                  aws.String(s3DestinationArn),
+		}
+		updateALSError := &vpclattice.ResourceNotFoundException{
+			ResourceType: aws.String("ACCESS_LOG_SUBSCRIPTION"),
+		}
+
+		mockLattice.EXPECT().UpdateAccessLogSubscriptionWithContext(ctx, updateALSInput).Return(nil, updateALSError)
```
mikhail-aws, line 539: Just a thought, which can reduce code for unit tests, makes it easier to read and change in future. 

In this mock we do 2 things, 1. assert input into lattice client is correct and 2. it produces output which use later in manager.

And every test here asserts that input to lattice is correct. I dont think we need that, we can create a single test that makes sure we call Lattice with expected fields populated. For remaining tests we use any() in place of input. For example:

```go
		mockLattice.EXPECT().
			FindService(gomock.Any(), gomock.Any()).
			Return(&vpclattice.ServiceSummary{
				Arn:  aws.String("svc-arn"),
				Id:   aws.String("svc-id"),
				Name: aws.String(svc.LatticeServiceName()),
			}, nil)
```

pkg/deploy/lattice/access_log_subscription_manager.go
```diff
@@ -72,58 +73,101 @@ func (m *defaultAccessLogSubscriptionManager) Create(
...
+		 * To gracefully handle this, we create a new ALS with the new destination, then delete the old one.
+		 */
+		alsStatus, err := m.Create(ctx, accessLogSubscription)
+		if err != nil {
+			return nil, err
+		}
+		err = m.Delete(ctx, accessLogSubscription)
+		if err != nil {
+			return nil, err
+		}
```
mikhail-aws, line 166: Comment and code contradicts to each other. From code it looks like we create and delete same accessLogSubscription, not "new" and "old" ones.

mikhail-aws (changes requested):

pkg/deploy/lattice/access_log_subscription_manager_test.go
mikhail-aws: it was a minor comment, I'm ok with current version

pkg/deploy/lattice/access_log_subscription_manager.go
xWink: Create uses the ALS Spec to call VPC Lattice Create API

Delete uses the ARN of the ALS to call VPC Lattice Delete API

I could change Delete function to simply accept an ARN, I think that would make more sense anyways

pkg/deploy/lattice/access_log_subscription_manager_test.go
xWink: I think I prefer the more explicit nature of the existing code

xWink:
New functionality:
- For create/update: if namespace of the policy does not match namespace of the targetRef, policy status is set to Invalid
- For update, previously, targetRef changes were ignored, this was a bug and is now fixed
  - Changing targetRef results in creating a new ALS and deleting the old one

mikhail-aws (approved):

PR #448, opened 2023-10-24
add IAM Auth Policy create/update/delete

mikhail-aws:
Note:
Add model and manager for IAM Auth Policy. Currently handles only happy path when SN or Svc exists. If Lattice targetRef does not exists will print NotFound exceptions and retry later.

Always change AuthPolicy for Sn and Svc to AWS_IAM on upsert.

Tests:
Created SN and Svc in Lattice console. Applied and deleted `examples/iam-auth-policy-example.yaml` for SN and Svc.

### DIFF
--- a/controllers/iamauthpolicy_controller.go
+++ b/controllers/iamauthpolicy_controller.go
@@ -61,6 +61,9 @@ func (c *IAMAuthPolicyController) Reconcile(ctx context.Context, req ctrl.Reques
 		} else {
 			reconcileFunc = c.upsertRoutePolicy
 		}
+	default:
+		c.log.Errorw("unsupported targetRef", "kind", kind, "req", req)
+		return ctrl.Result{RequeueAfter: time.Hour}, nil
 	}
 
 	latticeResourceId, err := reconcileFunc(ctx, k8sPolicy)
@@ -107,6 +110,10 @@ func (c *IAMAuthPolicyController) deleteGatewayPolicy(ctx context.Context, k8sPo
 		return "", err
 	}
 	err = c.policyMgr.Delete(ctx, snId)
+	if err != nil {
+		return "", err
+	}
+	err = c.policyMgr.DisableSnIAMAuth(ctx, snId)
 	if err != nil {
 		return "", err
 	}
@@ -158,6 +165,10 @@ func (c *IAMAuthPolicyController) deleteRoutePolicy(ctx context.Context, k8sPoli
 	if err != nil {
 		return "", err
 	}
+	err = c.policyMgr.DisableSvcIAMAuth(ctx, svcId)
+	if err != nil {
+		return "", err
+	}
 	return svcId, nil
 }
 
@@ -184,14 +195,4 @@ func (c *IAMAuthPolicyController) putPolicy(ctx context.Context, resId, policy s
 	}
 	_, err := c.policyMgr.Put(ctx, modelPolicy)
 	return err
-}
-
-func (c *IAMAuthPolicyController) reconcileGateway(ctx context.Context, policy *anv1alpha1.IAMAuthPolicy) error {
-	c.log.Debugw("reconcile gateway iam policy", "policy", policy)
-	return nil
-}
-
-func (c IAMAuthPolicyController) reconcileRoute(ctx context.Context, policy *anv1alpha1.IAMAuthPolicy) error {
-	c.log.Debugw("reconcile route iam policy", "policy", policy)
-	return nil
 }
--- a/controllers/iamauthpolicy_controller.go
+++ b/controllers/iamauthpolicy_controller.go
@@ -2,25 +2,32 @@ package controllers
 
 import (
 	"context"
-	"fmt"
+	"time"
 
 	anv1alpha1 "github.com/aws/aws-application-networking-k8s/pkg/apis/applicationnetworking/v1alpha1"
+	pkg_aws "github.com/aws/aws-application-networking-k8s/pkg/aws"
+	"github.com/aws/aws-application-networking-k8s/pkg/aws/services"
+	deploy "github.com/aws/aws-application-networking-k8s/pkg/deploy/lattice"
+	model "github.com/aws/aws-application-networking-k8s/pkg/model/lattice"
+	"github.com/aws/aws-application-networking-k8s/pkg/utils"
 	"github.com/aws/aws-application-networking-k8s/pkg/utils/gwlog"
 
-	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 type IAMAuthPolicyController struct {
-	log    gwlog.Logger
-	client client.Client
+	log       gwlog.Logger
+	client    client.Client
+	policyMgr deploy.IAMAuthPolicyManager
 }
 
-func RegisterIAMAuthPolicyController(log gwlog.Logger, mgr ctrl.Manager) error {
+func RegisterIAMAuthPolicyController(log gwlog.Logger, mgr ctrl.Manager, cloud pkg_aws.Cloud) error {
 	controller := &IAMAuthPolicyController{
-		log:    log,
-		client: mgr.GetClient(),
+		log:       log,
+		client:    mgr.GetClient(),
+		policyMgr: deploy.IAMAuthPolicyManager{Cloud: cloud},
 	}
 	err := ctrl.NewControllerManagedBy(mgr).
 		For(&anv1alpha1.IAMAuthPolicy{}).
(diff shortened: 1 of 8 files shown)


controllers/iamauthpolicy_controller.go
```diff
@@ -2,67 +2,80 @@ package controllers
...
-		err = c.reconcileGateway(ctx, policy)
-		break
-	case "HTTPRoute":
-	case "GRPCRoute":
-		err = c.reconcileRoute(ctx, policy)
-		break
-	default:
-		err = fmt.Errorf("unsupported targetRef type, req=%s, kind=%s",
-			req, policy.Spec.TargetRef.Kind)
+	authPolicyFinalizer := "iamauthpolicy.k8s.aws/resources"
```
xWink, line 53: Could be const

mikhail-aws: could be, but not used outside of this func

controllers/iamauthpolicy_controller.go
```diff
@@ -2,67 +2,80 @@ package controllers
...
+
+	if policySpec.DeletionTimestamp.IsZero() {
+		policyStatus, err := c.policyMgr.Put(ctx, policyModel)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		policySpec.Annotations["ARN"] = policyStatus.Resource.ARN.String()
```
xWink, line 60: It appears as though there is a convention in Kubernetes to set annotations with a prefix indicating the domain, such as what we use in ServiceImport: `multicluster.x-k8s.io`/<annotation-key>`.

> Automated system components (e.g. kube-scheduler, kube-controller-manager, kube-apiserver, kubectl, or other third-party automation) which add annotations to end-user objects must specify a prefix.

https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#:~:text=Syntax%20and%20character%20set

I'll make a change to fix this in AccessLogPolicy, since it's missed there too.

controllers/iamauthpolicy_controller.go
```diff
@@ -2,67 +2,80 @@ package controllers
...
+		}
+		c.log.Infof("updated policy for name=%s, latticeId=%s", policyModel.ResourceName, policyStatus.Resource.Id)
+	} else {
+		if controllerutil.ContainsFinalizer(policySpec, authPolicyFinalizer) {
+			err = c.policyMgr.Delete(ctx, policyModel)
+			if err != nil {
+				return ctrl.Result{}, pkg_aws.IgnoreNotFound(err)
+			}
+			controllerutil.RemoveFinalizer(policySpec, authPolicyFinalizer)
+			c.log.Infof("deleted policy for name=%s", policyModel.ResourceName)
```
xWink, line 72: Should we not be logging the policy namespaced name? Or is that the resourceName?

controllers/iamauthpolicy_controller.go
mikhail-aws: make sense, will address in following pr. What's our approach on this? "aws.vpc.lattice.arn"?

controllers/iamauthpolicy_controller.go
mikhail-aws: it's LatticeServiceName, name+namespace truncated
https://github.com/aws/aws-application-networking-k8s/blob/main/pkg/utils/common.go#L38

pkg/aws/cloud.go
```diff
@@ -128,3 +129,18 @@ func (c *defaultCloud) IsArnManaged(arn string) (bool, error) {
 func getManagedByTag(cfg CloudConfig) string {
 	return fmt.Sprintf("%s/%s/%s", cfg.AccountId, cfg.ClusterName, cfg.VpcId)
 }
+
+func IgnoreNotFound(err error) error {
```
xWink, line 133: Unit tests would be nice for these new utilities

pkg/aws/services/vpclattice.go
```diff
@@ -315,3 +316,7 @@ func accountIdMatches(accountId string, itemArn string) (bool, error) {
 
 	return accountId == parsedArn.AccountID, nil
 }
+
+func (d *defaultLattice) FindServiceByName(ctx context.Context, name string) (*vpclattice.ServiceSummary, error) {
```
xWink, line 320: Love this, thank you!

pkg/deploy/lattice/iamauthpolicy_manager.go
```diff
@@ -0,0 +1,89 @@
...
+		if err != nil {
+			return empty, err
+		}
+		arn, err := arn.Parse(*out.SvcNetwork.Arn)
+		if err != nil {
+			return empty, err
+		}
+		return model.IAMAuthPolicyResource{
+			Name: policy.ResourceName,
+			Id:   *out.SvcNetwork.Id,
```
xWink, line 67: If this is specifically the service network id, perhaps Id should be changed to ServiceNetworkId

pkg/deploy/lattice/iamauthpolicy_manager.go
```diff
@@ -0,0 +1,89 @@
...
+			ARN:  arn,
```
xWink, line 68: Do we need both the ARN and the id?

pkg/model/lattice/iamauthpolicy.go
```diff
@@ -0,0 +1,60 @@
...
+	anv1alpha1 "github.com/aws/aws-application-networking-k8s/pkg/apis/applicationnetworking/v1alpha1"
+	"github.com/aws/aws-application-networking-k8s/pkg/utils"
+	"github.com/aws/aws-sdk-go/aws/arn"
+)
+
+type AuthPolicyKind string
+
+const (
+	GatewayAuthPolicy AuthPolicyKind = "gateway"
+	ServiceAuthPolicy AuthPolicyKind = "service"
```
xWink, line 15: Isn't it Gateway, HTTPRoute, and GRPCRoute? Service is the VPC Lattice resource

xWink: Also, if we are using the K8s Kinds, it would be good to capitalize them so they can be safely compared with other Kind strings

pkg/model/lattice/iamauthpolicy.go
```diff
@@ -0,0 +1,60 @@
...
+
+type IAMAuthPolicyResource struct {
+	ARN  arn.ARN
+	Id   string
+	Name string
+}
+
+type IAMAuthPolicyStatus struct {
+	Resource IAMAuthPolicyResource
+	State    string
```
xWink, line 32: What's this?

xWink (changes requested):
Looks good for happy path, just a few small requests

controllers/iamauthpolicy_controller.go
xWink: `vpc-lattice.amazonaws.com/arn` would follow the pattern, though this might warrant more discussion since it's a one-way door.

Also, regarding the suffix, `arn` doesn't explain what arn it is. Is this the arn of the resource on which the policy applies? If yes, the common name across AWS for this is `resourceArn`.

pkg/model/lattice/iamauthpolicy.go
zijun726911: Should be LatticeServiceNetworkAuthPolicy and LatticeServiceAuthPolicy?

pkg/model/lattice/iamauthpolicy.go
```diff
@@ -0,0 +1,60 @@
...
+
+type IAMAuthPolicy struct {
+	Kind         AuthPolicyKind
+	ResourceName string
+	Policy       string
+}
+
+type IAMAuthPolicyResource struct {
+	ARN  arn.ARN
+	Id   string
```
zijun726911, line 26: I remember in the vpc lattice, AuthPolicy don't have ARN and ID?

controllers/iamauthpolicy_controller.go
mikhail-aws: will change ARN to ResourceId. `application-networking.k8s.aws/resourceId`

pkg/aws/cloud.go
mikhail-aws: ok

pkg/model/lattice/iamauthpolicy.go
mikhail-aws: IAMAuthPolicy - State. It's lattice state. I think our current alignment is to use xxxStatus models for lattice managers responses, so I would read it as it's state of IAMAuthPolicy in lattice

pkg/model/lattice/iamauthpolicy.go
mikhail-aws: probably better to remove arn and leave only ID, seems confusing having different identifiers

pkg/deploy/lattice/iamauthpolicy_manager.go
mikhail-aws: need to rework this part, will leave only Id

pkg/aws/cloud.go
mikhail-aws: removed

controllers/iamauthpolicy_controller.go
```diff
@@ -29,34 +36,156 @@ func RegisterIAMAuthPolicyController(log gwlog.Logger, mgr ctrl.Manager) error {
...
 
+func (c *IAMAuthPolicyController) handleFinalizer(ctx context.Context, k8sPolicy *anv1alpha1.IAMAuthPolicy) {
+	authPolicyFinalizer := "iamauthpolicy.k8s.aws/resources"
+	if k8sPolicy.DeletionTimestamp.IsZero() {
+		if !controllerutil.ContainsFinalizer(k8sPolicy, authPolicyFinalizer) {
+			controllerutil.AddFinalizer(k8sPolicy, authPolicyFinalizer)
+		}
+	} else {
+		if controllerutil.ContainsFinalizer(k8sPolicy, authPolicyFinalizer) {
+			controllerutil.RemoveFinalizer(k8sPolicy, authPolicyFinalizer)
```
zijun726911, line 99: When handling resouce deletion, we should do `RemoveFinalizer()` at very last, to avoid actual reconcile logic return err and k8s api server lose track of that IAMAuthPolicy resource?

For example, in our other controller:
https://github.com/aws/aws-application-networking-k8s/blob/658bb6e2cac0a8c5b36c6c0ac9a80a4d2487b412/controllers/route_controller.go#L195-L204

zijun726911 (commented):
Could you describe some manually tests you did for the PR?

Overall LGTM, This code got rid of our old cumbersome style: ModelBuilder, StackDeployer... Hopfully all our later controller code(that handle new kind of k8s resource) could do that. I really love this new version.

controllers/iamauthpolicy_controller.go
```diff
@@ -29,34 +36,156 @@ func RegisterIAMAuthPolicyController(log gwlog.Logger, mgr ctrl.Manager) error {
...
+		return "", err
+	}
+	return *svcInfo.Id, nil
+}
+
+func (c *IAMAuthPolicyController) deleteRoutePolicy(ctx context.Context, k8sPolicy *anv1alpha1.IAMAuthPolicy) (string, error) {
+	svcId, err := c.findSvcId(ctx, k8sPolicy)
+	if err != nil {
+		return "", err
+	}
```
zijun726911, line 156: Do we need to `DisableSvcIAMAuth`  here by updating the latticeService AuthType to "NONE"? (I am not 100% sure on lattice side how this work)


(Same thing for `DisableSnIAMAuth`)

mikhail-aws: Make sense

mikhail-aws:
> Could you describe some manually tests you did for the PR?

@zijun726911 updated description with manual tests

zijun726911 (approved):
LGTM

PR #456, opened 2023-10-27
add iam auth policy conflict handling

mikhail-aws:
Note:
add IAM auth policy conflicts handling

Tested manually by creating multiple policies for same target ref.
Log example:
```
2023-10-27T13:19:32.780-0700	ERROR	controller.iam-auth-policy	controllers/iamauthpolicy_controller.go:52	conflict with other policies for same TargetRef, policy: test-iam-auth-policy, conflicted with: [test-iam-auth-policy2]
```

Policy Status updated:
```
Status:
  Conditions:
    Last Transition Time:  2023-10-27T20:13:30Z
    Message:
    Observed Generation:   1
    Reason:                Conflicted
    Status:                False
```

### DIFF
--- a/controllers/iamauthpolicy_controller.go
+++ b/controllers/iamauthpolicy_controller.go
@@ -2,6 +2,7 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	anv1alpha1 "github.com/aws/aws-application-networking-k8s/pkg/apis/applicationnetworking/v1alpha1"
@@ -12,9 +13,12 @@ import (
 	"github.com/aws/aws-application-networking-k8s/pkg/utils"
 	"github.com/aws/aws-application-networking-k8s/pkg/utils/gwlog"
 
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	gwv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 )
 
 type IAMAuthPolicyController struct {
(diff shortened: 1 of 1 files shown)


controllers/iamauthpolicy_controller.go
```diff
@@ -68,8 +78,8 @@ func (c *IAMAuthPolicyController) Reconcile(ctx context.Context, req ctrl.Reques
 
 	latticeResourceId, err := reconcileFunc(ctx, k8sPolicy)
 	if err != nil {
-		if services.IsNotFoundError(err) {
-			c.log.Infof("reconcile error, retry in 30sec: %s", err)
+		if !isDelete && services.IsNotFoundError(err) {
```
zijun726911, line 81: What this condition used for `!isDelete`?

controllers/iamauthpolicy_controller.go
```diff
@@ -43,6 +47,12 @@ func (c *IAMAuthPolicyController) Reconcile(ctx context.Context, req ctrl.Reques
 	}
 	c.log.Infow("reconcile", "req", req, "targetRef", k8sPolicy.Spec.TargetRef)
 
+	err = c.handleConflicts(ctx, k8sPolicy)
+	if err != nil {
+		c.log.Error(err)
+		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
```
zijun726911, line 53: for the Conflicted policy, should it just be "reconciled" with PolicyReasonConflicted status, instead of keep looping?

zijun726911 (approved):

controllers/iamauthpolicy_controller.go
mikhail-aws: Thanks for pointing again, logic is wrong in the code. Intention was:
If it's policy delete request and SN or Svc not found, dont throw error, just remove finalizer and we done.

controllers/iamauthpolicy_controller.go
mikhail-aws: with loop it should converge by itself when conflict is resolved

controllers/iamauthpolicy_controller.go
mikhail-aws: updated condition

mikhail-aws: we should start doing unit tests in controllers :D

zijun726911: Why I think old code is correct... `!isDelete && services.IsNotFoundError(err)`

mikhail-aws: if it's policy delete and svc not found ignore, otherwise throw -> (isDelete && notFound) -> ignore, !(isDelete && notFound) -> throw

PR #466, opened 2023-11-01
Add and use identification tags for Lattice service

solmonk:
**What type of PR is this?**

feature

* Adds identification tags for lattice service - name, namespace, routeType
* Even if the service name matches, stop reconciliation when these information does not match
* Utilize ownership concept, stop reconciliation when the service is "clearly" not owned by controller
* Backfill - if service does not have new tags, consider it is owned and add necessary tags
  * This might be controversial, but I don't see a big downside doing this.

**Automation added to e2e**:

```release-note

```


pkg/deploy/lattice/service_manager.go
```diff
@@ -124,7 +125,62 @@ func svcStatusFromCreateSvcResp(resp *CreateSvcResp) ServiceInfo {
 	return svcInfo
 }
 
+func (m *defaultServiceManager) checkAndUpdateOwnership(ctx context.Context, tags services.Tags, svc *Service, arn string) error {
```
mikhail-aws, line 128: I think ownership should be more abstract. Probably in cloud/lattice interface. I would think of following methods that should work together for different use-cases.

IsNotManaged(tags) bool (not managed at all)
IsManaged (tags) bool (managed by current controller)
tryOwn(arn) error (best effort of trying to acquire ownership)
handleOwnership(arn) error (current version of checkAndUpdateOwnership)

mikhail-aws: Also in long run we would need to have ownership check on every write operation. To make it robust there should be a cache with known ownership. So I see it that ownership should stand aside/above of any of resource. If each resource will have own ownership pattern, then it should be interface.

solmonk: Will think of way to abstract this in the next rev, thanks

controllers/route_controller.go
```diff
@@ -337,6 +337,17 @@ func (r *routeReconciler) reconcileUpsert(ctx context.Context, req ctrl.Request,
 	}
 
 	if _, err := r.buildAndDeployModel(ctx, route); err != nil {
+		if services.IsConflictError(err) {
```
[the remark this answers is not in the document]
solmonk: fixed missing status update

pkg/model/lattice/service.go
```diff
@@ -27,6 +25,29 @@ type ServiceStatus struct {
 	Dns string `json:"dns"`
 }
 
+type ServiceTagFields struct {
+	RouteName      string
+	RouteNamespace string
+	RouteType      core.RouteType
+}
+
+func ServiceTagFieldsFromTags(tags map[string]*string) ServiceTagFields {
```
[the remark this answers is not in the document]
solmonk: fixed

pkg/model/lattice/service.go
```diff
@@ -27,6 +25,29 @@ type ServiceStatus struct {
 	Dns string `json:"dns"`
 }
 
+type ServiceTagFields struct {
+	RouteName      string
+	RouteNamespace string
+	RouteType      core.RouteType
+}
```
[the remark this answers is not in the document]
solmonk: Trying to group fields that makes a resource identification. It makes things easier to compare. Mostly inspired by target group identification work

controllers/route_controller.go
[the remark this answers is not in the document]
mikhail-aws: I hope we dont override it later ;)

pkg/aws/cloud.go
```diff
@@ -107,22 +111,58 @@ func (c *defaultCloud) DefaultTagsMergedWith(tags services.Tags) services.Tags {
...
+	return c.isOwner(managedBy), nil
+}
+
+func (c *defaultCloud) CheckAndAcquireOwnershipFromTags(ctx context.Context, arn string, tags services.Tags) (bool, error) {
+	// For resources that need backwards compatibility - not having managedBy is considered as owned by controller.
+	managedBy := c.getManagedByFromTags(tags)
+	if managedBy == "" {
+		err := c.acquireOwnership(ctx, arn)
+		if err != nil {
+			return false, nil
```
mikhail-aws, line 149: return false, err ?

pkg/deploy/lattice/service_manager.go
```diff
@@ -124,7 +124,49 @@ func svcStatusFromCreateSvcResp(resp *CreateSvcResp) ServiceInfo {
...
+			fmt.Sprintf("Found existing resource with conflicting service name: %s", *svcSum.Arn))
+	}
+	return nil
+}
+
 func (m *defaultServiceManager) updateServiceAndAssociations(ctx context.Context, svc *Service, svcSum *SvcSummary) (ServiceInfo, error) {
+	err := m.checkAndUpdateTags(ctx, svc, svcSum)
+	if err != nil {
+		return ServiceInfo{}, err
+	}
```
mikhail-aws, line 168: It's probably ok for near future to imply that ownership and tags are related. And operate on them together to avoid redundant getTags calls. But conceptually ownership should be abstracted. For example this function tells that we do something with Tags, but underneath we manage ownership.

mikhail-aws (approved):
looks good, please take a look on missing error return

Issue #439, opened 2023-10-16
Get rid of side-effect coding style, trying to use stateless functions


zijun726911:
We once have this discusion topic in this thread:
https://github.com/aws/aws-application-networking-k8s/pull/430#discussion_r1355585456

mikhail-aws:
can you please update description with details, code examples, and your proposal

mikhail-aws:
I think we should rid of ModelBuilders and Synthesizers. I agree it should be functional. Model should only depend on k8sSpec. Once model is ready we call one or more managers to deploy.

I would like to have Reconcile function descriptive enough to understand full Reconcile flow. 

Pesudocode:

```go
func (c *Controller) Reconcile(req Req) {
  
  // spec
  mySpec := &MySpec{}
  c.client.Get(req, mySpec)

  // model 
  myModel := NewMyModel(mySpec)
  
  // deploy
  c.svcMgr.Create(myModel.Svc)
  c.tgMgr.Create(myModel.TargetGroup)

}
```

Of course in reality it's more complex, but Reconcile should have higher level functions that describe flow. One of the things that stands out to me is when we need to call lattice for Model building. I think we should not go that way.

I would like reconcile to glue parts together, we always build model from spec only, and for deployment if we need to call lattice multiple times make these successive calls in the Reconcile and pass data forward.

mikhail-aws:
We can try to write a guide. Also we need an exemplar controller for the reference. Good picks can be recent CRDs: accesslog, iamauth

PR #471, opened 2023-11-02
Fixed deletion of invalid Access Log Policies

xWink:
**What type of PR is this?**

bug

N/A

Currently, when an Access Log Policy (ALP) has its spec changed such that it becomes invalid, it cannot be deleted because the controller validation short circuits to setting the ALP status to Invalid and doesn't remove finalizers nor delete the underlying Access Log Subscription (ALS).

This change removes the short circuit from the deletion logic so that ALPs with invalid specs and their underlying ALS can be deleted.

1. Create valid ALP
   - ALP status should be Accepted
3. Update ALP spec to have invalid Group (e.g. `foo`) or no `destinationArn` in `targetRef`
   - ALP status should be Invalid
5. Delete ALP
   - **OLD:** ALP deletion hangs because finalizers are not removed
   - **NEW:** ALP deletion succeeds and underlying ALS is deleted

Reran all Access Log Policy e2e tests
```
Ran 13 of 51 Specs in 738.077 seconds
SUCCESS! -- 13 Passed | 0 Failed | 0 Pending | 38 Skipped
--- PASS: TestIntegration (739.34s)
PASS
ok      github.com/aws/aws-application-networking-k8s/test/suites/integration   740.098s
```

**Automation added to e2e**:

N/A, covered with unit tests

No

No

No, this is part of the Access Log Policy feature release

```release-note

```

### DIFF
--- a/pkg/gateway/model_build_access_log_subscription.go
+++ b/pkg/gateway/model_build_access_log_subscription.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/aws/aws-sdk-go/aws"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	anv1alpha1 "github.com/aws/aws-application-networking-k8s/pkg/apis/applicationnetworking/v1alpha1"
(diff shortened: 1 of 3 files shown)


pkg/gateway/model_build_access_log_subscription.go
```diff
@@ -57,26 +58,29 @@ type accessLogSubscriptionModelBuildTask struct {
...
-
-	var eventType = core.CreateEvent
-	if t.accessLogPolicy.DeletionTimestamp != nil {
-		eventType = core.DeleteEvent
-	} else if _, ok := t.accessLogPolicy.Annotations[anv1alpha1.AccessLogSubscriptionAnnotationKey]; ok {
-		eventType = core.UpdateEvent
+		if eventType != core.DeleteEvent {
+			return fmt.Errorf("access log policy's destinationArn cannot be nil")
+		}
+		destinationArn = aws.String("")
```
mikhail-aws, line 83: I dont have good suggestion yet, but it bothers me that we do conditional code branching on every step - reconciler, model, synthezier. Every time we do something that checks if it's delete/update/create, makes it harder to reason about how it actually works.

I'm thinking that create/update/delete branching should happen at Reconcile function. The rest of the code should not have these. It might create code duplicates, but those can be written into smaller functions. Right now I have to look at every stage and assemble little piece of what is delete logic and build some mental map of it.

It's not a blocker, but I believe we come back to this code soon.

mikhail-aws (approved):

pkg/gateway/model_build_access_log_subscription.go
xWink: Agreed, I think this is going to be related to #439 and #457

PR #472, opened 2023-11-02
Enabled Dependabot for monthly minor version updates

xWink:
Enabling Dependabot for monthly minor version updates on all dependencies except Gateway API

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    target-branch: "main"
+    versioning-strategy: "lockfile-only"
+    ignore:
+      - dependency-name: "gateway-api"
+    allow:
+      - dependency-type: "direct"
+        update-types: ["semver:minor"]


zijun726911 (approved):
Do you know how current Dependabot is triggered?
for example:
https://github.com/aws/aws-application-networking-k8s/pull/432

mikhail-aws:
where this config comes from?

mikhail-aws:
also dependency security scan is already enabled - https://github.com/aws/aws-application-networking-k8s/security

xWink:
Security scan and automated updates are two distinct features of Dependabot. One is meant to resolve security vulnerabilities when alerted, and the other is meant to keep dependencies up to date.

The goal here is just to keep dependencies up to date with minor version updates. We can change this to track major versions if we want to be more aggressive with updates.

#432 looks like it was triggered by a security-related event
```
You can disable automated security fix PRs for this repo from the [Security Alerts page](https://github.com/aws/aws-application-networking-k8s/network/alerts).
```

xWink:
> where this config comes from?

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

mikhail-aws:
Can you explain why this set of configurations. Also which part of the link above you refer, it's relatively large, there are many different examples.

xWink:
> Can you explain why this set of configurations. Also which part of the link above you refer, it's relatively large, there are many different examples.

Per previous discussion, it was stated that we would benefit from maintaining regularly scheduled automatic dependency updates. The configuration isn't a copy from any particular place, just intended to keep PRs of dependency updates coming so we don't forget about updating them. Security is one concern, but there are also QoL improvements that can come from these updates. It would be a nice-to-have to be notified of what changes are available in our dependencies.

Issue #467, opened 2023-11-01
Change AuthType to None Before Deleting IAMAuthPolicy


xWink:
In our IAMAuthPolicy controller, when [deleting a policy](https://github.com/aws/aws-application-networking-k8s/blob/main/controllers/iamauthpolicy_controller.go#L117), we delete the policy and then change the AuthType to None on the parent resource. This order should be reversed.

mikhail-aws:
why?

PR #478, opened 2023-11-07
Add iam auth policy validation and handle targetRef changes

mikhail-aws:
Note:

Add spec validations
- Spec is invalid
- TargetRef does not exists in k8s
- Multiple targetRefs point to same resource

Add resource cleanup - delete previous policy - when spec is changed.
All these use-cases tested manually.

Code changes:
- Move branching between ServiceNetwork and Service from Controller to Manager.
- Remove unnecessary k8s updates. Right now call only once for Status (on upsert) and Spec updates.
- Added some unit tests around controller

Leftovers:
- Remove manager's test code due to refactoring, will add later. 
- When policy is in conflict and conflict is resolved need to trigger update for other policies, otherwise policy stays in conflict until other unrelated triggers happen


controllers/errors.go
```diff
@@ -0,0 +1,10 @@
+package controllers
+
+import "errors"
+
+var (
+	TargetGroupNameErr = errors.New("wrong group name")
```
xWink, line 6: Error message is generic among all groups, so why is the error variable specific to target groups?

controllers/errors.go
```diff
@@ -0,0 +1,10 @@
...
+	TargetKindErr      = errors.New("target kind error")
+	TargetRefNotExists = errors.New("targetRef does not exists")
```
xWink, line 8: nit: `targetRef does not exist`

controllers/iamauthpolicy_controller.go
```diff
@@ -27,17 +29,26 @@ import (
 	gwv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 )
 
+const (
+	IAMAuthPolicyAnnotation      = "iam-auth-policy"
+	IAMAuthPolicyAnnotationResId = k8s.AnnotationPrefix + IAMAuthPolicyAnnotation + "-resource-id"
+	IAMAuthPolicyAnnotationType  = k8s.AnnotationPrefix + IAMAuthPolicyAnnotation + "-resource-type"
+	IAMAuthPolicyFinalizer       = k8s.AnnotationPrefix + IAMAuthPolicyAnnotation
```
xWink, line 36: Is this the typical finalizer format for CRDs? We follow a different pattern in other controllers, like `service.k8s.aws/resources`

xWink: From what I've read, finalizers are typically in the format of `example.com/finalizer`. This appears to be something we should change across all our resources, and have one finalizer for any resource type (which makes sense, since we have no reason to put the resource Kind in the finalizer for that kind, the purpose of the finalizer name is to avoid collisions with other controllers).

This is a non-blocker for the current PR, I'll make an issue to address it.

controllers/iamauthpolicy_controller.go
```diff
@@ -52,201 +63,174 @@ func RegisterIAMAuthPolicyController(log gwlog.Logger, mgr ctrl.Manager, cloud p
...
-	if k8sPolicy.DeletionTimestamp.IsZero() {
-		if !controllerutil.ContainsFinalizer(k8sPolicy, authPolicyFinalizer) {
-			controllerutil.AddFinalizer(k8sPolicy, authPolicyFinalizer)
-		}
-	} else {
-		if controllerutil.ContainsFinalizer(k8sPolicy, authPolicyFinalizer) {
-			controllerutil.RemoveFinalizer(k8sPolicy, authPolicyFinalizer)
+func (c *IAMAuthPolicyController) reconcileDelete(ctx context.Context, k8sPolicy *anv1alpha1.IAMAuthPolicy) (ctrl.Result, error) {
+	err := c.validateSpec(ctx, k8sPolicy)
+	if err == nil {
```
xWink, line 114: What is err != nil?

xWink (approved):
Minor comments, can be addressed later. Liking the new utility functions, will be nice QoL improvements for other reconcilers.

controllers/iamauthpolicy_controller.go
mikhail-aws: Does not really matter, as long as each controller has own finalizer there wont be issues.

controllers/errors.go
mikhail-aws: will change names

controllers/iamauthpolicy_controller.go
mikhail-aws: ignore validation checks on deletion

pkg/deploy/lattice/iamauthpolicy_manager.go
```diff
@@ -10,39 +10,127 @@ import (
...
+	return model.IAMAuthPolicyStatus{ResourceId: resourceId}, nil
+}
+
+func (m *IAMAuthPolicyManager) putSvc(ctx context.Context, policy model.IAMAuthPolicy) (model.IAMAuthPolicyStatus, error) {
+	svc, err := m.cloud.Lattice().FindService(ctx, policy.Name)
+	if err != nil {
+		return model.IAMAuthPolicyStatus{}, err
+	}
+	resourceId := *svc.Id
+	err = m.enableSvcIAMAuth(ctx, resourceId)
```
zijun726911, line 54: If I understand it correctly from our previous standup meeting, for this issue: https://github.com/aws/aws-application-networking-k8s/issues/467

When create the AuthPolicy, the order should be: (not correct in current PR)
1. PutAuthPolicy
2.  Change AuthType to AWS_IAM

When create the AuthPolicy, the order should be: (correct in current PR)
1. Change AuthType to NONE
2. DeleteAuthPolicy

Do you need to change creation handling order?

mikhail-aws: to put policy we first enable iam and then put policy, for delete we disable iam and remove policy

zijun726911: @mohakkohli Could you help to double confirm?

PR #480, opened 2023-11-07
Detach Gateway from ServiceNetwork

solmonk:
**What type of PR is this?**

Cleanup

* Fixes ServiceNetwork ownership, no longer automatically creates SN based on Gateway.
* VpcAssociationPolicy is still supported, I have created a separate reconciler for it
* VPC Association is only created when policy is defined
* VpcAssociationPolicy is properly managed with ownership, does not touch if it is owned by other controller

E2E tests now require test-gateway service network associated with VPC to run.

**Automation added to e2e**:

```release-note

```


pkg/aws/cloud.go
```diff
@@ -140,6 +141,22 @@ func (c *defaultCloud) IsArnManaged(ctx context.Context, arn string) (bool, erro
...
+	}
+	if managedBy == "" {
+		err := c.acquireOwnership(ctx, arn)
+		if err != nil {
+			return false, err
+		}
+		return true, nil
+	}
+	return c.isOwner(managedBy), nil
+}
```
mikhail-aws, line 158: I would change both ownership functions. Omit errors for brevity.

```go
func (c *defaultCloud) TryOwn(ctx context.Context, arn string) (bool, error) {
  tags := c.getTags(arn)
  return c.TryOwnFromTags(ctx, arn, tags)
}

func (c *defaultCloud) TryOwnFromTags(ctx context.Context, arn string, tags Tags) (bool, error) {
  managedBy := c.getManagedByFromTags(tags)
	if managedBy == "" {
		err := c.acquireOwnership(ctx, arn)
		if err != nil {
			return false, err
		}
		return true, nil
	}
	return c.isOwner(managedBy), nil
}
```

pkg/deploy/lattice/service_network_manager.go
```diff
@@ -43,6 +40,89 @@ type defaultServiceNetworkManager struct {
...
+	}
+}
+
+func (m *defaultServiceNetworkManager) DeleteVpcAssociation(ctx context.Context, snName string) error {
+	sn, err := m.cloud.Lattice().FindServiceNetwork(ctx, snName, "")
+	if err != nil {
+		return err
+	}
+
+	associated, snva, _, err := m.isServiceNetworkAlreadyAssociatedWithVPC(ctx, *sn.SvcNetwork.Id)
```
mikhail-aws, line 98: isServiceNetworkAlreadyAssociatedWithVPC deserve a change, 4 return values...
I believe we it can return (&SnVpcAssociation, error). nil means nothing found

pkg/gateway/model_build_targetgroup.go
```diff
@@ -137,15 +138,17 @@ func (t *svcExportTargetGroupModelBuildTask) buildTargetGroup(ctx context.Contex
...
 	if err != nil {
 		return nil, err
 	}
 
 	protocol := "HTTP"
 	protocolVersion := vpclattice.TargetGroupProtocolVersionHttp1
 	var healthCheckConfig *vpclattice.HealthCheckConfig
-	if tgp != nil {
+	if len(tgps) > 0 {
+		// TODO: TGP conflicts should be handled correctly w/ status update, for now just picking up one
```
mikhail-aws, line 150: what does it mean 

> TGP conflicts should be handled correctly w/ status update

pkg/k8s/policyhelper/policy_test.go
```diff
@@ -16,7 +16,7 @@ import (
 	"github.com/aws/aws-application-networking-k8s/pkg/model/core"
 )
 
-func Test_getAttachedPolicy(t *testing.T) {
+func Test_GetAttachedPolicies(t *testing.T) {
 	type args struct {
```
zijun726911, line 20: Could you extend the `Test_GetAttachedPolicies` to make it cover the code path that one obj have multiple different same kind policies? 

IIRC, in our CRDs for example: targetgrouppolicy_types.go vpcassociationpolicy_types.go

The `GetItems()` has a bug, this one is wrong:
```
func (pl *VpcAssociationPolicyList) GetItems() []core.Policy {
	items := make([]core.Policy, len(pl.Items))
	for i, item := range pl.Items {
		items[i] = &item
	}
	return items
}
```

it should change to this one:
```
func (pl *VpcAssociationPolicyList) GetItems() []core.Policy {
	items := make([]core.Policy, len(pl.Items))
	for i := range pl.Items {
		items[i] = &pl.Items[i]
	}
	return items
}

```

I think you need to do that change in this PR as well?

if you make `Test_GetAttachedPolicies` covers one obj having mutiple Policies, it could detect that bug.

pkg/deploy/lattice/service_network_manager.go
solmonk: I'm too afraid of touching existing code, so I made a new (more concise IMO) function that does the job. I will clean this up once I release PR for default SN.

controllers/vpcassociationpolicy_controller.go
```diff
@@ -0,0 +1,167 @@
...
+	kind := k8sPolicy.Spec.TargetRef.Kind
+	switch kind {
+	case "Gateway":
+		if isDelete || !isAssociation {
+			err = c.delete(ctx, k8sPolicy)
+		} else {
+			err = c.upsert(ctx, k8sPolicy)
+		}
+	default:
+		err = c.updatePolicyCondition(ctx, k8sPolicy, gwv1alpha2.PolicyReasonTargetNotFound)
```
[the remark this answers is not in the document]
mikhail-aws: "invalid target resource" is not specific, for accesslog and auth policy we use gwv1alpha2.PolicyReasonInvalid, for wrong Kind. And not found only when targetRef is literally not found in k8s api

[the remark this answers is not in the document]
mikhail-aws: NotFound sounds like it's not Policy's issue, but something does not exists. But in fact issue in policy itself.
If I would see NotFound error and my resource exists, I'll find it confusing.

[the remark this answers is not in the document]
solmonk: alright I will stick to what we have in common.

mikhail-aws (approved):

PR #481, opened 2023-11-07
change iam auth policy creation order put policy, enable IAM

mikhail-aws:
Note:
If we enable IAM without any policy attached it will reject all traffic until policy applied. 
To avoid temporary traffic rejection we first apply policy and then enable IAM auth.


zijun726911 (approved):
LGTM

Issue #455, opened 2023-10-26
Documentation: Ingress Security Group in deployment documentation doesn't properly use the prefix list


niax:
On the [deployment documentation](https://github.com/aws/aws-application-networking-k8s/blob/a2d5b973b3516f11784d1a764830e8de2eadf4a7/docs/deploy.md?plain=1#L19-L22) loads the prefix list and then just uses the first prefix. This will be a problem in the future when there's more than one prefix, or if the prefixes change.

Security Groups can reference prefix lists by ID, and dynamically update when the entries in them change:

```
$ aws ec2 authorize-security-group-ingress --group-id $SG_ID --ip-permissions "PrefixListIds=[{PrefixListId=${PREFIX_LIST_ID}}],IpProtocol=-1"
```

Should do the same as the current command, and also apply when the prefix lists change.

graehren:
Targeting this fix for upcoming [milestone release](https://github.com/aws/aws-application-networking-k8s/milestone/4)

PR #484, opened 2023-11-08
Docs: Update Docs to reference IPv4 and IPv6 Prefix lists IDs directly when configuring Security Groups

graehren:
**What type of PR is this?**
documentation

This change to the documentation references the prefix list ID directly when configuring Security Groups so they can be dynamically updated when the prefix list entries in them change.

**Which issue does this PR fix**: #455 

If the underlying IP range changes, traffic will be impacted to security groups that are not referencing the prefix list ID on the security group.

Run a fresh controller deployment using the documentation

No

Yes

```release-note
This change to the documentation references the prefix list ID directly when configuring Security Groups so they can be dynamically updated when the prefix list entries in them change.
```

### DIFF
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -38,12 +38,11 @@ Create an EKS cluster and allow Lattice traffic into cluster.
 ```bash
 eksctl create cluster --name $CLUSTER_NAME --region $AWS_REGION
 
-/ update security group to allow lattice traffic
-PREFIX_LIST_ID=$(aws ec2 describe-managed-prefix-lists --query "PrefixLists[?PrefixListName=="\'com.amazonaws.$AWS_REGION.vpc-lattice\'"].PrefixListId" | jq -r '.[]')
-MANAGED_PREFIX=$(aws ec2 get-managed-prefix-list-entries --prefix-list-id $PREFIX_LIST_ID --output json  | jq -r '.Entries[0].Cidr')
 CLUSTER_SG=$(aws eks describe-cluster --name $CLUSTER_NAME --output json| jq -r '.cluster.resourcesVpcConfig.clusterSecurityGroupId')
-aws ec2 authorize-security-group-ingress --group-id $CLUSTER_SG --cidr $MANAGED_PREFIX --protocol -1
-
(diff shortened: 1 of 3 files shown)


docs/developer.md
```diff
@@ -38,12 +38,11 @@ Create an EKS cluster and allow Lattice traffic into cluster.
...
-PREFIX_LIST_ID=$(aws ec2 describe-managed-prefix-lists --query "PrefixLists[?PrefixListName=="\'com.amazonaws.$AWS_REGION.vpc-lattice\'"].PrefixListId" | jq -r '.[]')
-MANAGED_PREFIX=$(aws ec2 get-managed-prefix-list-entries --prefix-list-id $PREFIX_LIST_ID --output json  | jq -r '.Entries[0].Cidr')
 CLUSTER_SG=$(aws eks describe-cluster --name $CLUSTER_NAME --output json| jq -r '.cluster.resourcesVpcConfig.clusterSecurityGroupId')
-aws ec2 authorize-security-group-ingress --group-id $CLUSTER_SG --cidr $MANAGED_PREFIX --protocol -1
-
+PREFIX_LIST_ID=$(aws ec2 describe-managed-prefix-lists --query "PrefixLists[?PrefixListName=="\'com.amazonaws.$AWS_REGION.vpc-lattice\'"].PrefixListId" | jq -r '.[]')
+aws ec2 authorize-security-group-ingress --group-id $CLUSTER_SG --ip-permissions "PrefixListIds=[{PrefixListId=${PREFIX_LIST_ID}}],IpProtocol=-1"
+PREFIX_LIST_ID_IPV6=$(aws ec2 describe-managed-prefix-lists --query "PrefixLists[?PrefixListName=="\'com.amazonaws.$AWS_REGION.ipv6.vpc-lattice\'"].PrefixListId" | jq -r '.[]')
+aws ec2 authorize-security-group-ingress --group-id $CLUSTER_SG --ip-permissions "PrefixListIds=[{PrefixListId=${PREFIX_LIST_ID_IPV6}}],IpProtocol=-1"
```
mikhail-aws, line 45: I think we should move these commands into shell script in our repo and document something like:

```
curl --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/aws/aws-application-networking-k8s/main/scripts/setup.sh | sh
```

golangci.yaml
```diff
@@ -0,0 +1,36 @@
+run:
```
mikhail-aws, line 1: thank you

mikhail-aws (approved):

PR #485, opened 2023-11-08
Bugfix: Remove deleting 5 min old and older TG logic

zijun726911:
Remove deleting 5 min and older target group logic in the `calculateTargetGroupsToDelete()`

This code is originally used as a kind of backward compatible change for the Lattice Target Group naming breaking change(Target Group Ownership change), however, this code cause a bug when doing the performance test: 

When the controller is trying to synthesize lots of lattice services and target groups, in the middle of reconciliation, it has one time that many target groups are created but don't have lattice service's route reference it. This code logic could wrongly recognize  that this target group need to be deleted, but actually should not.

### DIFF
--- a/pkg/deploy/lattice/target_group_synthesizer.go
+++ b/pkg/deploy/lattice/target_group_synthesizer.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"time"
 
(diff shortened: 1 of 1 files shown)


solmonk (approved):

erikfuller (approved):
Thanks for the PR. This should be fine to remove. If there end up being any future TG leaks we can address them then.

PR #490, opened 2023-11-10
Remove service_controller.go

zijun726911:


controllers/service_controller.go
```diff
@@ -1,109 +0,0 @@
...
-	// today, target registration is handled through the route controller
-	// if that proves not to be fast enough, we can look to do quicker
-	// target registration with the service controller
-
-	svc := &corev1.Service{}
-	if err := r.client.Get(ctx, req.NamespacedName, svc); err != nil {
-		return client.IgnoreNotFound(err)
-	}
-	if !svc.DeletionTimestamp.IsZero() {
-		r.finalizerManager.RemoveFinalizers(ctx, svc, serviceFinalizer)
```
solmonk, line 100: I think we still need this line, otherwise users cannot delete the service at all.

zijun726911: Should we do a breaking change to get rid of some unnecessary code before GA...

solmonk: This finalizer has been added to all services in a cluster that have controller installed. We should almost forever try to delete this. Removing the reconciler does not even solve the problem of controller uninstallation and only making it worse.

zijun726911: Ok, should we keep to just remove `"service.ki8s.aws/resources"` logic? Or remove all Finalizers, remove all finalizers could potentially have unexpected side effect?

solmonk: we should not touch other finalizers besides `service.ki8s.aws/resources`

solmonk: That being said, if we are worried about such performance a more effective move will be removing pod reconciler instead.

solmonk (approved):

PR #495, opened 2023-11-10
Optimize target group operations

solmonk:
**What type of PR is this?**

enhancement

* Add caching to ListTagsForResource calls, we may expand this to other calls as well
* Optimize list operation of target group
  * fetch tags at once using GetResources() api
  * Do not make unnecessary get() calls, as tgSummary is sufficient in most cases
* Add ProtocolVersion on TG tags, this saves multiple GetTargetGroup() calls for TG cleanup.
* Optimize find operation of target group
  * Utilizes GetResources() api to filter out target group

**Automation added to e2e**:

```release-note

```


pkg/aws/services/tagging.go
```diff
@@ -0,0 +1,97 @@
...
+	maxArnsPerGetResourcesApi = 100
+)
+
+type Tags = map[string]*string
+
+type Tagging interface {
+	taggingapiiface.ResourceGroupsTaggingAPIAPI
+
+	// Receives a list of arns and returns arn-to-tags map.
+	GetTagsFromArns(ctx context.Context, arns []*string) (map[string]Tags, error)
```
mikhail-aws, line 28: From -> For?

mikhail-aws: arns []string, I dont think we need reference type here

mikhail-aws: Also you can consider a strong typed version arn.ARN, much easier to use once constructed.

https://pkg.go.dev/github.com/aws/aws-sdk-go@v1.47.9/aws/arn#Parse

```go
tgArns := utils.SliceMap(resp, func(tg *vpclattice.TargetGroupSummary) arn.ARN { 
  tgArn , _ := arn.Parse(*tg.Arn)
  return tgArn
})
```

pkg/aws/services/tagging.go
```diff
@@ -0,0 +1,97 @@
...
+
+	// Finds one resource that matches the given set of tags.
+	FindResourceWithTags(ctx context.Context, resourceType string, tags Tags) (*string, error)
```
mikhail-aws, line 31: With -> By, also should return arn.ARN, not clear what *string means here

mikhail-aws: resourceType string, can be strong typed, we have limited options better to keep it enum-like

pkg/aws/services/tagging.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	resp, err := t.GetResourcesWithContext(ctx, input)
+	if err != nil {
+		return nil, err
+	}
+	if len(resp.ResourceTagMappingList) == 0 {
+		return nil, NewNotFoundError("tag", "matching criteria")
+	}
+	// assume one result
+	return resp.ResourceTagMappingList[0].ResourceARN, nil
```
mikhail-aws, line 72: I think assumption about how many results to expect should be on caller of this function, not here. So returning a slice of results would be better. Also empty slice as return an easy check for not found.

pkg/aws/services/vpclattice.go
```diff
@@ -118,9 +120,11 @@ func NewDefaultLattice(sess *session.Session, region string) *defaultLattice {
 		endpoint = latticeEndpoint
 	}
 
-	latticeSess = vpclattice.New(sess, aws.NewConfig().WithRegion(region).WithEndpoint(endpoint).WithMaxRetries(20))
+	latticeSess := vpclattice.New(sess, aws.NewConfig().WithRegion(region).WithEndpoint(endpoint).WithMaxRetries(20))
```
mikhail-aws, line 123: what do you think about 20 retries and it's impact on throttling/performance?

pkg/aws/services/vpclattice.go
```diff
@@ -224,6 +228,36 @@ func (d *defaultLattice) ListTargetGroupsAsList(ctx context.Context, input *vpcl
...
+	}
+	return out, nil
+}
+
+func tagCacheKey(arn string) string {
+	return "tag-" + arn
+}
+
+func (d *defaultLattice) TagResourceWithContext(ctx context.Context, input *vpclattice.TagResourceInput, option ...request.Option) (*vpclattice.TagResourceOutput, error) {
+	if d.cache != nil {
```
mikhail-aws, line 254: why cache should be nil?

pkg/deploy/lattice/target_group_manager.go
```diff
@@ -238,98 +237,98 @@ func (s *defaultTargetGroupManager) Delete(ctx context.Context, modelTg *model.T
...
-		if status == vpclattice.TargetGroupStatusCreateFailed {
-			continue
+	latticeTg, err := s.cloud.Lattice().GetTargetGroupWithContext(ctx, &vpclattice.GetTargetGroupInput{
+		TargetGroupIdentifier: arn,
+	})
+	if err != nil {
+		if services.IsLatticeAPINotFoundErr(err) {
+			return nil, nil
 		}
+		return nil, err
```
mikhail-aws, line 294: nit:

```go
return nil, services.IgnoreNotFound(err)
```

pkg/deploy/lattice/target_group_manager.go
```diff
@@ -238,98 +237,98 @@ func (s *defaultTargetGroupManager) Delete(ctx context.Context, modelTg *model.T
...
-			case vpclattice.TargetGroupStatusDeleteFailed:
-				return latticeTg, nil
-			case vpclattice.TargetGroupStatusDeleteInProgress:
-				return nil, errors.New(LATTICE_RETRY)
-			}
+	// we ignore create failed status, so may as well check for it first
+	status := aws.StringValue(latticeTg.Status)
+	if status == vpclattice.TargetGroupStatusCreateFailed {
+		return nil, nil
+	}
```
mikhail-aws, line 301: I dont follow why "findTargetGroup" make decisions about status and retry errors. It's a surprising behavior with "find" function. 

May be name is misleading? But it seems too much of responsibility for this function: searching, discarding "some" results, matching, retrying.

mikhail-aws (changes requested):

pkg/aws/services/tagging.go
solmonk: I would prefer string more, ARN is purely an identifier here and processing it adds no value.

pkg/aws/services/tagging.go
solmonk: makes sense

pkg/aws/services/tagging.go
solmonk: makes sense

pkg/aws/services/vpclattice.go
solmonk: I have never observed backoffs getting more than 2~3 on manual tests, so it is hard for me to think that it is affecting anything. we can definitely revisit this later though.

pkg/aws/services/vpclattice.go
solmonk: I think when using cache we should almost always have a way to switch off for testing purpose, etc.

pkg/deploy/lattice/target_group_manager.go
solmonk: makes sense

pkg/deploy/lattice/target_group_manager.go
solmonk: Discarding some status should make total sense. It is always easier to think that `CREATE_FAILED` target group doesn't exist, and `DELETE_FAILED` target group just exists.

On the other hand, `LATTICE_RETRY` is currently like a magical fallback for us to "hold off" decision whenever we would like. It makes sense if you argue we are abusing it too much, but I find it aligning with other controller behaviors so didn't decide to touch yet.

pkg/aws/services/tagging.go
```diff
@@ -0,0 +1,98 @@
...
+//go:generate mockgen -destination tagging_mocks.go -package services github.com/aws/aws-application-networking-k8s/pkg/aws/services Tagging
+
+type ResourceType string
+
+const (
+	resourceTypePrefix = "vpc-lattice:"
+
+	ResourceTypeTargetGroup ResourceType = resourceTypePrefix + "targetgroup"
+
+	maxArnsPerGetResourcesApi = 100
```
mikhail-aws, line 21: nit: can you add link to the limit please

pkg/aws/services/vpclattice.go
mikhail-aws: in this case should be interface with no-op cache

pkg/deploy/lattice/target_group_manager.go
```diff
@@ -124,19 +123,19 @@ func (s *defaultTargetGroupManager) update(ctx context.Context, targetGroup *mod
...
 	_, err := s.cloud.Lattice().UpdateTargetGroupWithContext(ctx, &vpclattice.UpdateTargetGroupInput{
 		HealthCheck:           healthCheckConfig,
-		TargetGroupIdentifier: latticeTgSummary.Id,
+		TargetGroupIdentifier: latticeTg.Id,
 	})
 
 	if err != nil {
 		return model.TargetGroupStatus{},
-			fmt.Errorf("Failed UpdateTargetGroup %s due to %s", aws.StringValue(latticeTgSummary.Id), err)
+			fmt.Errorf("Failed UpdateTargetGroup %s due to %s", aws.StringValue(latticeTg.Id), err)
```
mikhail-aws, line 131: nit: %w for err

pkg/deploy/lattice/target_group_manager.go
```diff
@@ -238,98 +237,99 @@ func (s *defaultTargetGroupManager) Delete(ctx context.Context, modelTg *model.T
...
+	arns, err := s.cloud.Tagging().FindResourcesByTags(ctx, services.ResourceTypeTargetGroup,
+		model.TagsFromTGTagFields(modelTargetGroup.Spec.TargetGroupTagFields))
 	if err != nil {
 		return nil, err
 	}
+	if len(arns) == 0 {
+		return nil, nil
+	}
+	// Can get only one arn through the above search criteria.
+	arn := arns[0]
```
mikhail-aws, line 289: lets throw or log error when assumption does not work

pkg/deploy/lattice/target_group_manager.go
```diff
@@ -238,98 +237,99 @@ func (s *defaultTargetGroupManager) Delete(ctx context.Context, modelTg *model.T
...
+		Type:          latticeTg.Type,
+		VpcIdentifier: latticeTg.Config.VpcIdentifier,
+	}, nil) // we already know that tags match
+	if err != nil {
+		return nil, err
+	}
+	if match {
+		switch status {
+		case vpclattice.TargetGroupStatusCreateInProgress, vpclattice.TargetGroupStatusDeleteInProgress:
+			return nil, errors.New(LATTICE_RETRY)
```
mikhail-aws, line 319: ```retrun fmt.Errorf("%w, delete in progress id=%d", runtime.RetryErr)```

mikhail-aws (approved):

Issue #493, opened 2023-11-10
Invalid BackendRef should not block route creation/update


erikfuller:
According to the spec, invalid backendRefs should return a 500 rather than block route creation or updates. Lattice does not currently support fixed responses outside of 404, so we will return that for now.

Current behaviour (as originally highlighted by @zijun726911 offline) and steps to reproduce:

```bash
kubectl apply -f examples/my-hotel-gateway.yaml
kubectl apply -f examples/inventory-route.yaml
```

By creating a route with no backing service, we will see errors like:

```
failed at BuildModel: {“logger”:“controller.route”,“caller”:“controllers/route_controller.go:282”,“msg”:“buildAndDeployModel, Failed build model for salmo-5-ok4zh8ytnj due to failed to build rules due to buildTargetGroupSpec err error finding backend service e2e-test/bugmu-2-tfbytur05o due to Service \“bugmu-2-tfbytur05o\” not found”}
```

This will never succeed until the service is created.

erikfuller:
Fix for this in the attached PR. Follow up is #494

Issue #463, opened 2023-10-30
Consider to remove generic stack


zijun726911:
[stack.go](https://github.com/aws/aws-application-networking-k8s/blob/5be20c35c4090e5ac4e5c3370eff97e9167a9e0c/pkg/model/core/stack.go#L14) as well as resource_graph.go, typological_traversal.go from core package are not quite fit our controller use case. We should consider to remove generic stack and use exact models of lattice resources instead.

Related discussion: https://github.com/aws/aws-application-networking-k8s/pull/457#discussion_r1376684924

PR #497, opened 2023-11-11
Return 404 for invalid backendRefs

erikfuller:
**What type of PR is this?**
bug fix

#493 

First, this PR aligns the implementation more with the gateway spec for invalid backendRefs. With this change, invalid backendRefs will either be omitted when another backendRef exists that is valid or will configure the rule to return a 404. 

Second, when building listener model objects, we will no longer create model objects for deleted listeners. As I was making this change, I simplified the listener logic that was tracking deletes. Since we never actually delete listeners and they are deleted when the service is deleted, I've removed the unneeded code.

***Manual Testing***
The following shows adding a route with an invalid backendRef, then adding the service (so the backendRef is valid), then deleting the service again (returns to an invalid backendRef).

First, the backendRef results in a fixedResponse 404
```
└─[$] <git:(tg-svc-fix*)> [laptop] kubectl apply -f examples/my-hotel-gateway.yaml
└─[$] <git:(tg-svc-fix*)> [laptop] kubectl apply -f examples/inventory-route.yaml
└─[$] <git:(tg-svc-fix*)> [laptop] aws vpc-lattice get-rule --service-identifier svc-X --listener-identifier listener-X --rule-identifier rule-X
{
    "action": {
        "fixedResponse": {
            "statusCode": 404
        }
    },
    "arn": "arn:aws:vpc-lattice:...",
...
    "priority": 1
}

```

Next, we get a real target group after creating the service
```
└─[$] <git:(tg-svc-fix*)> [laptop] kubectl apply -f examples/inventory-ver1.yaml
└─[$] <git:(tg-svc-fix*)> [laptop] aws vpc-lattice get-rule --service-identifier svc-X --listener-identifier listener-X --rule-identifier rule-X
{
    "action": {
        "forward": {
            "targetGroups": [
                {
                    "targetGroupIdentifier": "tg-X",
                    "weight": 10
                }
            ]
        }
    },
    "arn": "arn:aws:vpc-lattice:...",
...
}
```

Then we delete the service and get a fixed response again
```
└─[$] <git:(tg-svc-fix*)> [laptop] kubectl delete -f examples/inventory-ver1.yaml
└─[$] <git:(tg-svc-fix*)> [laptop] aws vpc-lattice get-rule --service-identifier svc-X --listener-identifier listener-X --rule-identifier rule-X
{
    "action": {
        "fixedResponse": {
            "statusCode": 404
        }
    },
    "arn": "arn:aws:vpc-lattice:...",
    ...
}
```

**Automation added to e2e**:
Changed ```deregister_targets_test.go``` to ```delete_target_groups_test.go```. What we actually want are target groups for a service to be deleted when that service is deleted. So I've updated the test to reflect this.

I also saw some issues in validating targets for target groups in ```https_listener_weighted_rule_with_service_export_import_test.go``` so fixed the logic that was checking the wrong lengths in ```framework.go```

No

n/a

```release-note
Reconciliation is now more tolerant of invalid backendRefs. Deleted services or incorrect references will no longer block route reconciliation.
```

### DIFF
--- a/pkg/gateway/model_build_targetgroup.go
+++ b/pkg/gateway/model_build_targetgroup.go
@@ -24,6 +24,15 @@ import (
 	model "github.com/aws/aws-application-networking-k8s/pkg/model/lattice"
 )
 
+type InvalidBackendRefError struct {
+	BackendRef core.BackendRef
+	Reason     string
+}
+
+func (e *InvalidBackendRefError) Error() string {
+	return e.Reason
+}
+
 type SvcExportTargetGroupModelBuilder interface {
 	// used during standard model build
 	Build(ctx context.Context, svcExport *anv1alpha1.ServiceExport) (core.Stack, error)
@@ -247,7 +256,7 @@ func (t *backendRefTargetGroupModelBuildTask) buildTargetGroup(ctx context.Conte
 	}
 	t.log.Debugf("Added target group for backendRef %s to the stack %s", t.backendRef.Name(), stackTG.ID())
 
-	stackTG.IsDeleted = !t.route.DeletionTimestamp().IsZero()
+	stackTG.IsDeleted = !t.route.DeletionTimestamp().IsZero() // should always be false
 	if !stackTG.IsDeleted {
 		t.buildTargets(ctx, stackTG.ID())
 	}
@@ -260,16 +269,10 @@ func (t *backendRefTargetGroupModelBuildTask) buildTargets(ctx context.Context, 
 		t.log.Debugf("Service import does not manage targets, returning")
 		return nil
 	}
-
 	backendRefNsName := getBackendRefNsName(t.route, t.backendRef)
 	svc := &corev1.Service{}
 	if err := t.client.Get(ctx, backendRefNsName, svc); err != nil {
-		if apierrors.IsNotFound(err) && !t.route.DeletionTimestamp().IsZero() {
-			t.log.Infof("Ignoring not found error for service %s on deleted route %s",
-				t.backendRef.Name(), t.route.Name())
-		} else {
-			return fmt.Errorf("error finding backend service %s due to %s", backendRefNsName, err)
-		}
+		return fmt.Errorf("error finding backend service %s due to %s", backendRefNsName, err)
 	}
 
 	targetsBuilder := NewTargetsBuilder(t.log, t.client, t.stack)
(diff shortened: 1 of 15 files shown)


pkg/gateway/model_build_targetgroup_test.go
```diff
@@ -418,47 +418,6 @@ func Test_TGModelByHTTPRouteBuild(t *testing.T) {
 			wantErrIsNil:        false,
 			wantIPv6TargetGroup: false,
 		},
-		{
-			name: "Create LatticeService where backend serviceimport does NOT exist",
```
erikfuller, line 422: This test was actually succeeding previously because it was supposed to error, but it wasn't erroring for the reason the test intended. The ```Build()``` method checks and errors immediately for ServiceImport types. We no longer create model TG objects for service imports, but rather create a ```SvcImportTargetGroup``` directly against the rule (see ```rule.go```).

I removed this test because it's misleading as written and was making it harder to write tests against the ```InvalidBackendRef``` cases.

test/pkg/test/framework.go
```diff
@@ -467,9 +467,9 @@ func (env *Framework) GetTargets(ctx context.Context, targetGroup *vpclattice.Ta
 					*target.Status == vpclattice.TargetStatusHealthy)
 		})
 
-		g.Expect(retrievedTargets).Should(HaveLen(len(targetIps)))
+		g.Expect(targetIps).Should(HaveLen(len(podIps)))
```
erikfuller, line 470: This condition was evaluating as true when there were zero targets, which is logically incorrect. What we actually want here is for the pod ips to all be present in the list of targets.

pkg/gateway/model_build_targetgroup.go
```diff
@@ -247,7 +256,7 @@ func (t *backendRefTargetGroupModelBuildTask) buildTargetGroup(ctx context.Conte
 	}
 	t.log.Debugf("Added target group for backendRef %s to the stack %s", t.backendRef.Name(), stackTG.ID())
 
-	stackTG.IsDeleted = !t.route.DeletionTimestamp().IsZero()
+	stackTG.IsDeleted = !t.route.DeletionTimestamp().IsZero() // should always be false
```
zijun726911, line 259: QQ: now, which part of controller code is responsible to do "delete the k8s Service trigger deleting Unused Target Groups"? I think should be  `targetGroupSynthesizer.SynthesizeUnusedDelete()` ?

Seems to understand changes in one function also need to have heavy knowledge on how other functions works... a little bit hard to trace code..

zijun726911 (approved):

pkg/gateway/model_build_targetgroup.go
erikfuller: Yes, unfortunately at the moment you kind of need to know what's happening elsewhere to know if what's being done locally is OK. The unit tests help explain some of that and enforce the invariants. Part of the issue with keeping this logic entirely separate was we were doing a lot of redundant/throw-away work. For example, we would create models for listeners that were going to be deleted then do nothing with them. I think this is a good example of optimisation being at odds with properties like decoupling/modularization. Having said that, I think we can look at making this better. My intuition is that as part of #463 we'll be able to have better performance AND readability.

To answer the specific question, any unused target groups or target groups which reference invalid services will be deleted as part of the ```SynthesizeUnusedDelete()```. We can and should run this cleanup on a schedule rather than only when a change is made.

PR #507, opened 2023-11-16
Added Auto-Generated API Reference

xWink:
## Changes
- Added `docs/api-reference.md`
    - Contains API Spec for all CRDs
- Changed `docs/reference` to `docs/api-types` to be more consistent with Gateway API directory format (and this naming makes more sense)
- Updated mkdocs to use the new API reference
- Added new `api-reference` target in Makefile to do the auto-generation
- Added condition in Makefile to only export env vars when not using the `help` target
    - The export commands use aws CLI commands which add a substantial slowdown to their execution and cause `make help` to take a long time
    - Additionally, `make help` changing the env is a weird behaviour that people likely don't expect


zijun726911 (approved):
LGTM

Issue #508, opened 2023-11-16
Consolidate TargetRef validation logic for Policy CRD's


mikhail-aws:
We have multiple CRD that use [Policy Attachment](https://gateway-api.sigs.k8s.io/geps/gep-713/): AccessLog, AuthPolicy, TargetGroupPolicy, VpcAssociationPolicy. 

All of above preform similar validation on TargetRef:
- GroupName and Kind
- Namespace
- Conflict with other Policies
- TargetRef is not found

Currently each policy has own "flavor" of doing it, but essentially it's same logic with different types.

PR #509, opened 2023-11-16
add target group policy controller and status updates

### LINKED ISSUE #440: TargetGroupPolicy and VpcAssociationPolicy CRD status Update
Currently controller behavior is: if multiple TargetGroupPolicies (tgp), VpcAssociationPolicies(vap) apply to the same k8s obj, only the first policy will take effect and the controller will ignore the rest of policies(will print a error log for it). If that happen, we should use a better way to represent this error by updating policy status: for example, If there are two tgps attached to one svc(serviceExport), one policy is set to `accepted` status and the other one is set to `conflicted` status.

(Other policies CRDs or even other controller managed k8s objs may also need proper status update )

mikhail-aws:
Note:
Add TargetGroupPolicy controller. Controller validate targetRef and update status. TargetGroup builder applies only Accepted and TargetRefNotFound policies. TargetRefNotFound is safe to apply, it means policy was created before TG.
I have to update CRD because Status wasnt there fully. Users need to re-apply CRD after this change.

Tested manually:
- Create policy before TG, after TG was created checked policy applied
- Create multiple conflicting policies and checked none of them applied
- Created policy and TG, then created conflicting policy. Checked second conflicting policy didnt override first one

We should refactor our PolicyAttachment code, right now every Policy CRD implements same validation and status update logic. Need to create general purpose PolicyAttachment handler for validation and status update. #508 

close #440


pkg/k8s/policyhelper/policy.go
```diff
@@ -20,7 +22,11 @@ type policyInfo struct {
 	kind       gwv1beta1.Kind
 }
```
zijun726911, line 23: I feel like that would be a better policyInfo naming:
```
type policyInfo struct {
	policyList core.PolicyList
	targetRefGroup      gwv1beta1.Group
	targetRefKind       gwv1beta1.Kind
}
```

pkg/k8s/policyhelper/policy.go
```diff
@@ -20,7 +22,11 @@ type policyInfo struct {
...
 
-func GetAttachedPolicies[T core.Policy](ctx context.Context, k8sClient client.Client, searchTargetRef types.NamespacedName, policy T) ([]T, error) {
+type PolicyConditionFilter struct {
+	Reasons []gwv1alpha2.PolicyConditionReason
+}
+
+func GetAttachedPoliciesConditionFilter[T core.Policy](ctx context.Context, k8sClient client.Client, searchTargetRef types.NamespacedName, policy T, filter PolicyConditionFilter) ([]T, error) {
```
zijun726911, line 29: Seems this function actually do:  `GetAttachedPoliciesFilterByAcceptedConditionReason` , would be good if we can add comments or change function name to make people more easily know this function's behaviour.

solmonk:
> I have to update CRD because Status wasnt there fully. Users need to re-apply CRD after this change.

So when the controller upgrades without CRD upgrade, it will disregard all TGPs and turn HTTP2 TGs into HTTP1 TGs, it seems. It is a bit worrying behavior to me actually. Should we add a fallback, like consider TGPs with no status as valid? Or, should we increment TGP version so that only old TGPs with no status are valid?

zijun726911:
> > I have to update CRD because Status wasnt there fully. Users need to re-apply CRD after this change.
> 
> So when the controller upgrades without CRD upgrade, it will disregard all TGPs and turn HTTP2 TGs into HTTP1 TGs, it seems. It is a bit worrying behavior to me actually. Should we add a fallback, like consider TGPs with no status as valid? Or, should we increment TGP version so that only old TGPs with no status are valid?

Just provide upgrade instructions and do this breaking change before v1.0.0 looks fine for me

pkg/k8s/policyhelper/policy.go
mikhail-aws: removed and refactored for conflict resolution logic

solmonk (approved):
The approach overall LGTM.

PR #510, opened 2023-11-16
Fix target group leaking and e2etest

solmonk:
**What type of PR is this?**

bug

* Use strongly consistent list call on ListTargetGroups, fixes target group leaking issue (which seems to happen more often when we have a lot of existing target groups)
* Correctly handle when there are multiple target groups matching tags in findTargetGroup(). This can happen in the middle of TG protocol update, etc. Target groups only become unique after matching more fields including protocol. Changed related e2e test to make sure unused target groups are getting replaced by this logic.
* Fix wrong e2e test cases on delete_target_group_test, deleting a deployment still results in target deregistration as it does not delete service.
* Fix bug where e2e test will pick up existing other hosted zone and delete it
* Made a few tests ordered. We are running tests sequentially anyways, and even in parallel execution we will likely keep one file per worker.
* Made a few tweaks to be ready for parallel execution, it seems working with 2 procs but not sure if it is reliable.

`make e2e-test`

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/deploy/lattice/target_group_manager.go
+++ b/pkg/deploy/lattice/target_group_manager.go
@@ -248,10 +248,7 @@ type tgListOutput struct {
 func (s *defaultTargetGroupManager) List(ctx context.Context) ([]tgListOutput, error) {
 	lattice := s.cloud.Lattice()
 	var tgList []tgListOutput
-	targetGroupListInput := vpclattice.ListTargetGroupsInput{
-		VpcIdentifier:   aws.String(config.VpcID),
-		TargetGroupType: aws.String(vpclattice.TargetGroupTypeIp),
-	}
+	targetGroupListInput := vpclattice.ListTargetGroupsInput{}
 	resp, err := lattice.ListTargetGroupsAsList(ctx, &targetGroupListInput)
 	if err != nil {
 		return nil, err
@@ -259,7 +256,11 @@ func (s *defaultTargetGroupManager) List(ctx context.Context) ([]tgListOutput, e
 	if len(resp) == 0 {
 		return nil, nil
 	}
-	tgArns := utils.SliceMap(resp, func(tg *vpclattice.TargetGroupSummary) string {
+	validTgs := utils.SliceFilter(resp, func(tg *vpclattice.TargetGroupSummary) bool {
+		return aws.StringValue(tg.VpcIdentifier) == config.VpcID &&
+			aws.StringValue(tg.Type) == vpclattice.TargetGroupTypeIp
+	})
+	tgArns := utils.SliceMap(validTgs, func(tg *vpclattice.TargetGroupSummary) string {
 		return aws.StringValue(tg.Arn)
 	})
 	tgArnToTagsMap, err := s.cloud.Tagging().GetTagsForArns(ctx, tgArns)
@@ -267,7 +268,7 @@ func (s *defaultTargetGroupManager) List(ctx context.Context) ([]tgListOutput, e
 	if err != nil {
 		return nil, err
 	}
-	for _, tg := range resp {
+	for _, tg := range validTgs {
 		tgList = append(tgList, tgListOutput{
 			tgSummary: tg,
 			tags:      tgArnToTagsMap[*tg.Arn],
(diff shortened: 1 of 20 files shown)


test/suites/integration/byoc_test.go
```diff
@@ -292,7 +292,11 @@ func findHostedZone(client *route53.Route53, name string) (*route53.HostedZone,
 	if len(out.HostedZones) == 0 {
 		return nil, nil
 	}
-	return out.HostedZones[0], nil
+	hz := out.HostedZones[0]
+	if aws.StringValue(hz.Name) != name {
+		return nil, nil
+	}
```
mikhail-aws, line 298: why name should not be equal? list zones should apply dns name filter
`out, err := client.ListHostedZonesByName(&route53.ListHostedZonesByNameInput{DNSName: &name})`

mikhail-aws: > "Fix bug where e2e test will pick up existing other hosted zone and delete it"

There was intention to use existing hosted zone if test was terminated and zone leaked. Test re-run should pick-up previous zone and cleanup after. Otherwise need to manually delete them from route53.

solmonk: ListHostedZonesByName always returns all hosted zones in the account. DNSName field only guarantees the first element that is getting picked up, and you still get other hosted zones in the following elements.

pkg/deploy/lattice/target_group_manager.go
```diff
@@ -288,44 +289,43 @@ func (s *defaultTargetGroupManager) findTargetGroup(
...
-			return latticeTg, nil
+		// Check the immutable fields to ensure TG is valid
+		match, err := s.IsTargetGroupMatch(ctx, modelTargetGroup, &vpclattice.TargetGroupSummary{
+			Arn:           latticeTg.Arn,
+			Port:          latticeTg.Config.Port,
+			Protocol:      latticeTg.Config.Protocol,
+			IpAddressType: latticeTg.Config.IpAddressType,
+			Type:          latticeTg.Type,
+			VpcIdentifier: latticeTg.Config.VpcIdentifier,
+		}, nil) // we already know that tags match
```
zijun726911, line 318: Do we need to check `ProtocolVersion` in the `(s *defaultTargetGroupManager) findTargetGroup()`? I saw in the e2e test framework code we check it. i.e., is that possible in one moment, in the middle of update TargetGroup, it have 2 ACTIVE status tgs both match?

pkg/deploy/lattice/target_group_manager.go
```diff
@@ -288,44 +289,43 @@ func (s *defaultTargetGroupManager) findTargetGroup(
...
+		if err != nil {
+			return nil, err
+		}
+		if match {
+			switch status {
+			case vpclattice.TargetGroupStatusCreateInProgress, vpclattice.TargetGroupStatusDeleteInProgress:
+				return nil, errors.New(LATTICE_RETRY)
+			case vpclattice.TargetGroupStatusDeleteFailed, vpclattice.TargetGroupStatusActive:
```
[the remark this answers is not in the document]
solmonk: I don't think so, they do exist.

test/pkg/test/framework.go
```diff
@@ -153,46 +153,28 @@ func (env *Framework) ExpectToBeClean(ctx context.Context) {
 		retrievedServices, _ := env.LatticeClient.ListServicesAsList(ctx, &vpclattice.ListServicesInput{})
 		for _, service := range retrievedServices {
 			env.Log.Infof("Found service, checking if created by current EKS Cluster: %v", service)
-			retrievedTags, err := env.LatticeClient.ListTagsForResourceWithContext(ctx, &vpclattice.ListTagsForResourceInput{
-				ResourceArn: service.Arn,
-			})
+			managed, err := env.Cloud.IsArnManaged(ctx, *service.Arn)
 			if err == nil { // for err != nil, it is possible that this service own by other account, and it is shared to current account by RAM
```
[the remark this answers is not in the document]
solmonk: its failing to fetch tags so most of the time it will be other account, not other cluster within the same account. I can update the comment though.

test/pkg/test/framework.go
```diff
@@ -153,46 +153,28 @@ func (env *Framework) ExpectToBeClean(ctx context.Context) {
...
-				tagValue, ok := retrievedTags.Tags[model.K8SSourceTypeKey]
-				if ok && *tagValue == string(model.SourceTypeSvcExport) {
-					env.Log.Infof("TargetGroup: %s was created by k8s controller, by a ServiceExport", *tg.Id)
-					//This tg is created by k8s controller, by a ServiceExport,
-					//ServiceExport still have a known targetGroup leaking issue,
-					//so we temporarily skip to verify whether ServiceExport created TargetGroup is deleted or not
-					continue
-				}
+			managed, err := env.Cloud.IsArnManaged(ctx, *tg.Arn)
+			if err == nil { // for err != nil, it is possible that this service own by other account, and it is shared to current account by RAM
```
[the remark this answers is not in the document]
solmonk: missed that, will change

test/pkg/test/framework.go
```diff
@@ -153,46 +153,28 @@ func (env *Framework) ExpectToBeClean(ctx context.Context) {
...
-				env.Log.Infof("Found Tags for service %v tags: %v", *service.Name, retrievedTags)
-				value, ok := retrievedTags.Tags[model.K8SServiceOwnedByVPC]
-				if ok {
-					g.Expect(*value).To(Not(Equal(CurrentClusterVpcId)))
-				}
+				g.Expect(managed).To(BeFalse())
 			}
 		}
 
 		retrievedTargetGroups, _ := env.LatticeClient.ListTargetGroupsAsList(ctx, &vpclattice.ListTargetGroupsInput{})
```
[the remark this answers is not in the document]
solmonk: this makes sense

pkg/deploy/lattice/target_group_manager.go
zijun726911, line 318: Just wondering should we add a `Protocol`(or other tg fields) tags for TargetGroup?
So that to make sure one `FindResourcesByTags()` api could definitely exactly match one TG?

 (not in this PR)

solmonk: Not sure if it is necessary. Actually.. I wanted to remove as many tags as possible. I "had to" add protocolVersion in the tag because it is not available in TGSummary, so it needed extra Get() call to be figured out. Ideally VPC Lattice should just return protocolVersion in the summary instead.

zijun726911 (approved):
LGTM, are we safe to add the `--procs` param in the Makefile?

solmonk:
I have thought about that but I think it is at experimental stage yet. Will add that once I have more confidence

PR #514, opened 2023-11-17
HTTPRoute, Service, and IAMAuthPolicy Documentation Enhancements

xWink:
- Added HTTPRoute and Service API reference
- Removed API specifications from API reference docs
    - The API reference itself is automatically updated, while these docs are manually updated. We should avoid having the same thing documented in two different places by two different methods, as this creates a risk of drift
- Added examples to IAMAuthPolicy API reference
- General clarity improvements to existing docs


docs/api-types/service.md
```diff
@@ -0,0 +1,53 @@
...
+microservice. The set of Pods targeted by a Service is determined by a `selector`.
+
+### Service Key Features & Limitations:
+
+**Features**:
+
+- **Load Balancing**: Services offer load balancing, distributing network traffic across the Pods.
+- **Service Types**: Supports different types, such as ClusterIP (default), NodePort, LoadBalancer, and ExternalName.
+- **Stable IP Address**: Each Service has a stable IP address, even when the Pods it routes to change.
+- **DNS Naming**: Automatically assigns a DNS name to the Service for easy referencing.
```
solmonk, line 15: Service is technically not a part of our API (we don't own the implementation) so I don't think we want to highlight every k8s service feature/limitation. It is probably better to focus on how it integrates with our controller, etc.

For example,
* Controller don't support all service types especially ExternalName which is prohibited by GWAPI spec.
* we don't want DNS naming to be recognized as our key feature as it is one of ways to bypass controller auth policies.

xWink: Fair point, removed DNS naming and updated ExternalName limitation to indicate that it is not supported at all.

solmonk (approved):

Issue #469, opened 2023-11-01
Gateway API v1.0 Support


solmonk:
Gateway API v1.0 is officially GA. We need to figure out what is missing in our implementation, and action accordingly.
* [Blog](https://kubernetes.io/blog/2023/10/31/gateway-api-ga/)
* [Release Notes](https://github.com/kubernetes-sigs/gateway-api/releases/tag/v1.0.0)
* [Implementer's Guide](https://gateway-api.sigs.k8s.io/reference/implementers-guide/)

Important action items:
* Upgrade Gateway API dependency
* Update GatewayClass resource status - need to implement `SupportedVersion` condition.
* Update GatewayClass resource status - need to provide `supportedFeature` field.
* Revisit conformance test (requires finishing resource status update work). Based on conformance we should determine what to populate in `supportedFeature` field.
* Remove CRD installation from deployment / helm chart - This is an official guideline now. For our own CRDs, we might want to provide a separate installation instruction to each resource.

solmonk:
* Dependency upgrade
* CRD installation is kept for now, but new CRDs will not be included in the future release.
* GWClass resource status should be tracked as a bigger conformance milestone.

PR #515, opened 2023-11-17
Fix e2e tests and examples to be compatible w/ latest CRDs

solmonk:
**What type of PR is this?**

bug

Partially - #469 

Changed behavior for empty matches in rules. They had default value in the past but they don't right now.
https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.HTTPRouteRule
https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1alpha2.GRPCRouteRule

Fixed our examples to be compatible with latest GWAPI CRDs.
* HTTPRoute/GRPCRoute - All Service backendRefs require ports
* Gateway - HTTPS listeners must have TLS section
* Gateway - TLS section must include at least one cert when mode is `Terminate`
  * This is a tricky part, I had to add dummy fields on our examples. It looks ugly but probably okay.
* Gateway - Listeners must be unique on name/port/hostname

`make e2e-test` all pass

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/gateway/model_build_rule_test.go
+++ b/pkg/gateway/model_build_rule_test.go
@@ -153,6 +153,8 @@ func Test_RuleModelBuild(t *testing.T) {
 			expectedSpec: []model.RuleSpec{ // note priority is only calculated at synthesis b/c it requires access to existing rules
 				{
 					StackListenerId: "listener-id",
+					PathMatchPrefix: true,
+					PathMatchValue:  "/",
 					Action: model.RuleAction{
 						TargetGroups: []*model.RuleTargetGroup{
 							{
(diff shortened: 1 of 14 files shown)


pkg/gateway/model_build_rule_test.go
```diff
@@ -1453,6 +1464,8 @@ func Test_RuleModelBuild(t *testing.T) {
 			expectedSpec: []model.RuleSpec{
 				{
 					StackListenerId: "listener-id",
+					PathMatchPrefix: true,
+					PathMatchValue:  "/",
 					Action: model.RuleAction{
```
xWink, line 1469: Do we have any tests for empty GRPCRoute rule?

xWink (approved):
Thank you for fixing these issues!

pkg/gateway/model_build_rule_test.go
solmonk: We have a exact e2e test case that creates GRPCRouteRule with no match.

PR #524, opened 2023-11-17
fix crd autogen

mikhail-aws:
update auto generated CRD's documentation


xWink (approved):

xWink:
Will need to run the autogen for this to apply to the actual api reference doc

PR #526, opened 2023-11-18
Removed Accepted status changes due to VPC Lattice validations

xWink:
**What type of PR is this?**

Cleanup

N/A

- Removes changes to the Accepted status when a VPC Lattice validation fails
    - Accepted status is now set once the Kubernetes resource validations are complete
    - If VPC Lattice validation fails, an event and log are still emitted
    - This type of validation failure should result in changes to the Programmed status, if we add support for it in the future

```
[SynchronizedAfterSuite] PASSED [31.573 seconds]

Ran 12 of 57 Specs in 465.564 seconds
SUCCESS! -- 12 Passed | 0 Failed | 0 Pending | 45 Skipped
--- PASS: TestIntegration (465.57s)
PASS
ok      github.com/aws/aws-application-networking-k8s/test/suites/integration   466.361s
```

**Automation added to e2e**:

Removed checks on Accepted status changes due to VPC Lattice validation

No

No

Yes

```release-note
- Access Log Policy Accepted status is now set to True after the controller's validation succeeds, instead of after VPC Lattice validation succeeds
```

### DIFF
--- a/controllers/accesslogpolicy_controller.go
+++ b/controllers/accesslogpolicy_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
 
 	"golang.org/x/exp/slices"
 	corev1 "k8s.io/api/core/v1"
@@ -75,7 +76,7 @@ func RegisterAccessLogPolicyController(
 ) error {
 	mgrClient := mgr.GetClient()
 	scheme := mgr.GetScheme()
-	evtRec := mgr.GetEventRecorderFor("access-log-policy-controller")
+	evtRec := mgr.GetEventRecorderFor("access-log-policy")
 
 	modelBuilder := gateway.NewAccessLogSubscriptionModelBuilder(log, mgrClient)
 	stackDeployer := deploy.NewAccessLogSubscriptionStackDeployer(log, cloud, mgrClient)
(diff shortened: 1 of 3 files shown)


controllers/accesslogpolicy_controller.go
```diff
@@ -213,17 +239,12 @@ func (r *accessLogPolicyReconciler) reconcileUpsert(ctx context.Context, alp *an
...
 	r.eventRecorder.Event(alp, corev1.EventTypeNormal, k8s.ReconciledEvent, "Successfully reconciled")
 
 	return nil
 }
 
-func (r *accessLogPolicyReconciler) targetRefExists(ctx context.Context, alp *anv1alpha1.AccessLogPolicy) (bool, error) {
+func (r *accessLogPolicyReconciler) findTargetRef(ctx context.Context, alp *anv1alpha1.AccessLogPolicy) (client.Object, error) {
 	targetRefNamespacedName := types.NamespacedName{
 		Name:      string(alp.Spec.TargetRef.Name),
 		Namespace: k8s.NamespaceOrDefault(alp.Spec.TargetRef.Namespace),
```
mikhail-aws, line 250: namespace should be same as alp itself based on GEP, also empty namespace should not fallback to default

mikhail-aws (changes requested):
Please take a look at target-group-policy. There are common functions for policy as policy exists and conflict resolution. 

https://github.com/aws/aws-application-networking-k8s/blob/main/controllers/targetgrouppolicy_controller.go

https://github.com/aws/aws-application-networking-k8s/blob/main/pkg/k8s/policyhelper/policy.go

mikhail-aws:
I'm going to do refactoring on the policy crds, to consolidate common logic. It will impact current code, and very likely I will remove some of it. We can put this PR on hold or cancel.

PR #543, opened 2023-11-24
add gw-api-v1 to controller's scheme

mikhail-aws:
Add gw-api-v1 to controller's Scheme. No idea why it worked before.


mikhail-aws:
It seems I need to update other places that watches for v1 resources rather than beta. Will close this PR, it will be part of larger PR for policy refactoring.

PR #547, opened 2023-11-28
Add generic policy handler and refactor IAM Auth Polciy

mikhail-aws:
Note:
Add generic policy handler struct. The high-level design is to introduce generic helper struct with collection of different methods to work with Policy CRD's. This helper/handler supposed to be used in controllers that interact with Policy CRDs. 

Currently applied only to IAMAuthPolicy CRD, other CRDs will come after this PR.

Controller's code should explicitly call different, and relatively small methods of handler, so that it should be easy to tell what controller does looking at controller's code.

Code includes some arcane golang around generics. For example:

```go
func NewPolicyHandler[T, TL any, P policyPtr[T], PL policyListPtr[TL, P]](cfg PolicyHandlerConfig) *PolicyHandler[P] {
```

This 4 generic types methods are results of having 2 types of policy structs Policy and PolicyList in every CRD. Also I need to make distinction between Policy and ref type *Policy for some functionality related to k8s client calls - Get and List.

All generic complexity stays in Policy file but interface is still simple with single type and I believe understandable methods.
PolicyHandler creation is as simple as it can be, not valid config to call all methods, but it's enough for compiler to understand it's IAMAuthPolicy and all related methods will be statically checked.

```go

         type iap = anv1alpha1.IAMAuthPolicy
         type iapl = anv1alpha1.IAMAuthPolicyList

         phcfg PolicyHandlerConfig := PolicyHandlerConfig{}
         _ = NewPolicyHandler[iap, iapl](cfg: phcfg)

```

Testing:
- Unit tests and FOCUS="IAM" e2e-test. 
- Manual tests around setting correct statuses for IAM policies.


pkg/k8s/policyhelper/policy.go
```diff
@@ -119,3 +156,290 @@ func conflictResolutionSort[T core.Policy](policies []T) {
...
+	Log            gwlog.Logger
+	Client         k8sclient.Client
+	TargetRefKinds *GroupKindSet
+}
+
+// Creates policy handler for specific policy. T and TL are type and list-type for Policy (struct type, not reference).
+// P and PL are reference types and should derive from T and TL. P and PL do not require explicit declaration. For example:
+//
+//	ph := NewPolicyHandler[IAMAuthPolicy, IAMAuthPolicyList](cfg)
+func NewPolicyHandler[T, TL any, P policyPtr[T], PL policyListPtr[TL, P]](cfg PolicyHandlerConfig) *PolicyHandler[P] {
```
xWink, line 202: Per offline discussion, would be nice if we could infer the policy list type from the policy type

mikhail-aws: That was my initial intent, but couldn't make it work. Go type inference is not that flexible and I'm getting into cyclic dependency issue or type cannot derived and need to be explicitly declared (in this case it's even worse and clumsy)

We can do runtime resolution through type switches, but wont get compiler errors on that. My intent is to let compiler figure-out type miss matches.

We still need at some point explicitly state relationship between Single policy and Plural. Current approach is least amount of code with cost of passing ListType upfront.

mikhail-aws: But I'm open to suggestions, tried few angles here none worked:
- add method to Policy interface to produce list NewList() PolicyList[P] will introduce cyclic dependency

- tried to create a Factory struct that can create new Policy and PolicyList. Go cannot infer type.

```go
type PolicyFactory[P Policy, L PolicyList[P]] interface {
        NewPolicy() P
        NewList() L
}

func NewPolicyHandler[P Policy, L PolicyList[P], F PolicyFactory[P, L]](f F) *PolicyHandler[P] {
        return &PolicyHandler[P]{p: f.NewPolicy()}
}

func main() {
        f := &DummyPolicyFactory{}
        // h := NewPolicyHandler(f) // err: cannot infer P and L
        h := NewPolicyHandler[*DummyPolicy, *DummyPolicyList](f)
}
```

Go cannot infer type from methods of struct, these needs to be explicitly declared.

pkg/k8s/policyhelper/policy.go
```diff
@@ -119,3 +156,290 @@ func conflictResolutionSort[T core.Policy](policies []T) {
...
+func (h *PolicyHandler[P]) AddWatchers(b *builder.Builder, objs ...k8sclient.Object) {
+	h.log.Debugf("add watchers for types: %v", NewGroupKindSet(objs).Items())
+	for _, watchObj := range objs {
+		b.Watches(watchObj, handler.EnqueueRequestsFromMapFunc(h.watchMapFn))
+	}
+}
+
+func (h *PolicyHandler[P]) watchMapFn(ctx context.Context, obj k8sclient.Object) []reconcile.Request {
+	out := []reconcile.Request{}
+	policies, err := h.client.List(ctx, obj.GetNamespace())
```
zijun726911, line 327: Change name from `h.client` to `h.policyClient` is actually more clear for me

pkg/k8s/policyhelper/kind.go
```diff
@@ -0,0 +1,50 @@
...
+	case GroupKind{gwv1beta1.GroupName, "Gateway"}:
+		return &gwv1beta1.Gateway{}
+	case GroupKind{gwv1beta1.GroupName, "HTTPRoute"}:
+		return &gwv1beta1.HTTPRoute{}
+	case GroupKind{gwv1alpha2.GroupName, "GRPCRoute"}:
+		return &gwv1alpha2.GRPCRoute{}
+	case GroupKind{corev1.GroupName, "Service"}:
+		return &corev1.Service{}
+	default:
+		panic("unsupported group kind")
```
zijun726911, line 48: Seems the call of this function is:
```
func (pc *k8sPolicyClient[T, U, P, PL]) TargetRefObj(ctx context.Context, p P) (k8sclient.Object, error) {
	tr := p.GetTargetRef()
	obj := GroupKindObj(TargetRefGroupKind(tr))
```

If the user create a policy with invalid TargetRef Kind, for example:

```
apiVersion: application-networking.k8s.aws/v1alpha1
kind: IAMAuthPolicy
metadata:
    name: test-iam-auth-policy
spec:
    targetRef:
        group: "gateway.networking.k8s.io"
        kind: Pod
        name: my-pod
    policy: |
        {
            .......
        }
```

Will the controller be panic? (I did check the code logic too deep)

pkg/controllers/iamauthpolicy_controller.go
```diff
@@ -109,10 +114,10 @@ func (c *IAMAuthPolicyController) Reconcile(ctx context.Context, req ctrl.Reques
...
 func (c *IAMAuthPolicyController) reconcileDelete(ctx context.Context, k8sPolicy *anv1alpha1.IAMAuthPolicy) (ctrl.Result, error) {
-	err := c.validateSpec(ctx, k8sPolicy)
+	err := c.ph.ValidateTargetRef(ctx, k8sPolicy)
 	if err == nil {
 		modelPolicy := model.NewIAMAuthPolicy(k8sPolicy)
-		_, err := c.policyMgr.Delete(ctx, modelPolicy)
+		_, err := c.pm.Delete(ctx, modelPolicy)
 		if err != nil {
 			return ctrl.Result{}, services.IgnoreNotFound(err)
 		}
```
zijun726911, line 123: For this function:
```
func (c *IAMAuthPolicyController) reconcileDelete(ctx context.Context, k8sPolicy *anv1alpha1.IAMAuthPolicy) (ctrl.Result, error) {
	err := c.ph.ValidateTargetRef(ctx, k8sPolicy)
	if err == nil {
		modelPolicy := model.NewIAMAuthPolicy(k8sPolicy)
		_, err := c.pm.Delete(ctx, modelPolicy)
		if err != nil {
			return ctrl.Result{}, services.IgnoreNotFound(err)
		}
	}
	err = c.handleLatticeResourceChange(ctx, k8sPolicy, model.IAMAuthPolicyStatus{})
	if err != nil {
		return ctrl.Result{}, err
	}
	c.removeFinalizer(k8sPolicy)
	return ctrl.Result{}, nil
}

```
Seem this function itself do a `c.pm.Delete()` and within ` c.handleLatticeResourceChange()`, it do a `c.pm.Delete()` again, is that expected?


Do you mind to write some code comments, or PR comments on how we handle deleting event for happyPathAuthPolicy, targetRefNotFoundPolicy and conflictedPolicy(i.e, didn't rank 1st after the `conflictResolutionSort` )?  This part logic is not really straightforward to me.


(I have similar question on `reconcileUpsert()`, seems we do `c.pm.Put()` first, then do a conditional `c.pm.Delete()`in `handleLatticeResourceChange()`, also expected?)

pkg/controllers/iamauthpolicy_controller.go
```diff
@@ -35,30 +29,41 @@ const (
...
-	controller := &IAMAuthPolicyController{
-		log:       log,
-		client:    mgr.GetClient(),
-		policyMgr: deploy.NewIAMAuthPolicyManager(cloud),
-		cloud:     cloud,
+	watchObj := []client.Object{&gwv1beta1.Gateway{}, &gwv1beta1.HTTPRoute{}, &gwv1alpha2.GRPCRoute{}}
+	phcfg := policy.PolicyHandlerConfig{
+		Log:            log,
+		Client:         mgr.GetClient(),
+		TargetRefKinds: policy.NewGroupKindSet(watchObj),
```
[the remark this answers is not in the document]
mikhail-aws: watchObj --> watchObjs yes, TargetRefKinds -> TargetRefGKSet not really

pkg/k8s/policyhelper/policy.go
```diff
@@ -119,3 +156,290 @@ func conflictResolutionSort[T core.Policy](policies []T) {
...
+	var t T
+	return &t
+}
+
+func (pc *k8sPolicyClient[T, U, P, PL]) List(ctx context.Context, namespace string) ([]P, error) {
+	l := pc.newList()
+	opts := &k8sclient.ListOptions{}
+	if namespace != "" { // policies suppose to be namespaced
+		opts.Namespace = namespace
+	}
```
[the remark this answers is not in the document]
mikhail-aws: probably yes

pkg/k8s/policyhelper/policy.go
xWink: I see, thanks for taking a look into it nonetheless!

pkg/k8s/policyhelper/kind.go
mikhail-aws: I'm still thinking about it. I'm not sure returning error here is good. Is there something like "RawObject" that implements client.Object interface?

pkg/k8s/policyhelper/kind.go
```diff
@@ -0,0 +1,50 @@
...
+	gwv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+	gwv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+)
+
+type GroupKind struct {
+	Group string
+	Kind  string
+}
+
+func ObjGroupKind(obj client.Object) GroupKind {
```
xWink, line 15: nit: Could this be renamed to ObjToGroupKind (same for other funcs below), given that it's a mapper method?

pkg/k8s/policyhelper/policy.go
mikhail-aws: There is single client instance in policy and it's PolicyClient. There is no ambiguity in here

pkg/k8s/policyhelper/policy.go
```diff
@@ -2,27 +2,64 @@ package policyhelper
...
 	gwv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 
 	anv1alpha1 "github.com/aws/aws-application-networking-k8s/pkg/apis/applicationnetworking/v1alpha1"
 	"github.com/aws/aws-application-networking-k8s/pkg/model/core"
+	"github.com/aws/aws-application-networking-k8s/pkg/utils"
+	"github.com/aws/aws-application-networking-k8s/pkg/utils/gwlog"
 )
 
+var (
+	ErrGroupKind         = errors.New("group/kind error")
```
xWink, line 29: "group/kind error" is not indicative of the issue. Is the problem that the user is attempting to create a resource with the wrong group/kind? If so, perhaps "invalid group/kind" is more descriptive.

pkg/k8s/policyhelper/policy.go
```diff
@@ -2,27 +2,64 @@ package policyhelper
...
+	ConditionTypeAccepted = gwv1alpha2.PolicyConditionAccepted
+
+	ReasonAccepted       = gwv1alpha2.PolicyReasonAccepted
+	ReasonInvalid        = gwv1alpha2.PolicyReasonInvalid
+	ReasonTargetNotFound = gwv1alpha2.PolicyReasonTargetNotFound
+	ReasonConflicted     = gwv1alpha2.PolicyReasonConflicted
+
+	// Non-GEP
+
+	ReasonUnknown = ConditionReason("Unknown")
```
xWink, line 52: When should we use this?

mikhail-aws: When reconciler cannot reconcile for other reasons than validation. Bugs, errors in calling k8s api

pkg/k8s/policyhelper/policy.go
mikhail-aws: this error for error type checking, actual descriptive error below in the code

https://github.com/aws/aws-application-networking-k8s/blob/af8774450d0a8c698c0d82945e81081a66c8b4e1/pkg/k8s/policyhelper/policy.go#L371-L374

xWink (changes requested):
Some small comments. Also would like to see unit tests on the new policy helper functions.

pkg/k8s/policyhelper/policy.go
```diff
@@ -119,3 +156,290 @@ func conflictResolutionSort[T core.Policy](policies []T) {
...
+			return fmt.Errorf("%w, target=%s/%s",
+				ErrTargetRefNotFound, policy.GetNamespace(), tr.Name)
+		}
+		return err
+	}
+	resolvedPolicy, err := h.ObjResolvedPolicy(ctx, targetRefObj)
+	if err != nil {
+		return err
+	}
+	if resolvedPolicy.GetName() != policy.GetName() {
```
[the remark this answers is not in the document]
mikhail-aws: gep 713 is explicit about conflict resolution https://gateway-api.sigs.k8s.io/geps/gep-713/#conflict-resolution

here I sort all policies for a given object using conflict-resolution rules and take first one from list which is "non-conflicting" or "resolved" policy

solmonk:
I feel like this PR kinda gives a vibe of a sophisticated TypeScript code doing type circuses.

Doing a recap of why I thought this was hard to read:

* need to understand relationship between T, U, P, and PL. so P is a pointer of T and PL is a pointer of U
* need to understand relationship between type constraints in different structs. policyListPtr's T is policyClient's TL
* the name policyPtr sounds like a interface ptr but actually a struct ptr pointing to underlying policy implementation
* U doesn't relate to any initial of an object, should be TL or something else?
* T is an `any` type which doesn't say much by itself, and can't actually constraint anything (I think this is most critical)

If we manage to reduce it to just P and PL I think it is more consistent and readable (I mean, except for a few lines of reflection magic)

And I'm okay with having both single and list types on handler generics. I think it has benefits, as you mentioned.

mikhail-aws:
@solmonk I agree it would be nicer to have only 2 types: policy and list. I dont mind to follow up after this pr, currently this behavior is isolated to local k8sClient struct and not exposed outside, it should be an easy change after.

pkg/k8s/policyhelper/kind.go
mikhail-aws: I'm thinking to return client.Object and bool, similar to type coercion

pkg/controllers/iamauthpolicy_controller.go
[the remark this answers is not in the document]
mikhail-aws: even apimachinery refer to GroupVersionKind as Kinds https://github.com/kubernetes/apimachinery/blob/fa98d6eaedb4caccd69fc07d90bbb6a1e551f00f/pkg/runtime/scheme.go#L243

pkg/controllers/iamauthpolicy_controller.go
```diff
@@ -126,102 +131,29 @@ func (c *IAMAuthPolicyController) reconcileDelete(ctx context.Context, k8sPolicy
...
 	if err != nil {
-		return err
-	}
-	if !refExists {
-		return fmt.Errorf("%w: %s", TargetRefNotFound, tr.Name)
+		return reconcile.Result{}, services.IgnoreNotFound(err)
 	}
-	conflictingPolicies, err := c.findConflictingPolicies(ctx, k8sPolicy)
+	c.updateLatticeAnnotaion(k8sPolicy, statusPolicy.ResourceId, modelPolicy.Type)
+	err = c.handleLatticeResourceChange(ctx, k8sPolicy, statusPolicy)
```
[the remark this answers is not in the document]
mikhail-aws: actually it follows our naming convention - https://github.com/aws/aws-application-networking-k8s/blob/main/docs/contributing/developer-cheat-sheet.md#service-service-or-service

I extended that to status too, because it's same as modelPolicy

mikhail-aws:
@xWink addressed minor comments, but not unit tests. I will do follow-up on this PR with unit tests, currently it's being manually tested and ran full e2e test suite.

xWink (approved):

xWink:
LGTM. Thanks for putting this together!

solmonk (approved):

zijun726911 (approved):

Issue #401, opened 2023-09-19
Add missing API references to site


graehren:
Improve documentation for API references (similar to TGPolicy and gRPC)

solmonk:
ServiceExport/ServiceImport is missing at the moment

PR #551, opened 2023-11-29
Add service export/import documentations

solmonk:
**What type of PR is this?**

documentation

#401 

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/docs/api-types/grpc-route.md
+++ b/docs/api-types/grpc-route.md
@@ -27,7 +27,7 @@ This allows you to define and manage the routing of gRPC traffic within your Kub
 
 ### Annotations
 
-- `application-networking.k8s.aws/lattice-assigned-domain-name`
+- `application-networking.k8s.aws/lattice-assigned-domain-name`  
   Represents a VPC Lattice generated domain name for the resource. This annotation will automatically set
   when a `GRPCRoute` is programmed and ready.
 
--- a/docs/api-types/http-route.md
+++ b/docs/api-types/http-route.md
@@ -26,7 +26,7 @@ This allows you to define and manage the routing of HTTP and HTTPS traffic withi
 
 ### Annotations
 
-- `application-networking.k8s.aws/lattice-assigned-domain-name`
+- `application-networking.k8s.aws/lattice-assigned-domain-name`  
   Represents a VPC Lattice generated domain name for the resource. This annotation will automatically set
   when a `HTTPRoute` is programmed and ready.
 
--- a/docs/api-types/service-export.md
+++ b/docs/api-types/service-export.md
@@ -2,14 +2,15 @@
 
 ## Introduction
 
-In AWS Gateway API Controller, `ServiceExport` is the way of enabling a Service for multi-cluster traffic setup.
-Once a Service is exported, other clusters can refer to it using [`ServiceImport`](service-import.md) resource.
+In AWS Gateway API Controller, `ServiceExport` enables a Service for multi-cluster traffic setup.
+Clusters can import the exported service with [`ServiceImport`](service-import.md) resource.
 
-ServiceExport is a term originated from Kubernetes [Multicluster Service APIs](https://multicluster.sigs.k8s.io/concepts/multicluster-services-api/);
-but AWS Gateway API Controller uses its own version of ServiceExport resource with a certain set of features.
+Internally, creating a ServiceExport creates a standalone VPC Lattice [target group](https://docs.aws.amazon.com/vpc-lattice/latest/ug/target-groups.html).
+Even without ServiceImports, creating ServiceExports can be useful in case you only need the target groups created;
+for example, using target groups in the VPC Lattice setup outside Kubernetes.
 
-Internally, creating a ServiceExport creates a VPC Lattice [target group](https://docs.aws.amazon.com/vpc-lattice/latest/ug/target-groups.html).
-You can take advantage of this by using the created target group in VPC Lattice setup outside Kubernetes.
+Note that ServiceExport is not the implementation of Kubernetes [Multicluster Service APIs](https://multicluster.sigs.k8s.io/concepts/multicluster-services-api/);
+instead AWS Gateway API Controller uses its own version of the resource for the purpose of Gateway API integration.
 
 For more multi-cluster use cases and detailed explanation, please refer to [our recommended multi-cluster architecture](../guides/multi-cluster.md).
 
--- a/docs/api-types/service-export.md
+++ b/docs/api-types/service-export.md
@@ -1,0 +1,38 @@
+# ServiceExport API Reference
+
+## Introduction
+
+In AWS Gateway API Controller, `ServiceExport` is the way of enabling a Service for multi-cluster traffic setup.
+Once a Service is exported, other clusters can refer to it using [`ServiceImport`](service-import.md) resource.
+
+ServiceExport is a term originated from Kubernetes [Multicluster Service APIs](https://multicluster.sigs.k8s.io/concepts/multicluster-services-api/);
+but AWS Gateway API Controller uses its own version of ServiceExport resource with a certain set of features.
+
+Internally, creating a ServiceExport creates a VPC Lattice [target group](https://docs.aws.amazon.com/vpc-lattice/latest/ug/target-groups.html).
+You can take advantage of this by using the created target group in VPC Lattice setup outside Kubernetes.
+
+For more multi-cluster use cases and detailed explanation, please refer to [our recommended multi-cluster architecture](../guides/multi-cluster.md).
+
+### Limitations
+* The exported Service can only be used in HTTPRoutes. GRPCRoute is currently not supported.
+* Limited to one ServiceExport per Service. If you need multiple exports representing each port,
+  you should create multiple Service-ServiceExport pairs.
+
+### Annotations
+
+* `application-networking.k8s.aws/port`  
+  Represents which port of the exported Service will be used.
+  When a comma-separated list of ports is provided, the traffic will be distributed to all ports in the list.
+
+## Example Configuration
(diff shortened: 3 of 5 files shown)


docs/api-types/service-export.md
```diff
@@ -0,0 +1,38 @@
+# ServiceExport API Reference
+
+## Introduction
+
+In AWS Gateway API Controller, `ServiceExport` is the way of enabling a Service for multi-cluster traffic setup.
```
mikhail-aws, line 5: > is the way of enabling a Service for multi-cluster traffic setup.

is the way of enabling = enables

docs/api-types/service-export.md
```diff
@@ -0,0 +1,38 @@
...
+Once a Service is exported, other clusters can refer to it using [`ServiceImport`](service-import.md) resource.
```
mikhail-aws, line 6: Try not to use passive voice. 

>Cluster can import exported Service with ServiceImport.

docs/api-types/service-export.md
```diff
@@ -0,0 +1,38 @@
...
+
+ServiceExport is a term originated from Kubernetes [Multicluster Service APIs](https://multicluster.sigs.k8s.io/concepts/multicluster-services-api/);
+but AWS Gateway API Controller uses its own version of ServiceExport resource with a certain set of features.
```
mikhail-aws, line 9: >AWS Gateway API Controller's ServiceExport is not the implementation of Kubernetess MCS specification. We might support MCS spec in future.

docs/api-types/service-export.md
```diff
@@ -0,0 +1,38 @@
...
+
+Internally, creating a ServiceExport creates a VPC Lattice [target group](https://docs.aws.amazon.com/vpc-lattice/latest/ug/target-groups.html).
+You can take advantage of this by using the created target group in VPC Lattice setup outside Kubernetes.
```
mikhail-aws, line 12: > You can take advantage

That applies to everything that controller creates in Lattice. Probably be more specific what kind of advantage or just remove this sentence.

docs/api-types/service-import.md
```diff
@@ -0,0 +1,54 @@
+# ServiceImport API Reference
+
+## Introduction
+
+`ServiceImport` is a resource referring to a Service outside the cluster, paired with [`ServiceExport`](service-export.md)
```
mikhail-aws, line 5: > paired with ServiceExport

Is it constrained to ServiceExport or any target group can be imported?

docs/api-types/service-export.md
```diff
@@ -0,0 +1,38 @@
...
+## Example Configuration
+
+The following yaml will create a ServiceExport for a Service named `service-1`:
+```yaml
+apiVersion: application-networking.k8s.aws/v1alpha1
+kind: ServiceExport
+metadata:
+  name: service-1
+  annotations:
+    application-networking.k8s.aws/port: "9200"
```
zijun726911, line 36: Can we show a list of comma-separated port in here, e.g.,
`application-networking.k8s.aws/port: "9200,9201,9203"`

docs/api-types/service-export.md
```diff
@@ -0,0 +1,38 @@
...
+
+Internally, creating a ServiceExport creates a VPC Lattice [target group](https://docs.aws.amazon.com/vpc-lattice/latest/ug/target-groups.html).
+You can take advantage of this by using the created target group in VPC Lattice setup outside Kubernetes.
+
+For more multi-cluster use cases and detailed explanation, please refer to [our recommended multi-cluster architecture](../guides/multi-cluster.md).
+
+### Limitations
+* The exported Service can only be used in HTTPRoutes. GRPCRoute is currently not supported.
+* Limited to one ServiceExport per Service. If you need multiple exports representing each port,
+  you should create multiple Service-ServiceExport pairs.
```
zijun726911, line 19: What this mean?
```
If you need multiple exports representing each port, you should create multiple Service-ServiceExport pairs.
```
Sounds like we support multiple ServiceExports per Service?

docs/api-types/service-import.md
```diff
@@ -0,0 +1,54 @@
...
+* ServiceImport shares the limitations of [ServiceExport](service-export.md).
+* The controller only supports ServiceImport through HTTPRoute; sending traffic directly is not supported.
+* BackendRef ports pointing to ServiceImport is not respected. Use [port annotation](service-export.md#annotations) of ServiceExport instead.
+
+### Annotations
+* `application-networking.k8s.aws/aws-eks-cluster-name`  
+  Indicates the cluster name owning the exported service.
+* `application-networking.k8s.aws/aws-vpc`  
+  Indicates the VPC ID owning the exported service.
+
```
zijun726911, line 25: Should we mention these 2 annotations are optional? The controller is able to automatically identify cluster name and vpc if not specified?

docs/api-types/service-export.md
solmonk: There is almost no chance that users will need comma-separated list behavior here, tbh I don't like that feature and I think it only causes confusion. I don't want to highlight that use case as an example.

docs/api-types/service-import.md
solmonk: this is a good point, I'll make a note of that use case.

docs/api-types/service-export.md
solmonk: I think it is important to know that creating a ServiceExport is an officially recommended method to create a target group. Will try to rephrase.

docs/api-types/grpc-route.md
```diff
@@ -27,7 +27,7 @@ This allows you to define and manage the routing of gRPC traffic within your Kub
 
 ### Annotations
 
-- `application-networking.k8s.aws/lattice-assigned-domain-name`
+- `application-networking.k8s.aws/lattice-assigned-domain-name`
```
xWink, line 30: nit: trailing space

docs/api-types/http-route.md
```diff
@@ -26,7 +26,7 @@ This allows you to define and manage the routing of HTTP and HTTPS traffic withi
 
 ### Annotations
 
-- `application-networking.k8s.aws/lattice-assigned-domain-name`
+- `application-networking.k8s.aws/lattice-assigned-domain-name`
```
xWink, line 29: same nit as above

docs/api-types/service-export.md
xWink: I would avoid making suggestions about future support at all if not necessary.

docs/api-types/service-import.md
solmonk: I can update this. btw the fields are optional but it is not like the controller automatically identifies them. They are like an additional sanity check when finding exported TG

docs/api-types/grpc-route.md
solmonk: It is actually intended to create a line break. Seems like backslash doesn't work in mkdocs so this is an alternative to it

docs/api-types/service-export.md
mikhail-aws: I would keep first sentence to stress "ServiceExport is not MCS". Short and clear. 

Also dont use vague language "a certain set of features." it applies to everything. You can enlist these features or comparison with mcs. It would be nice to elaborate why it's not MCS but we still use same term and reference.

solmonk: makes sense. I brought that phrase

docs/api-types/service-import.md
```diff
@@ -0,0 +1,59 @@
...
+`ServiceImport` is a resource referring to a Service outside the cluster. The primary use-case of ServiceImport is
+to pair with [`ServiceExport`](service-export.md) resource, defined in the other clusters.
+
+Just like Services, ServiceImports can be a backend reference of HTTPRoutes. Along with the cluster's own Services
+(and ServiceImports from even more clusters), you can distribute the traffic across multiple VPCs and clusters.
+
+Internally, `ServiceImport` is simply a reference to a VPC Lattice target group. Even if a target group is not created
+by ServiceExport, the controller can still recognize it as long as its name matches the ServiceImport.
+This way, you can also send traffic to target groups with EC2 or Lambda targets.
```
mikhail-aws, line 13: > the controller can still recognize it as long as its name matches the ServiceImport

I think "name match" is important to highlight. It should be exact name match with TG? I think it also can be TG-id and ARN, right?

Also curious for ServiceExport - do we create TG without random suffix?

solmonk: oh yes, I almost forgot that we have changed the way of identification. I think importing non-exported target group won't work anymore at this point. I will remove that use case instead.

docs/api-types/service-export.md
```diff
@@ -0,0 +1,39 @@
+# ServiceExport API Reference
+
+## Introduction
+
+In AWS Gateway API Controller, `ServiceExport` enables a Service for multi-cluster traffic setup.
+Clusters can import the exported service with [`ServiceImport`](service-import.md) resource.
+
+Internally, creating a ServiceExport creates a standalone VPC Lattice [target group](https://docs.aws.amazon.com/vpc-lattice/latest/ug/target-groups.html).
+Even without ServiceImports, creating ServiceExports can be useful in case you only need the target groups created;
+for example, using target groups in the VPC Lattice setup outside Kubernetes.
```
mikhail-aws, line 10: A second thought on it, I dont think we should advertise TG creation with ServiceExports, because TG now "ephemeral" with random suffix. It will break use of TG outside of controller.

solmonk: Target groups are not ephemeral, especially when created from ServiceExports. The only exception is when users intentionally change immutable fields, but that happens regardless of random suffix anyways.

zijun726911 (approved):

PR #556, opened 2023-11-30
Enable golangci-lint in GitHub Action and local

zijun726911:
**What type of PR is this?** Enable golangci-lint for local dev and GitHub Action

### Changes:
- Install golangci-lint in the toolchain.sh
- Add `golangci-lint` Makefile target, `presubmit` depends on it
- Run `golangci-lint` job in Github Action, run this workflow for every time people send Pull Requests

### Test
- Test  toolchain.sh install `golangci-lint` in local ( CI!=true) and github action (CI==true) environment, in local it could install golangci-lint if it is not yet installed in your machine. And in github action environment it will still to install (Since we have separate github action job to run the golangci-lint by ` uses: golangci/golangci-lint-action@v3` github plugin)
- Test `make presubmit` running `golangci-lint`  in local ( CI!=true) and github action (CI==true) environment both work as expect. local env can run it and CI environment skip it
- For GitHub Action happy path test, you could check this PR's "checks":
image

https://github.com/aws/aws-application-networking-k8s/actions/runs/7053275325/job/19200030069?pr=556

- For GitHub Action unhappy path test, you could see  this PR https://github.com/aws/aws-application-networking-k8s/pull/555/files, and for the code that violate golangci-lint check, you will see the error message in the PR "Files changed" tab:
image

**Will this PR introduce any new dependencies?**: developers need to run `make toolchain` again to install golangci-lint in their local 


Makefile
```diff
@@ -67,6 +67,20 @@ vet: ## Vet the code and dependencies
 			exit 1;\
 		fi;}
 
+.PHONY: golangci-lint
+golangci-lint:
```
mikhail-aws, line 71: can be just "lint"

Makefile
```diff
@@ -67,6 +67,20 @@ vet: ## Vet the code and dependencies
...
+# Run golangci-lint only from local machine
+	if [ "${CI}" != "true" ]; then \
+		if command -v golangci-lint &> /dev/null; then \
+			echo "Running golangci-lint"; \
+			golangci-lint run; \
+		else \
+			echo "Error: golangci-lint is not installed. Please run the 'make toolchain'"; \
+			exit 1; \
+		fi \
+	fi
```
mikhail-aws, line 81: why we need CI conditional?

zijun726911: In the new [presubmit.yaml](https://github.com/aws/aws-application-networking-k8s/blob/6fcdde8e913f85e92dcc40b4005e3c2f29a377cf/.github/workflows/presubmit.yaml), it has two jobs:
1. golangci-lint
2. presubmit
within the `presubmit` job , it does:
```
    - run: make toolchain
    - run: make presubmit (need to add a "${CI}" != "true" condition to prevent running golangci-lint)
    - run: make docs
    - name: Send coverage
```

Don't hope both the  `golangci-lint` github action job and `presubmit` github action job run the `golangci-lint run` two time, so we need a CI conditional

Makefile
zijun726911: ok, let me change it

Makefile
mikhail-aws: Does not make sense to me, if it's in presubmit then it should not be in separate action, or remove it from presubmit.

zijun726911: We need to use the `uses: golangci/golangci-lint-action@v3` github plugin  if its running as github action, this plugin has better integration will github, it could directly show report inline in the PR website UI: 
image

When people do local development, usually one step before people sending pull request is running `make presubmit`, people could see the   `golangci-lint` error message before send the PR

So, we need to config 2 way to run the golangci-lint.


Or, we could change people's habit to run bother `make lint` and  `make presubmit` before sending a PR?

mikhail-aws: I'm fine with that, you add new make step that will include lint and presubmit only for local dev if you want to save extra command to run

Makefile
```diff
@@ -53,7 +53,7 @@ run: ## Run in development mode
 
 
 .PHONY: presubmit
-presubmit: golangci-lint manifest vet test ## Run all commands before submitting code
+presubmit: lint manifest vet test ## Run all commands before submitting code
```
mikhail-aws, line 56: manifest should be first, it generates code

mikhail-aws: manifest vet lint test

mikhail-aws (approved):

zijun726911:
> Can you add a task please to split toolchain for development and CI into different scripts. There is very little overlap in practice. In CI we might use specialized containers and more complex testing environment. And for development we might need more people oriented tools.

Ok, make sense,  will do

PR #563, opened 2023-12-07
update validations and status on route reconciler

### LINKED ISSUE #277: Set status condition on HTTPRoute resources

### LINKED ISSUE #540: Error in logs when Service Network already have a VPC association, print memory addr instead of SNVA ARN
The logs shows the internal address of the VPC instead of the real value

```
aws-gateway-api-controller-aws-gateway-controller-chart-7bsfsd9 manager {"level":"debug","ts":"2023-11-22T13:05:42.660Z","logger":"controller.gateway","caller":"lattice/service_network_manager.go:195","msg":"ServiceNetwork app-services-gw already has VPC association %!s(*string=0xc0005fe0f0)"}
```

Seems the vpc are is a pointer, here we print it's address which is not useful 

https://github.com/aws/aws-application-networking-k8s/blob/1d3c19d94b706ba9f24ec1826128da79d3fda2a8/pkg/deploy/lattice/service_network_manager.go#L195

### LINKED ISSUE #544: The Route's status blindly change to ResolvedRefs==True even if the controller was NOT able to resolve all the object references for the Route
The Route's status blindly change to ResolvedRefs==True even if the controller was NOT able to resolve all the object references for the Route

How to reproduce it: for the controller version v1.0.1

1. `kubectl apply -f  examples/my-hotel-gateway.yaml`
2. Change the content of `examples/inventory-route.yaml`:
```
        kind: HTTPRoute
        metadata:
          name: inventory
        spec:
          parentRefs:
          - name: my-hotel
            sectionName: http
          rules:
          - backendRefs:
            - name: my-no-existing-service
              kind: Service
              port: 80
              weight: 10
```
3. do `kubectl apply -f  examples/inventory-route.yaml`
4.  Wait until the controller reconciled, I could see in the vpc lattice console that the lattice service `inventory-default` and serviceAssociation created successfully
5. do `kubectl get httproute  inventory -o yaml` I could see:
```
status:
  parents:
  - conditions:
    - lastTransitionTime: "2023-11-24T21:14:23Z"
      message: 'DNS Name: inventory-default-09b2aa51cffc3ccf5.7d67968.vpc-lattice-svcs.us-west-2.on.aws'
      observedGeneration: 1
      reason: Accepted
      status: "True"
      type: Accepted
    - lastTransitionTime: "2023-11-24T21:14:23Z"
      message: 'DNS Name: inventory-default-09b2aa51cffc3ccf5.7d67968.vpc-lattice-svcs.us-west-2.on.aws'
      observedGeneration: 1
      reason: ResolvedRefs
      status: "True"
      type: ResolvedRefs
    controllerName: application-networking.k8s.aws/gateway-api-controller
```
`ResolvedRefs == True` is not correct since my cluster don't have `my-no-existing-service` k8s Service. Also seem it miss the `Programmed` status

mikhail-aws:
Note:
Add Gateway API Spec validations to route controller. Reconciler will return non-retryable error when Spec is not valid.
Spec is not valid if there is at least single invalid property in ParentRef or BackendRef.

Tests:
- manually tested with #544 scenario
- unit test - partly covered by existing unit test - happy path
- e2e test - in progress

Example for new status:
```
Status:
  Parents:
    Conditions:
      Last Transition Time:  2023-12-07T19:32:33Z
      Message:
      Observed Generation:   5
      Reason:                Accepted
      Status:                True
      Last Transition Time:  2023-12-07T19:32:33Z
      Message:               backendRef name: inventory-ver1
      Observed Generation:   5
      Reason:                BackendNotFound
      Status:                False
      Type:                  ResolvedRefs
    Controller Name:         application-networking.k8s.aws/gateway-api-controller
    Parent Ref:
      Group:         gateway.networking.k8s.io
      Kind:          Gateway
      Name:          my-hotel
      Section Name:  http
```

close #544 
close #540 
close #277

### DIFF
--- a/pkg/controllers/route_controller.go
+++ b/pkg/controllers/route_controller.go
@@ -20,30 +20,26 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/aws/aws-application-networking-k8s/pkg/aws/services"
-	"github.com/aws/aws-application-networking-k8s/pkg/utils"
-	"github.com/aws/aws-application-networking-k8s/pkg/utils/gwlog"
-
 	"github.com/pkg/errors"
-
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/external-dns/endpoint"
+	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
 	gwv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 	gwv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 
-	"sigs.k8s.io/external-dns/endpoint"
-
-	"sigs.k8s.io/controller-runtime/pkg/builder"
-	"sigs.k8s.io/controller-runtime/pkg/predicate"
-	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
-
 	anv1alpha1 "github.com/aws/aws-application-networking-k8s/pkg/apis/applicationnetworking/v1alpha1"
 	"github.com/aws/aws-application-networking-k8s/pkg/aws"
+	"github.com/aws/aws-application-networking-k8s/pkg/aws/services"
 	"github.com/aws/aws-application-networking-k8s/pkg/config"
 	"github.com/aws/aws-application-networking-k8s/pkg/controllers/eventhandlers"
 	"github.com/aws/aws-application-networking-k8s/pkg/deploy"
@@ -52,6 +48,9 @@ import (
 	"github.com/aws/aws-application-networking-k8s/pkg/k8s"
 	"github.com/aws/aws-application-networking-k8s/pkg/model/core"
 	lattice_runtime "github.com/aws/aws-application-networking-k8s/pkg/runtime"
+	"github.com/aws/aws-application-networking-k8s/pkg/utils"
+	k8sutils "github.com/aws/aws-application-networking-k8s/pkg/utils"
+	"github.com/aws/aws-application-networking-k8s/pkg/utils/gwlog"
 )
 
 var routeTypeToFinalizer = map[core.RouteType]string{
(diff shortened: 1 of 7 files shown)


xWink:
Is the message of an Accepted status supposed to state the backendref's name in the spec, or is this a custom decision?

pkg/controllers/route_controller.go
```diff
@@ -456,3 +430,196 @@ func (r *routeReconciler) validateBackendRefsIpFamilies(ctx context.Context, rou
...
+
+// Validation for route spec. Will validate and update route status. Returns error if not valid.
+// Validation rules are suppose to be compliant to Gateway API Spec.
+//
+//	https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.RouteConditionType
+//
+// There are 3 condition types: Accepted, PartiallyInvalid, ResolvedRefs.
+// We dont support PartiallyInvalid for now and reject entire route if there is at least one invalid field.
+// Accepted type is related to parentRefs, and ResolvedRefs to backendRefs. These 2 are validated independently.
+func (r *routeReconciler) validateRoute(ctx context.Context, route core.Route) error {
```
xWink, line 448: Variable names in this function are hard to understand. Let's please stick to camelCase words

pkg/controllers/route_controller.go
```diff
@@ -456,3 +430,196 @@ func (r *routeReconciler) validateBackendRefsIpFamilies(ctx context.Context, rou
...
+	if err != nil {
+		return err
+	}
+
+	cnd, err := r.validateBackedRefs(ctx, route)
+	if err != nil {
+		return err
+	}
+
+	// we need to update each parentRef with backendRef status
```
xWink, line 459: Why do we need to update parentRef with backendRef status?

mikhail-aws:
> Is the message of an Accepted status supposed to state the backendref's name in the spec, or is this a custom decision?

message is primarily for human readable error details
- if it's not found indicate what is not found
- if it's error kind indicate which kind it's and may be provide list of valid kinds
etc.

pkg/controllers/route_controller.go
```diff
@@ -456,3 +430,196 @@ func (r *routeReconciler) validateBackendRefsIpFamilies(ctx context.Context, rou
...
+
+// Validation rules for route parentRefs
+//
+// Will ignore status update when:
+// - parentRef does not exists, includes when parentRef Kind is not Gateway
+//
+// If parent GW exists will check:
+// - NoMatchingParent: parentRef sectionName and port matches Listener name and port
+// - TODO: NoMatchingListenerHostname: listener hostname matches one of route hostnames
+// - TODO: NotAllowedByListeners: listener allowedRoutes contains route GroupKind
```
xWink, line 519: Do we have action items associated with these TODOs?

pkg/model/core/grpcroute.go
```diff
@@ -79,6 +79,13 @@ func (r *GRPCRoute) Inner() *gwv1alpha2.GRPCRoute {
 	return &r.r
 }
 
+func (r *GRPCRoute) GK() metav1.GroupKind {
```
xWink, line 82: Why GK() instead of GroupKind()?

mikhail-aws: its GK or GKV in other places of kubernetes: apimachinery for example

pkg/controllers/route_controller.go
mikhail-aws: yes

pkg/model/core/grpcroute.go
xWink: I am only able to find examples of GroupKind(), such as [here](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/builder/controller.go#L367C17-L367C17)

mikhail-aws: https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.16.3/pkg/client/apiutil

if it bothers you I can change

pkg/controllers/route_controller.go
mikhail-aws: updated

xWink (approved):

Issue #565, opened 2023-12-08
Controller failing to associate lattice service with service network in different account.


prys:
I have a VPC Lattice Service Network ('my-hotel') defined in AWS Account A. This is RAM-shared to Account B, where I am running my EKS cluster. I can see the Service Network when logged into the console and cli of Account B. I can successfully associate my EKS VPC with the 'my-hotel' service network. 

Following the Quick Start guide, I deploy the 'parking' and 'review' kubernetes deployments and services into my cluster, and then the 'rates' HTTPRoute resource using the rate-route-path.yaml in the example folder. This creates the 'rates-default' lattice service and two target groups - 'k8s-default-review-qmsuilsxyq' and 'k8s-default-parking-koyqfohudh'. These are visible in the console of Account B, but not visible from Account A.

The controller fails to associate the 'rates-default' service with the 'my-hotel' Service Network. With debug logging on, the controller logs: -

`{"level":"debug","ts":"2023-12-08T11:03:40.393Z","logger":"cloud","caller":"aws/cloud.go:65","msg":"response","serviceName":"VPC Lattice","operation":"ListServiceNetworks","params":"{\n\n}"}
{"level":"info","ts":"2023-12-08T11:03:40.394Z","logger":"controller.route","caller":"controllers/route_controller.go:155","msg":"reconcile error","name":"rates","message":"error during service synthesis failed ServiceManager.Upsert rates-default due to Service network my-hotel not found"}`

I am able to manually associate the 'rates-default' service with the 'my-hotel' service network via the console, logged in to Account B.

Have I missed something to allow the controller to create lattice services / target groups _and_ associate with a lattice service in a different aws account?

mikhail-aws:
There is confirmed issue on controller/lattice side with RAM. It might be your case, but I recommend double check on your side.

There was recent change on controller side to use tags for ownership for shared resource like SN. When SN is shared with you through RAM there is no option to allow ListTagsForResource.

`aws vpc-lattice list-tags-for-resource --resource-arn ${shared_sn_arn}`

This call will fail, even if you grant "full" permission in RAM AWSRAMPermissionVpcLatticeServiceNetworkReadWrite
You should see something like this in debug or error logs:

[...]

PR #566, opened 2023-12-08
change FindServiceNetwork algo, use name and id match

mikhail-aws:
Note:

Address bug for RAM sharing scenario with ServiceNetwork. Currently FindServiceNetwork will not find shared SNs, but if find will fail on fetching Tags. It's not possible to fetch tags for external resources.

Important changes in vpclattice.go

New algorithm will do:
- List all SN
- Filter by Name match
- Filter by ID match. I use ID as fallback strategy. If there multiple SN with same name (which is possible with RAM sharing, but not in same account). Operator can create GW with SN-ID as a name to disambiguate naming.

Added unit tests.
Manually tested with shared SN scenarios.

New error log example:
```
reconcile error	{"name": "share-to-mhlbrz", "message": "multiple SN found by name, cannot disambiguate: [name=share-to-mhlbrz, id=sn-09ee4f73b2985af2b name=share-to-mhlbrz, id=sn-08542bb99cb2f3bae]"}
```

Related #565


pkg/aws/services/vpclattice.go
```diff
@@ -313,60 +318,78 @@ func (d *defaultLattice) ListServiceNetworkServiceAssociationsAsList(ctx context
...
 	}
-	if snMatch == nil {
-		return nil, NewNotFoundError("Service network", name)
+
+	// try to fetch tags only if SN in the same aws account with controller's config
+	tags := Tags{}
+	if snArn.AccountID == d.ownAccount || d.ownAccount == "" {
+		tagsInput := vpclattice.ListTagsForResourceInput{ResourceArn: snMatch.Arn}
+		tagsOutput, err := d.ListTagsForResourceWithContext(ctx, &tagsInput)
+		if err != nil {
```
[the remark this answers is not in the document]
mikhail-aws: It's true that we are not using tags for SN. But we might need them for cross-cluster same aws account sharing. For example for read-only access. I'm ok to remove them, but I have sense that we will need them very soon once we dig more into cross-cluster use-cases.

pkg/aws/services/vpclattice.go
```diff
@@ -104,17 +105,17 @@ type Lattice interface {
 	ListTargetsAsList(ctx context.Context, input *vpclattice.ListTargetsInput) ([]*vpclattice.TargetSummary, error)
 	ListServiceNetworkVpcAssociationsAsList(ctx context.Context, input *vpclattice.ListServiceNetworkVpcAssociationsInput) ([]*vpclattice.ServiceNetworkVpcAssociationSummary, error)
 	ListServiceNetworkServiceAssociationsAsList(ctx context.Context, input *vpclattice.ListServiceNetworkServiceAssociationsInput) ([]*vpclattice.ServiceNetworkServiceAssociationSummary, error)
-	FindServiceNetwork(ctx context.Context, name string, accountId string) (*ServiceNetworkInfo, error)
+	FindServiceNetwork(ctx context.Context, name string) (*ServiceNetworkInfo, error)
```
zijun726911, line 108: Need to change name to  `nameOrId string` in interface?

pkg/aws/services/vpclattice.go
```diff
@@ -313,60 +318,78 @@ func (d *defaultLattice) ListServiceNetworkServiceAssociationsAsList(ctx context
...
+	// match. Ideally name match should work just fine, but in desperate
+	// scenario of shared SN when naming collision happens using ID can be
+	// an option
+	var snMatch *vpclattice.ServiceNetworkSummary
+	nameMatch := utils.SliceFilter(allSn, func(snSum *vpclattice.ServiceNetworkSummary) bool {
+		return aws.StringValue(snSum.Name) == nameOrId
+	})
+	idMatch := utils.SliceFilter(allSn, func(snSum *vpclattice.ServiceNetworkSummary) bool {
+		return aws.StringValue(snSum.Id) == nameOrId
 	})
```
zijun726911, line 351: For the discussion in slack, filter by name then id make sense. We could also add another filter by ARN, i.e. change function name to 
`FindServiceNetwork(ctx context.Context, nameOrIdOrArn string)` and add a `arnMatch=`

pkg/aws/services/vpclattice.go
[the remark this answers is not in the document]
zijun726911: > we might need them for cross-cluster same aws account sharing

I am still not quite sure what same account cross-cluster sharing need tags. How's the SN read-only access looks like?

pkg/aws/services/vpclattice.go
mikhail-aws: I would drop ID and use ARN only

pkg/aws/services/vpclattice.go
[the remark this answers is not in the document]
mikhail-aws: It's really hard to make SN fit into Gateway spec and include lattice features.

pkg/aws/services/vpclattice.go
mikhail-aws: ARN is not a valid name for Gateway, "/" if not allowed

zijun726911 (approved):

PR #574, opened 2023-12-15
Add more controller permissions in recommended-inline-policy.json

zijun726911:
- Add more controller permissions in recommended-inline-policy.json
- Fix e2e test code compiling error

### Tests

New permissions could run access_log_policy_test.go success

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,6 +66,7 @@ vet: ## Vet the code and dependencies
 		if [ "${CI}" = true ]; then\
 			exit 1;\
 		fi;}
+	cd test && go vet ./...
 
 
 .PHONY: lint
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -66,6 +66,7 @@ vet: ## Vet the code and dependencies
 		if [ "${CI}" = true ]; then\
 			exit 1;\
 		fi;}
+	cd test && go vet ./...
```
mikhail-aws, line 69: go vet ./... should include all sub directories, it's not module based afaik

mikhail-aws (changes requested):

Makefile
zijun726911: From my test, seems the root dir `go vet ./...` cannot detect e2etest folder compiling error

mikhail-aws: what about go vet ./test/... ?

zijun726911: I just tried many `go vet` commands, seems only `cd test && go vet ./...` can work

mikhail-aws (approved):

Issue #575, opened 2023-12-18
env-config configmap losing values on helm upgrade


hscheib:
Hi,

I just helm upgraded from v1.0.1 to v1.0.2 using mostly default values
```
  values:
    serviceAccount:
      create: false
      name: vpc-lattice-controller
    deployment:
      replicas: 2
```

Now the new pods fail to start with this kube event

```
3s          Warning   Failed              pod/gateway-api-controller-aws-gateway-controller-chart-647859v68n9         Error: couldn't find key awsRegion in ConfigMap aws-application-networking-system/env-config
```

I investigated and sure enough the defaulted empty fields are gone

```
apiVersion: v1
items:
- apiVersion: v1
  data:
    logLevel: info
  kind: ConfigMap
```

Originally the configmap looks like 
```
apiVersion: v1
items:
- apiVersion: v1
  data:
    awsAccountId: ""
    awsRegion: ""
    clusterName: ""
    clusterVpcId: ""
    defaultServiceNetwork: ""
    latticeEndpoint: ""
    logLevel: info
  kind: ConfigMap
```

I did update my helm values to 

```
  values:
    awsRegion: ""
    awsAccountId: ""
    clusterVpcId: ""
    clusterName: ""
    defaultServiceNetwork: ""
    latticeEndpoint: ""
    serviceAccount:
      create: false
      name: vpc-lattice-controller
    deployment:
      replicas: 2
```

and the helm upgrade succeeded and 1.0.2 pods spun up.

So I am not sure if the default values in the chart's values.yaml need to be set to `""` instead of blank or I am required to supply `""` values?  

I followed https://www.gateway-api-controller.eks.aws.dev/guides/deploy/#controller-installation and am under the impression I did not need to supply these values from the comments since we are on EKS, don't want to create a defaultServiceNetwork, and expect IMDS to be available
```
   # awsRegion, clusterVpcId, awsAccountId are required for case IMDS is not available.
   --set=awsRegion= \
   --set=clusterVpcId= \
   --set=awsAccountId= \
   # clusterName is required except for EKS cluster.
   --set=clusterName= \
   # When specified, the controller will automatically create a service network with the name.
   --set=defaultServiceNetwork=my-hotel
```

Please advise, thanks!

zijun726911:
Hi hscheib, thank you to report this bug. I am able to reproduce this issue. If I install a brand new controller by:
```
helm install gateway-api-controller \
[...]

Issue #541, opened 2023-11-22
Remove configmap from helm template


mikhail-aws:
We use configmap as intermediate env variable placeholder for helm. I find it redundant and just another place to make a mistake when change env variables. Proposal is to move from this pattern:
```
values.yaml
  clusterName: 

templates/configmap.yaml
  data:
    clusterName: {{ .Values.clusterName | quote }}

templates/deployment.yaml
  ...
  env:
    - name: CLUSTER_NAME
      valueFrom:
        configMapKeyRef:
          name: env-config
          key: clusterName
```

Into this pattern:

```
values.yaml
  clusterName:

templates/deployment.yaml
  ...
  env:
    - name: CLUSTER_NAME
      value: {{ .Values.clusterName | quote }}
```

Reference from https://github.com/bitnami/charts/tree/main/bitnami. A collection of helm charts.

PR #576, opened 2023-12-18
Fix helm chart config map default value issue #575 

zijun726911:
Fix the issue #575 

Reference: 
https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function

### Testing

Use the new code to create a new image version  [v1.0.3-rc.1](https://gallery.ecr.aws/aws-application-networking-k8s/aws-gateway-controller-chart). Used the v1.0.3-rc.1 to test the upgrade:

1. Install an old version helm chart:
```
helm install gateway-api-controller \
      --version=v1.0.2 \
      --set=serviceAccount.create=false --namespace aws-application-networking-system \
       --set=log.level=debug
```
Confirmed controller pods are in Running status, checked the controller logs , it's running good.

2. Upgrade to  v1.0.3-rc.1 :
```helm upgrade gateway-api-controller \
      --version=v1.0.3-rc.1 \
      --set=serviceAccount.create=false --namespace aws-application-networking-system \
       --set=log.level=debug

```
Confirmed controller pods upgrade success and they are in Running status. And  env-config for  v1.0.3-rc.1 shows the correct data:
```
kubectl get configMap env-config  -n aws-application-networking-system -o yaml
apiVersion: v1
data:
  awsAccountId: ""
  awsRegion: ""
  clusterName: ""
  clusterVpcId: ""
  defaultServiceNetwork: ""
  latticeEndpoint: ""
  logLevel: debug
kind: ConfigMap
....
```
(If I don't pass `--set=log.level=debug` , the logLevel in configMap is `info` as expected)

### DIFF
--- a/helm/templates/configmap.yaml
+++ b/helm/templates/configmap.yaml
@@ -3,10 +3,10 @@ kind: ConfigMap
 metadata:
     name: env-config
 data:
-    awsRegion: {{ .Values.awsRegion | quote }}
-    awsAccountId: {{ .Values.awsAccountId | quote }}
-    clusterVpcId: {{ .Values.clusterVpcId | quote }}
-    clusterName: {{ .Values.clusterName | quote }}
-    latticeEndpoint: {{ .Values.latticeEndpoint | quote }}
-    defaultServiceNetwork: {{ .Values.defaultServiceNetwork | quote }}
-    logLevel: {{ .Values.log.level | quote }}
+    awsRegion: {{ .Values.awsRegion | default "" | quote }}
+    awsAccountId: {{ .Values.awsAccountId | default ""| quote }}
+    clusterVpcId: {{ .Values.clusterVpcId | default "" | quote }}
+    clusterName: {{ .Values.clusterName | default "" | quote }}
+    latticeEndpoint: {{ .Values.latticeEndpoint | default "" | quote }}
+    defaultServiceNetwork: {{ .Values.defaultServiceNetwork  | default "" | quote }}
+    logLevel: {{ .Values.log.level | default "info" | quote }}


mikhail-aws:
my personal preference would be to remove configMap from helm, I dont see any value in using configMap other than another place that can fail

https://github.com/aws/aws-application-networking-k8s/issues/541

mikhail-aws:
This issue is another prove that configMap is a nuisance, second time we are missing update that causes problems.

zijun726911:
I think it make sense to remove configMap, I found another issue during my personal test: Updating configMap content is not able to trigger gateway-api-controller-aws-gateway-controller-chart `Deployment` rollingUpdate. i.e.,:
 if I do this first:
```
 helm install gateway-api-controller \
      --version=v1.0.3-rc.1 \
      --set=serviceAccount.create=false --namespace aws-application-networking-system \
       --set=log.level=debug
```

And then do the: (remove the  --set=log.level=debug) 
```
helm upgrade gateway-api-controller \
      --version=v1.0.3-rc.1 \
      --set=serviceAccount.create=false --namespace aws-application-networking-system \
```

The gateway-api-controller-aws-gateway-controller-chart `Deployment` is not able to **preceive** this ENV variable change and the controller Deployment will not do rollingUpdate and the log is still the debug level.



I think same reason with:

```
templates/deployment.yaml
  ...
  env:
    - name: CLUSTER_NAME
      valueFrom:
        configMapKeyRef:
          name: env-config
          key: clusterName

```

Remove  env-config ConfigMap Probably could solve it

helm/templates/deployment.yaml
```diff
@@ -74,41 +74,17 @@ spec:
           readOnlyRootFilesystem: true
         env:
           - name: REGION
-            valueFrom:
-              configMapKeyRef:
-                name: env-config
-                key: awsRegion
-          - name: AWS_ACCOUNT_ID
```
mikhail-aws, line 81: why removed?

mikhail-aws (changes requested):
please update pr description

helm/templates/deployment.yaml
zijun726911: Oops, it's a typo

mikhail-aws (approved):
Nice, thank you

mikhail-aws:
close #541

zijun726911:
I am testing the  `v1.0.3-rc.3` image in which deployment.yaml has (without the `| default "" ` ):     

```
    env:
          - name: REGION
            value: {{ .Values.awsRegion | quote }}
          - name: AWS_ACCOUNT_ID
            value: {{ .Values.awsAccountId | quote }}
          - name: CLUSTER_VPC_ID
            value: {{ .Values.clusterVpcId | quote }}
          - name: CLUSTER_NAME
            value: {{ .Values.clusterName | quote }}
          - name: LATTICE_ENDPOINT
            value: {{ .Values.latticeEndpoint | quote }}
          - name: DEFAULT_SERVICE_NETWORK
            value: {{ .Values.defaultServiceNetwork | quote }}
          - name: LOG_LEVEL
            value: {{ .Values.log.level | quote }}
```

I think it's better, plan to send a new commit.

mikhail-aws:
what will happen after update with previous configmap? will it be removed?

PR #580, opened 2023-12-29
Add TG garbage collector

### LINKED ISSUE #552: Optimize unused target group cleanup
Unused target group cleanup is called at the end of every route reconciliation - this means if the controller is to reconcile 100 target groups, we are cleaning up target groups 100 times. Target group cleanup is a pretty heavy operation even when cleaning nothing, involving a few round trip of list/get calls. This is impacting controller throughput of creating resources, and degrading performance of controller startup.

We could either:
* make it an asynchronous process that runs every X minute.
* somehow make it not happen when there was a recent cleanup

We should also make sure to avoid race condition when taking the first approach, because current target group cleanup logic is built on the assumption that reconciliation is done. The simplest way might be just to put locks - not the best looking but will still be better than current logic in terms of performance.

mikhail-aws:
Note:
Address performance issue for unused target group cleanup. Implemented background garbage collector, that runs every X seconds. Might need to tune timers after. 

e2e tests pass:

```
Ran 58 of 58 Specs in 2358.172 seconds
SUCCESS! -- 58 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestIntegration (2358.17s)
```
close #552

### DIFF
--- a/pkg/deploy/stack_deployer.go
+++ b/pkg/deploy/stack_deployer.go
@@ -175,7 +175,7 @@ func (gc *TgGc) cycle() {
 	if err != nil {
 		gc.log.Debugf("gc cycle error: %s", err)
 	}
-	gc.log.Debugw("tg gc result",
+	gc.log.Debugw("gc stats",
 		"delete_attempts", res.att,
 		"delete_success", res.succ,
 		"duration", res.duration,
@@ -189,11 +189,12 @@ func (d *latticeServiceStackDeployer) Deploy(ctx context.Context, stack core.Sta
 	listenerSynthesizer := lattice.NewListenerSynthesizer(d.log, d.listenerManager, stack)
 	ruleSynthesizer := lattice.NewRuleSynthesizer(d.log, d.ruleManager, d.targetGroupManager, stack)
 
-	// we need to block GC when we deploy stack
-	// stack deployer first creates TG and then associate TG with Service
-	// if GC will run in between it can delete newly created TG before it associated
-	// this lock also prevents concurrent deployments, only one deployment cat run at the time
-	// TODO: This place can become a contention. May be add debug log with waiting time for lock?
+	// We need to block GC when we deploy stack. Stack deployer first creates TG and then
+	// associate TG with Service. If GC will run in between it can delete newly created TG
+	// before association since it's dangling TG. This lock also prevents concurrent
+	// deployments, only one deployment can run at the time.
+	//
+	// TODO: This place can become a contention. May be debug log with lock waiting time?
 	defer func() {
 		tgGc.lock.Unlock()
 	}()
--- a/pkg/deploy/stack_deployer.go
+++ b/pkg/deploy/stack_deployer.go
@@ -82,7 +82,7 @@ func NewLatticeServiceStackDeploy(
 		tgGcFn := NewTgGcFn(tgGcSynth)
 		tgGc = &TgGc{
 			lock:    sync.Mutex{},
-			log:     log.Named("gc"),
+			log:     log.Named("tg-gc"),
 			ctx:     context.TODO(),
 			isDone:  atomic.Bool{},
 			ivl:     TG_GC_IVL,
@@ -189,6 +189,11 @@ func (d *latticeServiceStackDeployer) Deploy(ctx context.Context, stack core.Sta
 	listenerSynthesizer := lattice.NewListenerSynthesizer(d.log, d.listenerManager, stack)
 	ruleSynthesizer := lattice.NewRuleSynthesizer(d.log, d.ruleManager, d.targetGroupManager, stack)
 
+	// we need to block GC when we deploy stack
+	// stack deployer first creates TG and then associate TG with Service
+	// if GC will run in between it can delete newly created TG before it associated
+	// this lock also prevents concurrent deployments, only one deployment cat run at the time
+	// TODO: This place can become a contention. May be add debug log with waiting time for lock?
 	defer func() {
 		tgGc.lock.Unlock()
 	}()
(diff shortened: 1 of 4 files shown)


pkg/deploy/stack_deployer.go
```diff
@@ -54,35 +60,140 @@ type latticeServiceStackDeployer struct {
...
 	targetGroupSynthesizer := lattice.NewTargetGroupSynthesizer(d.log, d.cloud, d.k8sClient, d.targetGroupManager, d.svcExportTgBuilder, d.svcBuilder, stack)
 	targetsSynthesizer := lattice.NewTargetsSynthesizer(d.log, d.cloud, d.targetsManager, stack)
 	serviceSynthesizer := lattice.NewServiceSynthesizer(d.log, d.latticeServiceManager, d.dnsEndpointManager, stack)
 	listenerSynthesizer := lattice.NewListenerSynthesizer(d.log, d.listenerManager, stack)
 	ruleSynthesizer := lattice.NewRuleSynthesizer(d.log, d.ruleManager, d.targetGroupManager, stack)
 
+	defer func() {
+		tgGc.lock.Unlock()
+	}()
+	tgGc.lock.Lock()
```
zijun726911, line 195: Can you add some comments here for the reason we need to do Lock and Unlock here?

pkg/deploy/stack_deployer.go
```diff
@@ -54,35 +60,140 @@ type latticeServiceStackDeployer struct {
...
+
+	tgGcOnce.Do(func() {
+		// TODO: need to refactor TG synthesizer. Remove stack from constructor
+		// arguments and use it as Synth argument. That will help with Synth
+		// reuse for GC purposes
+		tgGcSynth := lattice.NewTargetGroupSynthesizer(log, cloud, k8sClient, tgMgr, tgSvcExpBuilder, svcBuilder, nil)
+		tgGcFn := NewTgGcFn(tgGcSynth)
+		tgGc = &TgGc{
+			lock:    sync.Mutex{},
+			log:     log.Named("gc"),
```
zijun726911, line 85: log.Named("tgGc") ?

pkg/deploy/stack_deployer.go
```diff
@@ -54,35 +60,140 @@ type latticeServiceStackDeployer struct {
...
+func NewTgGcFn(tgSynth *lattice.TargetGroupSynthesizer) TgGcCycleFn {
+	return func(ctx context.Context) (TgGcResult, error) {
+		t0 := time.Now()
+		results, err := tgSynth.SynthesizeUnusedDelete(ctx)
+		if err != nil {
+			return TgGcResult{}, err
+		}
+		succ := 0
+		for _, res := range results {
+			if res.Err == nil {
```
zijun726911, line 120: Do we need to print `res.Err` or do errors.Join(res.Err) and return it with TgGcResult{}?

zijun726911 (commented):
Reviewed this PR and overall LGTM

pkg/deploy/stack_deployer.go
mikhail-aws: usually log names are lower case, can do "tg_gc" or "tg.gc"

pkg/deploy/stack_deployer.go
mikhail-aws: Not really, mostly errors happen when TG is in state that cannot be deleted yet. It pollutes logs.

pkg/deploy/stack_deployer.go
mikhail-aws: ok

pkg/deploy/stack_deployer.go
mikhail-aws: I think critical errors needs to be logged in SynthesizeUnusedDelete, or need to wrap them into some "NonOk" types and do type check here.

pkg/deploy/stack_deployer.go
xWink: Would we ever garbage collect services? I think this would be the only GC

pkg/deploy/stack_deployer.go
```diff
@@ -54,35 +60,140 @@ type latticeServiceStackDeployer struct {
...
 	k8sClient client.Client,
 ) *latticeServiceStackDeployer {
 	brTgBuilder := gateway.NewBackendRefTargetGroupBuilder(log, k8sClient)
 
+	tgMgr := lattice.NewTargetGroupManager(log, cloud)
+	tgSvcExpBuilder := gateway.NewSvcExportTargetGroupBuilder(log, k8sClient)
+	svcBuilder := gateway.NewLatticeServiceBuilder(log, k8sClient, brTgBuilder)
+
+	tgGcOnce.Do(func() {
+		// TODO: need to refactor TG synthesizer. Remove stack from constructor
```
xWink, line 78: Is this TODO a future issue?

xWink:
Overall LGTM, love the use of goroutines here.

xWink (approved):

zijun726911 (approved):

pkg/deploy/stack_deployer.go
mikhail-aws: I think we should not use GC at all. It will be a problem in future - global lock will cause contention and need carefully lock/unlock otherwise we will deadlock route reconciliations. This solution is a little pain relief from doing full TG clean up on each reconciliation loop.

We should have a better mechanism to identify when we create brand new TG or when we replace old one and deletion is required. In that case we dont need locks, since problem will be properly partitioned by single-threaded owners - reconcilers.

pkg/deploy/stack_deployer.go
mikhail-aws: We should revisit synthesizers, the concept of abstract stack that we build and how it's actually implemented is questionable.

Issue #365, opened 2023-08-28
Support GRPC ServiceExport


solmonk:
Currently GRPC target groups are only created if a Service is direct backendRef of GRPCRoute. It is not supported in case of ServiceExport, thus not available for ServiceImport either.

Since at the time of TG creation we are not aware of RouteType (Route may not even exist at this point), the ideal solution to create both HTTP/GRPC target groups for ServiceExport. Our key-value cache structure is making this really hard to achieve, because RouteType is not part of the key. Changing cache key is a big one - requires touching every part involving target groups, but seems like the only way to proceed.

evgeniy-khatko:
# Description
Hello, at Venmo we also need this capability. The high level requirements would be very similar to how `HTTPRoute` works, for example: 
```
apiVersion: gateway.networking.k8s.io/v1alpha2
kind: GRPCRoute
metadata:
  name: grpc-service-import
spec:
  parentRefs:
    - name: gateway-name
      sectionName: https
  rules:
    - backendRefs:
        - name: grpc-service-01
          kind: Service
          port: 50051
          weight: 50
        - name: grpc-service-02
          kind: ServiceImport
          port: 50051
          weight: 50
```
## Context
`GRPCRoute` already allows specifying `ServiceImport`-s as `backendRef`-s, but created `HTTPTargetGoup`-s are of type HTTP1. Tests also show that creating gRPC target groups independently and using them as targets in pre-created GRPCRoute works fine. So we assume cross-cluster gRPC support can be added with proper configuration and documentation changes?

rlymbur:
Thank you for your suggestions. I have addressed this issue in #756.

PR #583, opened 2024-01-09
ServiceExport gRPC Support

xWink:
**What type of PR is this?**

feature

#365 

- Creates a gRPC VPC Lattice TargetGroup alongside the HTTP one when a ServiceExport is created for a deployment
  - The gRPC target group is also updated or deleted alongside the HTTP one during ServiceExport update/delete
- Adds additional checks to existing ServiceExport e2e tests

**Automation added to e2e**:

- Added new checks to e2e tests
- Manual test by creating Greeter ServiceExport for Greeter Service

No

No

Yes

```release-note
Added support for gRPC traffic with ServiceExports
```


pkg/gateway/model_build_targetgroup.go
```diff
@@ -87,7 +87,7 @@ func (b *SvcExportTargetGroupBuilder) Build(
 	return task.stack, nil
 }
 
-func (b *SvcExportTargetGroupBuilder) BuildTargetGroup(ctx context.Context, svcExport *anv1alpha1.ServiceExport) (*model.TargetGroup, error) {
+func (b *SvcExportTargetGroupBuilder) BuildTargetGroups(ctx context.Context, svcExport *anv1alpha1.ServiceExport) (*model.TargetGroup, *model.TargetGroup, error) {
```
mikhail-aws, line 90: Since we are not following mcs spec anymore for import/export there is no constraint to create both HTTP/gRPC TG's at the same time. We can extend CRD with type of service and build only one TG that we need.

mikhail-aws: @solmonk

PR #587, opened 2024-01-17
Fix target group discovery logic

solmonk:
**What type of PR is this?**

bug

Fixes target group discovery logic that may cause service exports from other clusters are not shown.

I think this VPC-based filtering logic got into during the TG performance optimization. I don't think this adds a lot of performance value anyways so I completely removed it. List() call is currently used in two places:
* Unused target group cleanup. This filters result by VPC regardless.
* ServiceExport detection. This will now be fixed.

Unit testing and manual cross-cluster testing that verifies traffic.

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/deploy/lattice/target_group_manager.go
+++ b/pkg/deploy/lattice/target_group_manager.go
@@ -11,7 +11,6 @@ import (
 	"github.com/aws/aws-sdk-go/service/vpclattice"
 
 	pkg_aws "github.com/aws/aws-application-networking-k8s/pkg/aws"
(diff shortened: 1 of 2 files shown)


zijun726911 (approved):

allamand:
I will be happy to test this

zijun726911:
@allamand we will release a new vision helm chart and artifact soon to include this fix

allamand:
I’ve tested and it works as expected. Thanks for the quick patch and release

PR #590, opened 2024-01-19
Improve github-release.sh

zijun726911:
Improve github-release.sh, delete local release branch/ tag if it exist, show error message if release branch/ tag  exists remotely

### DIFF
--- a/scripts/lib/common.sh
+++ b/scripts/lib/common.sh
@@ -49,11 +49,9 @@ function sed_inplace() {
 
     if [[ "$OSTYPE" == "darwin"* ]]; then
         # macOS uses BSD sed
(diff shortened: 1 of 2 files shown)


scripts/lib/common.sh
```diff
@@ -49,9 +49,11 @@ function sed_inplace() {
 
     if [[ "$OSTYPE" == "darwin"* ]]; then
         # macOS uses BSD sed
+        echo "sed -i '' \"s|$pattern|$replacement|g\" \"$file\""
```
solmonk, line 52: you might want to look into something like `set -x` instead

zijun726911: ok, let me remove it, I think people are free to add `set -x` by themself when debugging

solmonk (approved):

Issue #132, opened 2023-03-15
enable e2e test to run automatically after each github PR or merge


liwenwu-amazon:
enable [e2e test](https://github.com/aws/aws-application-networking-k8s/pull/118) on github after each PR or Merge

erikfuller:
This has been done.

PR #597, opened 2024-02-02
Add e2e test automation workflow

solmonk:
**What type of PR is this?**

feature

#132 

* Adds e2e test on pull request check.
* Updated release script
  * Lifted some constraint to handle image build without release
  * Added option to do image build / push only, not helm
  * Removed information such as AWS account ID

Notes:
* One caveat is this only works on the existing test cluster - ideally, we can further improve the script by creating EKS cluster on demand. 
* This workflow will be put into `e2e-test` deployment environment so that it requires manual approval to run. I think this is needed for security purpose as well as ensuring only one e2e test is running at the time.

Tested in the forked repo

**Automation added to e2e**:
No new test cases added

```release-note

```

### DIFF
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -1,0 +1,79 @@
+name: e2e-test
+permissions:
+  id-token: write
+  contents: read
+on:
+  pull_request:
+  workflow_dispatch:
+jobs:
+  e2e-test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 240
+    environment: e2e-test
+    env:
+      # Image build config
+      AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
+      IMAGE_REPOSITORY: ${{ secrets.PRIVATE_IMAGE_REPOSITORY }}
+      ECR_PUBLISH_ROLE_ARN: ${{ secrets.ECR_PUBLISH_IAM_ROLE }}
+      COMMIT_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
+
+      # E2E integration test config
+      REGION: ${{ secrets.AWS_DEFAULT_REGION }}
+      CLUSTER_NAME: ${{ secrets.CLUSTER_NAME }}
+      CLUSTER_VPC_ID: ${{ secrets.CLUSTER_VPC_ID }}
+      SECONDARY_ACCOUNT_TEST_ROLE_ARN: ${{ secrets.SECONDARY_ACCOUNT_IAM_ROLE }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.GO_VERSION }}
+          check-latest: true
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Setup AWS credentials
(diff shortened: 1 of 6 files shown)


solmonk:
Still not sure if we should run the test directly over pull request's head. This is something we want ideally, but the e2e test suite requires pretty powerful permission. We can always make the change later though, as it is a simple fix.

solmonk:
a successful run from my own fork: https://github.com/solmonk/aws-application-networking-k8s/actions/runs/7761811200/job/21171173210

scripts/lib/login.sh
```diff
@@ -9,11 +9,21 @@
...
+    local __pw=$(aws ecr-public get-login-password --region us-east-1)
+    echo "$__pw" | docker login --username AWS --password-stdin public.ecr.aws
+    export HELM_EXPERIMENTAL_OCI=1
+    echo "$__pw" | helm registry login -u AWS --password-stdin public.ecr.aws
+  else
+    local __pw=$(aws ecr get-login-password)
+    echo "$__pw" | docker login --username AWS --password-stdin $registry_url
+    export HELM_EXPERIMENTAL_OCI=1
+    echo "$__pw" | helm registry login -u AWS --password-stdin $registry_url
+  fi
```
zijun726911, line 26: Nit, could simplify to this way:


```
if [ "$registry_url" = "public.ecr.aws" ]; then
    local __pw=$(aws ecr-public get-login-password --region us-east-1)
  else
    local __pw=$(aws ecr get-login-password --region $(aws configure get region))
  fi
  echo "$__pw" | docker login --username AWS --password-stdin $registry_url
  export HELM_EXPERIMENTAL_OCI=1
  echo "$__pw" | helm registry login -u AWS --password-stdin $registry_url
```

.github/workflows/e2e-test.yaml
```diff
@@ -0,0 +1,82 @@
+name: e2e-test
+permissions:
+  id-token: write
+  contents: read
+on:
+  push:
+    branches:
+      - main
```
zijun726911, line 8: Can we add a TODO here? e.g, in the future, the repo admin is able to **manually** trigger a e2e test for a specific PR number(git commit)

scripts/release-controller.sh
```diff
@@ -49,110 +55,64 @@ check_is_installed git
...
+      "${DOCKER_BUILD_CONTEXT}"
+  else
+    docker buildx build -t "$IMG" \
+      -f "$CONTROLLER_IMAGE_DOCKERFILE_PATH" \
+      --build-arg service_controller_git_version="$VERSION" \
+      --build-arg service_controller_git_commit="$CONTROLLER_GIT_COMMIT" \
+      --build-arg build_date="$BUILD_DATE" \
+      --push \
+      "${DOCKER_BUILD_CONTEXT}"
+  fi
```
zijun726911, line 102: Out of the scope: We can build both amd64 and arm64 image by `--platform` option for the controller public release

zijun726911 (commented):
Overall looks good

zijun726911:
> I have first thought about running this on the pull request's context, but seems like it is a security hazard, as the e2e test runner is capable of doing pretty much everything on VPC Lattice + has some other AWS permissions on the internal account.

Seems current workflow has a `workflow_dispatch`,

Does it currently support to manually trigger a e2e test workflow against the PR's feature branch? 

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

zijun726911 (approved):

Issue #579, opened 2023-12-26
Support Multi-Arch Builds(ARM64)


kamirendawkins:
Currently the Gateway API Controller Pods do not appear to support ARM64 Arch, as an AWS consumer using Graviton where possible is a preference so multi-arch support would be greatly appreciated.

Issue based on:
1. Public ECR not specifying supported archs: https://gallery.ecr.aws/aws-application-networking-k8s/aws-gateway-controller
1. Pods launched on ARM based instances resulting in:
  ```bash
  exec /manager: exec format error
  ```

mikhail-aws:
Yes, there is no ARM64 support right now and it would be nice to support. There are few active tasks related to our build and release automation which should help us with new platforms.

mikhail-aws:
We can do few tweaks in our release script, to make it easier run custom builds, so you can build it by yourself for any platform. That should serve short term, until we have proper support for ARM.

@zijun726911 , I see there are values in script that hard-coded for our own build(ECR_PUBLISH_ROLE_ARN), can we make them customizable? like this:

```
IMAGE_TAG=${IMAGE_TAG:-$VERSION}
```

https://github.com/aws/aws-application-networking-k8s/blob/main/scripts/release-controller.sh

zijun726911:
Yes, will try to figure out how to build multi platform in the next release

https://github.com/aws/amazon-vpc-cni-k8s/blob/048a66a55cae2e7e7c9123b5dc6b159c1a4d04f2/Makefile#L173

zijun726911:
Done in the #598

PR #598, opened 2024-02-07
Optimize docker build and support multi-arch (ARM64)

solmonk:
**What type of PR is this?**

feature

#579 - should be able to support ARM64 next release.

Fixed dockerfile to support multiarch and optimized build to utilize cache

Tested using personal fork

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -79,7 +79,7 @@ jobs:
           kubectl apply -f examples/gatewayclass.yaml
       - name: Run test
         run: |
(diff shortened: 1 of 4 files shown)


zijun726911:
Are different arch images reside in the same repo? how user select different arch to deploy helm chart to their k8s cluster? can we mention it in our doc?

solmonk:
All platforms should be part of the same image. I believe users don't need to select anything, k8s picks it up based on the platform.

zijun726911 (approved):

PR #604, opened 2024-02-15
Listen to endpointslice event

solmonk:
**What type of PR is this?**

feature

Not yet - this is an initiation for pod readiness support project.

* Record pod readiness on targets model builder
* Listen to endpointslice event, has multiple benefits:
  * has advanced read of status
  * can scale to 1000+ pods, old endpoint objects have limits on 1000 addresses
* Keep the old behavior yet - non-ready targets are filtered at targets manager. this will be changed on the next PR.

Splitting PR to keep reviewability.

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/deploy/lattice/targets_manager.go
+++ b/pkg/deploy/lattice/targets_manager.go
@@ -50,6 +50,12 @@ func (s *defaultTargetsManager) Update(ctx context.Context, modelTargets *model.
 			modelTg.ID(), modelTargets.Spec.StackTargetGroupId)
 	}
 
+	// Only take care of pods that are ready, for backwards compatibility.
+	// TODO: Pod readiness support.
+	modelTargets.Spec.TargetList = utils.SliceFilter(modelTargets.Spec.TargetList, func(t model.Target) bool {
+		return t.Ready
+	})
+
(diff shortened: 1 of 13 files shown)


pkg/deploy/lattice/targets_manager.go
```diff
@@ -50,6 +50,12 @@ func (s *defaultTargetsManager) Update(ctx context.Context, modelTargets *model.
 			modelTg.ID(), modelTargets.Spec.StackTargetGroupId)
 	}
 
+	// Only take care of pods that are ready, for backwards compatibility.
+	// TODO: Pod readiness support.
+	modelTargets.Spec.TargetList = utils.SliceFilter(modelTargets.Spec.TargetList, func(t model.Target) bool {
+		return t.Ready
+	})
```
erikfuller, line 57: We should have a unit test that ensures we filter out items that aren't ready. Happy to have this as a follow up.

erikfuller (approved):
lgtm

pkg/deploy/lattice/targets_manager.go
erikfuller: Though I think the readiness support means we won't do this filter anymore anyway.

Issue #596, opened 2024-01-29
Inject pod readiness gate


solmonk:
In addition to supporting pod readiness gate conditions, ideally we need to "inject" `readinessGates` field via webhook.

We have to make sure only the desired resources are impacted. [aws-load-balancer-controller](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.1/deploy/pod_readiness_gate/) defines a label to address this problem - when a pod is created under the labelled namespace, the controller webhook will update the pod definition with `readinessGates` field. We can refer to the documentation and take the similar approach.

If the user needs more fine-grained control than namespaces, they may still define `readinessGates` field manually on pods / pod templates instead.

Requires #560 - but can be developed independently.

erikfuller:
Adding some notes here as automatically injecting the readiness gate will require creation of a kubernetes secret if installying with ```deploy.yaml```. Helm has functionality to provision these certificates and secrets, so that should all be automatically configured.

Plan is to use the following for manually provisioning the certificate and configuring the webhook:
```
CERT_FILE=tls.crt
KEY_FILE=tls.key

WEBHOOK_SVC_NAME=webhook-service
WEBHOOK_NAME=aws-appnet-gwc-mutating-webhook
WEBHOOK_NAMESPACE=aws-application-networking-system
WEBHOOK_SECRET_NAME=webhook-cert

# Step 1: generate a certificate if needed, can also be provisioned through orgnanizational PKI, etc
# This cert includes a 100 year expiry
HOST=${WEBHOOK_SVC_NAME}.${WEBHOOK_NAMESPACE}.svc
openssl req -x509 -nodes -days 36500 -newkey rsa:2048 -keyout ${KEY_FILE} -out ${CERT_FILE} -subj "/CN=${HOST}/O=${HOST}" \
   -addext "subjectAltName = DNS:${HOST}, DNS:${HOST}.cluster.local"
   
# Step 2: add the secret - can be done so long as the namespace exists
# note that running "kubectl delete -f deploy.yaml" will remove the controller namespace AND this secret.
kubectl create secret tls $WEBHOOK_SECRET_NAME --namespace $WEBHOOK_NAMESPACE --cert=${CERT_FILE} --key=${KEY_FILE}

# Step 3: after applying deploy.yaml, patch the webhook CA bundle to exactly the cert being used
# this will ensure Kubernetes API server is able to trust the certificate presented by the webhook
CERT_B64=$(cat tls.crt | base64)
[...]

PR #605, opened 2024-02-26
Adding core webhook libraries sourced from aws lb controller

erikfuller:
**What type of PR is this?**
feature

https://github.com/aws/aws-application-networking-k8s/issues/596

This PR adds useful types for handling webhook invocation. They are sourced from the [aws-load-balancer-controller](https://github.com/kubernetes-sigs/aws-load-balancer-controller/tree/main/pkg/webhook), which I believe originally came from the [aws-app-mesh-controller-for-k8s](https://github.com/aws/aws-app-mesh-controller-for-k8s/tree/master/pkg/webhook).

Changes from the original were minor, but I've added notes where appropriate

```make presubmit```

All unit tests pass.

**Automation added to e2e**:
n/a

no

n/a

no

### CODE REVIEW

File: pkg/webhook/core/mutating_handler.go [hunk 1]
```diff
@@ -0,0 +1,83 @@
...
+var mutatingHandlerLog = ctrl.Log.WithName("mutating_handler")
+
+type mutatingHandler struct {
+	mutator Mutator
+	decoder *admission.Decoder
+}
+
+func (h *mutatingHandler) SetDecoder(d *admission.Decoder) {
+	h.decoder = d
+}
```
erikfuller, line 21: The biggest difference is that the other repos are not working off the v1 spec for admission control and include an API to "inject" a decoder. Instead, we have to create the decoder manually.

pkg/webhook/core/mutator.go
```diff
@@ -0,0 +1,28 @@
...
+	// MutateUpdate handles Object update and returns the object after mutation and error if any.
+	MutateUpdate(ctx context.Context, obj runtime.Object, oldObj runtime.Object) (runtime.Object, error)
+}
+
+// MutatingWebhookForMutator creates a new mutating Webhook.
+func MutatingWebhookForMutator(scheme *runtime.Scheme, mutator Mutator) *admission.Webhook {
+	return &admission.Webhook{
+		Handler: &mutatingHandler{
+			mutator: mutator,
+			decoder: admission.NewDecoder(scheme),
```
erikfuller, line 25: This is where we set the decoder instead of having it injected automatically.

mikhail-aws:
Curious what controller's runtime and kubebuilder does not provide, so we need to implement own mutators.

https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.17.2/pkg/webhook/admission
https://github.com/kubernetes-sigs/controller-runtime/blob/main/examples/builtins/mutatingwebhook.go
https://github.com/kubernetes-sigs/controller-runtime/blob/main/examples/builtins/validatingwebhook.go

PR #606, opened 2024-02-27
Inject pod readiness gate

erikfuller:
**What type of PR is this?**
feature

https://github.com/aws/aws-application-networking-k8s/issues/596

This PR adds a pod mutating webhook to inject the controller's pod readiness gate.

One of the biggest changes is that webhooks must be network-reachable by the API server. Previously, the controller only required communication from itself to the API server. This means local testing against a remote cluster is very challenging. Most of my testing was done by pushing a custom ECR image for the controller and deploying it to my test cluster.

Webhook invocation is *only* over TLS, so the controlller must now present a TLS certificate to the API server. This certificate is validated by its DNS name(s) and must also pass trust validation checks. More details are in the ```pod-readiness-gates.md``` file.

For the TLS cert we require a standard kubernetes tls secret called ```webhook-cert```. It must be present in the same namespace as the controller itself. The certificate is mounted as read-only by the controller container at ```/etc/webhook-cert```.

By default, the webhook includes a namespace label selector, meaning the webhook will only be called for namespaces which include the label ```aws-application-networking-k8s/pod-readiness-gate-inject``` with value ```enabled```.

Otherwise, this work is based off the webhook work in [aws-load-balancer-controller](https://github.com/kubernetes-sigs/aws-load-balancer-controller) and [aws-app-mesh-controller-for-k8s](https://github.com/aws/aws-app-mesh-controller-for-k8s). Note that, unlike the load balancer controller, the webhook configuration here is created by default. Note however, the webhook logic is only enabled by the namespace label described above, otherwise the webhook will not execute or inject the controller readiness gate.

**What is left to be done?**
These will follow in separate PRs:
1. The actual pod readiness gate implementation
2. Helm chart update to support the webhook and configure the TLS certificate
3. Update to build scripts to automatically run webhook e2e tests

The general testing process:
```console
# after build and deploy image to ECR 
# deploy controller (and webhook) to cluster
kubectl create -f deploy.yaml 

# recreate the webhook secret
kubectl delete secret tls $WEBHOOK_SECRET_NAME --namespace $WEBHOOK_NAMESPACE
kubectl create secret tls $WEBHOOK_SECRET_NAME --namespace $WEBHOOK_NAMESPACE --cert=${CERT_FILE} --key=${KEY_FILE}

# patch the webhook with the cert details to enable trust from API server
kubectl patch mutatingwebhookconfigurations.admissionregistration.k8s.io aws-appnet-gwc-mutating-webhook --namespace $WEBHOOK_NAMESPACE --type='json' -p="[{'op': 'replace', 'path': '/webhooks/0/clientConfig/caBundle', 'value': '${CERT_B64}'}]"
mutatingwebhookconfiguration.admissionregistration.k8s.io/aws-appnet-gwc-mutating-webhook patched

# run manual tests OR webhook-e2e-tests
make webhook-e2e-test
```

**Automation added to e2e**:
Since webhooks must be invoked by the API server, they do not fit into the standard ```make e2e-test```, at least for local development and testing. Instead, I have created a new make target ```make webhook-e2e-test```. These tests can be run in a local dev environment against a remote cluster once the controller is deployed and the webhook secret updated.

```console
make webhook-e2e-test
=== RUN   TestIntegration
Running Suite: WebhookIntegration - /.../aws-application-networking-k8s/test/suites/webhook
===============================================================================================================
Random Seed: 1709056486

Will run 2 of 2 specs
[SynchronizedBeforeSuite] 
/.../aws-application-networking-k8s/test/suites/webhook/suite_test.go:23
[SynchronizedBeforeSuite] PASSED [0.000 seconds]
Readiness Gate Inject create deployment in untagged namespace, no readiness gate
/.../aws-application-networking-k8s/test/suites/webhook/readiness_gate_inject_test.go:35
{"level":"info","ts":"2024-02-27T09:54:46.310-0800","caller":"test/framework.go:264","msg":"Deleting objects: *v1.Namespace/webhook-e2e-test-no-tag, *v1.Namespace/webhook-e2e-test-tagged"}
{"level":"info","ts":"2024-02-27T09:54:46.409-0800","caller":"test/framework.go:282","msg":"Waiting for NotFound, objects: *v1.Namespace/webhook-e2e-test-no-tag, *v1.Namespace/webhook-e2e-test-tagged"}
{"level":"info","ts":"2024-02-27T09:54:46.425-0800","caller":"test/framework.go:186","msg":"Creating objects: *v1.Namespace/webhook-e2e-test-no-tag, *v1.Namespace/webhook-e2e-test-tagged"}
{"level":"info","ts":"2024-02-27T09:54:46.447-0800","caller":"test/framework.go:186","msg":"Creating objects: *v1.Deployment/sparr-1-inmjcpummz"}
{"level":"info","ts":"2024-02-27T09:54:46.506-0800","caller":"test/pod_manager.go:68","msg":"deployment.Spec.Selector.MatchLabels: map[app:untagged-test-pod]"}
• [0.215 seconds]
Readiness Gate Inject create deployment in tagged namespace, has readiness gate
/.../aws-application-networking-k8s/test/suites/webhook/readiness_gate_inject_test.go:56
{"level":"info","ts":"2024-02-27T09:54:46.526-0800","caller":"test/framework.go:186","msg":"Creating objects: *v1.Deployment/finge-2-hicyysf5br"}
{"level":"info","ts":"2024-02-27T09:54:46.569-0800","caller":"test/pod_manager.go:68","msg":"deployment.Spec.Selector.MatchLabels: map[app:tagged-test-pod]"}
{"level":"info","ts":"2024-02-27T09:54:46.588-0800","caller":"test/framework.go:264","msg":"Deleting objects: *v1.Namespace/webhook-e2e-test-no-tag, *v1.Namespace/webhook-e2e-test-tagged"}
{"level":"info","ts":"2024-02-27T09:54:46.608-0800","caller":"test/framework.go:282","msg":"Waiting for NotFound, objects: *v1.Namespace/webhook-e2e-test-no-tag, *v1.Namespace/webhook-e2e-test-tagged"}
• [11.283 seconds]
[SynchronizedAfterSuite] 
/.../aws-application-networking-k8s/test/suites/webhook/suite_test.go:39
[SynchronizedAfterSuite] PASSED [0.000 seconds]

Ran 2 of 2 Specs in 11.500 seconds
SUCCESS! -- 2 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestIntegration (11.50s)
PASS
ok  	github.com/aws/aws-application-networking-k8s/test/suites/webhook	12.869s
```

We now have a dependency on the presence of the webhook secret. Note that, if the webhook is not invoked

I tested deploying with the previous build's ```deploy.yaml```, then applying the new ```deploy.yaml``` which includes the webhook configuration, without issue.

Yes

```release-note
Support for automatically injecting pod readiness gate. For more details see docs/guides/pod-readiness-gates.md
```


cmd/aws-application-networking-k8s/main.go
```diff
@@ -128,14 +128,24 @@ func main() {
...
+	isLocalDev := config.DevMode != ""
+	if !isLocalDev {
+		webhookServer = k8swebhook.NewServer(k8swebhook.Options{
+			Port:     9443,
+			CertDir:  "/etc/webhook-cert/",
+			CertName: "tls.crt",
+			KeyName:  "tls.key",
+		})
+	}
+
```
erikfuller, line 142: When using ```make run```, DevMode will be non-empty, so locally running the controller will skip this webhook configuration. Otherwise, the controller will only start if the webhook cert files are present (at ```/etc/webhook-cert/tls.crt``` and ```tls.key```).

config/default/kustomization.yaml
```diff
@@ -1,26 +1,10 @@
-# Adds namespace to all resources.
```
erikfuller, line 1: Much of this commented code came from the aws-load-balancer-controller. I've tried to remove what we don't need. One notable difference is that the gateway controller enables the webhook configuration by default, even though you still have to opt into it for any given namespace.

config/manager/manager.yaml
```diff
@@ -60,5 +60,26 @@ spec:
...
+            name: webhook-cert
+            readOnly: true
       serviceAccountName: gateway-api-controller
       terminationGracePeriodSeconds: 10
+      volumes:
+        - name: webhook-cert
+          secret:
+            defaultMode: 420
+            secretName: webhook-cert
+---
```
erikfuller, line 74: This placeholder certificate and key were generated with:

```
openssl req -x509 -nodes -days -365 -newkey rsa:2048 -keyout ${KEY_FILE} -out ${CERT_FILE} -subj "/CN=not-a-real-cn/O=not-a-real-o"
```

When calling a webhook, the API server performs the following validations:
1. Cert validity window has not passed
2. SAN matches DNS (e.g. ```webhook-service.aws-application-networking-system.svc```)
3. Cert is signed by a known authority OR matches the ```ca``` field on the client configuration for the webhook (either signer or direct match)

Having a placeholder here is important because the controller's webhook server needs this value to start up successfully. The webhook server does not perform strict validation.

Once the secret is patched with a valid cert, Kubernetes will distribute the value and the webhook server will be notified when the files are updated on the volume mount. See [controller-runtime/pkg/webhook/server.go](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/webhook/server.go#L207) for more details.

docs/guides/pod-readiness-gates.md
```diff
@@ -0,0 +1,144 @@
+# Pod readiness gate
```
erikfuller, line 1: Good amount of structure/content sourced from [aws-load-balancer-controller](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/deploy/pod_readiness_gate.md)

config/manager/manager.yaml
erikfuller: I've removed the placeholder values entirely and instead populate them during ```make build-deploy```

pkg/webhook/pod_readiness_gate_injector.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"context"
+	"github.com/aws/aws-application-networking-k8s/pkg/utils/gwlog"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	PodReadinessGateConditionType = "aws-application-networking-k8s/pod-readiness-gate"
```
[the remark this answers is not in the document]
erikfuller: I don't really have a preference, other than it being a self-descriptive value so people can know what it is when they see it on their pods. Happy to change this.

.gitignore
```diff
@@ -18,4 +18,8 @@ go.work*
 pkg/aws/services/gomock_reflect_*
 
 # Image build tarballed bundles
-*.tgz
\ No newline at end of file
+*.tgz
+
+# generated during make-deploy
+tls.crt
+tls.key
```
solmonk, line 25: I think it is better to remove them at the end of Makefile

docs/guides/pod-readiness-gates.md
```diff
@@ -0,0 +1,144 @@
...
+      - enabled
+  objectSelector:
+    matchLabels:
+      aws-application-networking-k8s/pod-readiness-gate-inject: enabled
+  ...
+```
+When you specify multiple selectors, pods matching all the conditions will get mutated.
+
+## Disabling the readiness gate inject
+You can specify the controller flag `--enable-pod-readiness-gate-inject=false` during controller startup to disable the controller from modifying the pod spec.
```
solmonk, line 116: Is this feature implemented? having hard time finding this

pkg/webhook/pod_readiness_gate_injector.go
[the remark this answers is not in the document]
solmonk: cool, I think it is better to change

.gitignore
erikfuller: Good idea. WIll change this.

docs/guides/pod-readiness-gates.md
erikfuller: Good catch. It is not. This is something in the load balancer controller. Will scrub through this file again now that we have the other PR open too.

erikfuller: removed

.gitignore
erikfuller: added ```rm``` to makefile, removed from ```.gitignore```

pkg/webhook/pod_readiness_gate_injector.go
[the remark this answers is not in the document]
erikfuller: done

solmonk (approved):

PR #627, opened 2024-04-29
Add support for clusters running in private VPC

aaroniscode:
NOTE: This is a draft PR for review by the team. If the design of the feature is approved, I will add any required tests.

**What type of PR is this?**
feature

no open Issue

Add support for clusters running in private VPC.  Today the controller won't function in a fully private VPC as there is no PrivateLink support for the ResourceTagging API

Create a private VPC (one that cannot access the Internet) and try to the [Getting Started Guide](https://github.com/aws/aws-application-networking-k8s/blob/main/docs/guides/getstarted.md)

ran the controller locally and tested the Getting Started Guide

**Automation added to e2e**:
none at this time

No

No breakage

Yes.

```release-note
New `DISABLE_TAGGING_SERVICE_API` environment variable. If you are running a cluster in a Private VPC, this will allow Tag queries using the Lattice API which can be accessed using a PrivateLink endpoint.

```


pkg/aws/cloud.go
```diff
@@ -25,6 +26,7 @@ type CloudConfig struct {
 	AccountId   string
 	Region      string
 	ClusterName string
+	PrivateVPC  bool
```
erikfuller, line 29: I think it might be better to be explicit here, as over time this field could be interpreted in different ways. For example, at some stage the Tagging API will likely be enabled for PrivateLink, but then the customer still might not have provisionined the VPC endpoint for Tagging. Once they eventually do, their VPC is still "private" but then they would remove the "PrivateVPC" flag to get to the tagging service, so that semantic doesn't quite feel right.

Instead, we could have ```TaggingServiceAPIDisabled``` which defaults to ```false``` but then for configurations that can't access that service they can set it to ```true```. Later, when they can use the API, they set it back to ```false```.

pkg/aws/cloud.go
```diff
@@ -38,6 +40,12 @@ type Cloud interface {
 	// creates lattice tags with default values populated and merges them with provided tags
 	DefaultTagsMergedWith(services.Tags) services.Tags
 
+	// find tags on lattice resources
+	FindTagsForARNs(ctx context.Context, arns []string) (map[string]services.Tags, error)
+
+	// find lattice target group ARNs using tags
+	FindTargetGroupARNs(context.Context, services.Tags) ([]string, error)
+
```
erikfuller, line 48: I like the idea of including these methods in the Tagging() interface. I think we should change the current interface to remove the inherited ```taggingapiiface.ResourceGroupsTaggingAPIAPI``` interface. Then you can have two implementations, one that uses the tagging API and one that uses the Lattice tag APIs. Then, in ```NewCloud(...)``` you can check the ```TaggingServiceAPIDisabled``` option and assign one implementation or the other, rather than having switches inside each method.

erikfuller (commented):
Thanks for putting this together! I think we can move forward with some minor tweaks. Please let me know if you have any questions or concerns on my feedback. Please also be sure to include unit tests for the new logic.

pkg/aws/services/tagging.go
erikfuller, line 1: So it doesn't look like there are any unit tests for this file yet, probably because there wasn't really much business logic in here previously.

Please create a new `tagging_test.go` file. It's probably OK to skip `GetTagsForArns` since there isn't much there. For FindResourcesByTags you'll want tests that check:

1. non-target group types return error
2. no results for ListTargetGroups returns empty list
3. one item with tags are found
4. multiple items with tags are found
4. tags are not found (a. target groups do not have tags, b. tag case mismatch, c. only subset of tags match)
5. error on ListTags returns error

pkg/config/controller_config.go
```diff
@@ -82,6 +84,12 @@ func configInit(sess *session.Session, metadata EC2Metadata) error {
 		ServiceNetworkOverrideMode = true
 	}
 
+	privateVPC := os.Getenv(DISABLE_TAGGING_SERVICE_API)
```
erikfuller, line 87: Please update `privateVPC` variable name to `disableTaggingServiceAPI` or similar

pkg/aws/services/tagging.go
```diff
@@ -80,6 +86,66 @@ func NewDefaultTagging(sess *session.Session, region string) *defaultTagging {
...
+}
+
+func containsTags(source, check Tags) bool {
+	for k, v := range source {
+		if aws.StringValue(check[k]) != aws.StringValue(v) {
+			return false
+		}
+	}
+	return true
+}
```
erikfuller, line 147: You might want to create some separate unit tests just for this. I actually think this logic is currently incorrect. For example, I believe a `source` with tag `"X"=""` will return true even if the `check` does not contain the `"X"` tag. I think if you had something like `checkV, ok := check[k]` and ensure `ok` is `true` that would be enough.

Here, you would want to comprehensively check all the options between the two sets where you want it to return true or false. I listed a few above, but you might want more here like tags being present, having different values, or having matching keys but with different tags. 

This logic has to be super precise, so let's be sure to get good edge case coverage here.

erikfuller (changes requested):
Overall looking good. Added suggestions for unit tests, but we can also talk more over Slack if you need more details.

While this will unblock the "private VPC" use case, I am concerned about the performance implications for this approach, since these O(1) operations are now O(N) in terms of number of remote API calls. I suppose when performance becomes an issue we can address that too - there are other places in the codebase we'll also likely have to revisit from a performance standpoint.

pkg/aws/services/tagging.go
aaroniscode: thank you so much for the guidance on how to approach the unit tests ... tests added!

pkg/aws/services/tagging.go
aaroniscode: you are correct about the error!  I created a unit test where this failed and corrected it in the code as you suggested to make the test pass. Note that I was using the arguments backwards and now range over check (the tags we want to see if the source contains).

pkg/config/controller_config.go
aaroniscode: oops, didn't realize I missed this. Updated.

aaroniscode:
@erikfuller I think the PR has addresses all your feedback and I appreciate the test guidance. Also rebased as there was a conflict.

pkg/aws/services/tagging_test.go
```diff
@@ -0,0 +1,421 @@
...
+			findTags: Tags{
+				"Key1": aws.String("Value1"),
+				"Key2": aws.String("Value2"),
+			},
+			sourceTags: Tags{
+				"Key1": aws.String("Value1"),
+				"Key2": aws.String("Value2"),
+				"Key3": aws.String("Value3"),
+			},
+			testName: "match subnet of tags from source's multiple tags",
```
erikfuller, line 65: "subset"

pkg/aws/services/tagging_test.go
```diff
@@ -0,0 +1,421 @@
...
+		{
+			contains: true,
+			findTags: Tags{
+				"Key1": aws.String("Value1"),
+				"Key2": aws.String("Value2"),
+			},
+			sourceTags: Tags{
+				"Key1": aws.String("Value1"),
+				"Key2": aws.String("Value2"),
+				"Key3": aws.String("Value3"),
```
erikfuller, line 63: just to mix this up can you please reverse the ordering here? So key3, key2, key1, just to make sure we aren't relying on ordering in our logic.

pkg/aws/services/tagging_test.go
```diff
@@ -0,0 +1,421 @@
...
+			},
+			testName: "no match for tag with different key case",
+		},
+		{
+			contains: false,
+			findTags: Tags{
+				"Key1": aws.String("Value"),
+			},
+			sourceTags: Tags{},
+			testName:   "no match for tag when source has no tags",
```
erikfuller, line 113: Might also be good to include a case where "findTags" is empty.

pkg/aws/services/tagging_test.go
```diff
@@ -0,0 +1,421 @@
...
+func Test_latticeTagging_FindResourcesByTags_nonTargetGroupTypeError(t *testing.T) {
+	c := gomock.NewController(t)
+	ctx := context.TODO()
+	mockLattice := NewMockLattice(c)
+	lt := &latticeTagging{Lattice: mockLattice}
+
+	nonTargetGroupResourceType := ResourceTypeService
+
+	_, err := lt.FindResourcesByTags(ctx, nonTargetGroupResourceType, Tags{})
+	assert.NotNil(t, err)
```
erikfuller, line 161: This is fine, but I think `assert.Error` is a bit more idiomatic.

erikfuller (changes requested):
Thanks, @aaroniscode, tests are looking good. Just a typo and a couple small adjustments and we'll be good to merge.

pkg/aws/services/tagging_test.go
aaroniscode: Done

pkg/aws/services/tagging_test.go
aaroniscode: Done

pkg/aws/services/tagging_test.go
aaroniscode: Done. Had to change the `containsTags` function to cover this use case.

pkg/aws/services/tagging_test.go
aaroniscode: Done

PR #624, opened 2024-04-24
Adding Trace IDs to log lines

shulin-sq:
**What type of PR is this?**

**Which issue does this PR fix**:
https://github.com/aws/aws-application-networking-k8s/issues/603

related conversation: https://github.com/uber-go/zap/issues/654

**What does this PR do / Why do we need it**:

**If an issue # is not available please add repro steps and logs from aws-gateway-controller showing the issue**:

**Testing done on this change**:

running in a development environment:
```
{"level":"info","ts":"2024-04-24T02:23:58.515Z","logger":"controller.route","caller":"controllers/route_controller.go:157","msg":"reconcile starting","type":"route","name":"lattice-redacted-srv-http","trace_id":"bc67d345-3b33-499a-9e3e-329e4dff92b9"}
{"level":"info","ts":"2024-04-24T02:23:58.515Z","logger":"controller.route","caller":"controllers/route_controller.go:277","msg":"Found aws-vpc-lattice for Route for lattice-redacted-srv-http, redacted-srv","trace_id":"bc67d345-3b33-499a-9e3e-329e4dff92b9"}
{"level":"info","ts":"2024-04-24T02:23:58.515Z","logger":"controller.route","caller":"controllers/route_controller.go:323","msg":"reconcile, adding or updating","name":"lattice-redacted-srv-http","trace_id":"bc67d345-3b33-499a-9e3e-329e4dff92b9"}
{"level":"info","ts":"2024-04-24T02:23:58.524Z","logger":"controller.iam-auth-policy","caller":"controllers/iamauthpolicy_controller.go:85","msg":"reconcile starting","type":"iamauthpolicy","name":"lattice-redacted-srv-http","trace_id":"ccdf0c8a-6fd0-43d6-86d4-5e4840e4a28c"}
{"level":"info","ts":"2024-04-24T02:23:58.524Z","logger":"controller.iam-auth-policy","caller":"controllers/iamauthpolicy_controller.go:96","msg":"reconcile IAM policy","req":"redacted-srv/lattice-redacted-srv-http","targetRef":{"group":"gateway.networking.k8s.io","kind":"HTTPRoute","name":"lattice-redacted-srv-http"},"tra
ce_id":"ccdf0c8a-6fd0-43d6-86d4-5e4840e4a28c"}
[...]

Issue #603, opened 2024-02-14
Add trace ids and context to log lines


shulin-sq:
Currently log lines from the controller are difficult to attribute between one another. For example

```
	r.log.Infow("reconcile", "name", req.Name)
	recErr := r.reconcile(ctx, req)
	if recErr != nil {
		r.log.Infow("reconcile error", "name", req.Name, "message", recErr.Error())
```

can result in
```
reconcile name serviceA
reconcile name serviceB
// stuff happening for serviceB (with no indiciatioon that log line is for serviceB)
// stuff happening for serviceA (with no indication that log line is for serviceA)
reconcile serviceA finished
reconcile error name serviceB error etc
```
making it difficult to debug errors since the log lines of multiple resources/events are intermixed.

suggestion:
* Starting from when we receive an event from the kubernetes api, start a trace and add this to context or log line context
* for every log line that occurs due to that event trigger, append the trace id.

therefore you can see
* what caused the reconciliation to be triggered
* metadata about what resource was worked on
* at what step an error occurred and for what resource.

Another way to approach this is to add some metadata to every log line, such as what resource we are trying to reconcile. But I think trace might be easier to implement.

references:
* https://stripe.com/blog/canonical-log-lines is a similar idea. In addition to adding metadata to log lines, the idea here is add a summary message of what happened at the end of a request handler (or in this case event handler)
* https://pkg.go.dev/context#WithValue how I've seen metadata be passed around in the past via context
* https://go.dev/blog/context mentions that context is passed in for every request, which might make context a good candidate for storing this data.

If we like the context idea, I can give it a go with some pointers on anything I should be aware of with the logging library used in the project.

erikfuller:
Hi @shulin-sq, I think this will be really helpful. I believe the main logging setup is in ```pkg/utils/gwlog/gwlog.go```.

If you can get a PoC going with the context approach and share it as a draft PR or branch we can have a look before you get too deep in the implementation.

PR #648, opened 2024-06-07
Add Trace IDs and metadata to log lines

shulin-sq:
**What type of PR is this?**
feature (?)

https://github.com/aws/aws-application-networking-k8s/issues/603

This PR adds the ability to track trace id and metadata to all log lines. It also changes the log interface so that in most cases, developers are forced to add a context instance to the log line.

The previous PR was here: https://github.com/aws/aws-application-networking-k8s/pull/624 but was based on an older release version. I decided to recreate the branch on top of main.

This PR also addresses some of the comments brought up in the previous PR

Additional resources:
- the reconcile start/end is inspired from https://stripe.com/blog/canonical-log-lines
- this is also a feature that has been requested in zap https://github.com/uber-go/zap/issues/654

ran 
```
go test ./...
```

and also ran a built version in my test EKS cluster.

**Automation added to e2e**:

no

no

no, will not break since it's a change to do with the logging interface

no

```release-note
```

## How to use:

For all Reconcile I have added something like this at the beginning of the method
```
	ctx = gwlog.StartReconcileTrace(ctx, c.log, "iamauthpolicy", req.Name)
	defer func() {
		gwlog.EndReconcileTrace(ctx, c.log)
	}()
```
this should produce a log line like so:
```
{"level":"info","ts":"2024-06-07T06:27:53.311Z","logger":"controller.iam-auth-policy","caller":"gwlog/metadata.go:79","msg":"[ACTION_RECONCILE_START]","type":"iamauthpolicy","name":"foo-test","trace_id":"3442054d-0f06-46ac-84bb-7e0112cfa60d"}
{"level":"info","ts":"2024-06-07T06:27:53.311Z","logger":"controller.iam-auth-policy","caller":"controllers/iamauthpolicy_controller.go:92","msg":"reconcile IAM policy","req":"lattice/foo-test","targetRef":{"group":"gateway.networking.k8s.io","kind":"HTTPRoute","name":"foo-test-all
","namespace":"foo-test"},"trace_id":"3442054d-0f06-46ac-84bb-7e0112cfa60d"}
{"level":"info","ts":"2024-06-07T06:27:53.333Z","logger":"controller.iam-auth-policy","caller":"controllers/iamauthpolicy_controller.go:110","msg":"reconciled IAM policy","req":"lattice/foo-test","targetRef":{"group":"gateway.networking.k8s.io","kind":"HTTPRoute","name":"foo-test-a
ll","namespace":"foo-test"},"isDeleted":false,"trace_id":"3442054d-0f06-46ac-84bb-7e0112cfa60d"}
{"level":"info","ts":"2024-06-07T06:27:53.333Z","logger":"controller.iam-auth-policy","caller":"gwlog/metadata.go:85","msg":"[ACTION_RECONCILE_END]","type":"iamauthpolicy","name":"foo-test","trace_id":"3442054d-0f06-46ac-84bb-7e0112cfa60d"}
```

Something to do next, but not as part of this PR as I do not want to conflict with future changes that want to add more log lines is that for errors you can add why a reconcile failed as part of the metadata.

for example in TargetGroupPolicyController you can change this logic to
```
	_, err = c.ph.ValidateAndUpdateCondition(ctx, tgPolicy)
	if err != nil {
		gwlog.AddMetadata(ctx, "failed_reason", err.Error())
		return ctrl.Result{}, err
	}
```

and then the failed_reason will show up in the `[ACTION_RECONCILE_END]` message.

You can still keep using the original logger as it's exposed as InnerLogger in TracedLogger.

### DIFF
--- a/pkg/utils/gwlog/actions.go
+++ b/pkg/utils/gwlog/actions.go
@@ -1,0 +1,4 @@
+package gwlog
+
+const ReconcileStart = "[ACTION_RECONCILE_START]"
+const ReconcileEnd = "[ACTION_RECONCILE_END]"
--- a/pkg/utils/gwlog/actions.go
+++ b/pkg/utils/gwlog/actions.go
@@ -1,4 +1,4 @@
 package gwlog
 
-const ReconcileStart = "[ACTION_RECONCILE_START]"
-const ReconcileEnd = "[ACTION_RECONCILE_END]"
+const ReconcileStart = "RECONCILE_START_MARKER"
+const ReconcileEnd = "RECONCILE_END_MARKER"
--- a/pkg/utils/gwlog/gwlog.go
+++ b/pkg/utils/gwlog/gwlog.go
@@ -14,6 +14,10 @@ type TracedLogger struct {
 	InnerLogger *zap.SugaredLogger
 }
 
+func (s *TracedLogger) Infoln(args ...interface{}) {
+	s.InnerLogger.Infoln(args...)
+}
+
 func (t *TracedLogger) Infow(ctx context.Context, msg string, keysAndValues ...interface{}) {
 	if GetTrace(ctx) != "" {
 		keysAndValues = append(keysAndValues, string(traceID), GetTrace(ctx))
--- a/pkg/utils/gwlog/gwlog.go
+++ b/pkg/utils/gwlog/gwlog.go
@@ -1,6 +1,8 @@
 package gwlog
 
 import (
+	"context"
+	"fmt"
 	"log"
 	"os"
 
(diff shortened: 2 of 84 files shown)


erikfuller:
> you can see that trace 3442054d-0f06-46ac-84bb-7e0112cfa60d is separate from 3442054d-0f06-46ac-84bb-7e0112cfa60d

Just checking this is what you meant as these ids are the same?

shulin-sq:
@erikfuller oops you're correct. I've updated my description. 

I tried looking through my logs to see if I Can find a better example but I have very low traffic in my dev env so I don't see a lot of overlapping actions. However I can confirm that the uuid is correctly generated as unique per action

```
{"level":"info","ts":"2024-06-10T16:36:21.798Z","logger":"controller.gateway","caller":"gwlog/metadata.go:85","msg":"[ACTION_RECONCILE_END]","type":
"gateway","name":"snarl-10-isglf3puxg","trace_id":"b952b248-bf75-49d0-8984-311374ad9012"}
{"level":"info","ts":"2024-06-10T16:36:21.801Z","logger":"controller.gateway","caller":"gwlog/metadata.go:79","msg":"[ACTION_RECONCILE_START]","type
":"gateway","name":"shulilntest","trace_id":"287c9b37-faaf-4429-8b8c-a85dbb778de3"}```

File: pkg/utils/gwlog/metadata.go [hunk 1]
```diff
@@ -0,0 +1,86 @@
...
+		}
+		return ctx.Value(metadata).(*metadataValue).m[traceID]
+	}
+	return ""
+}
+
+func StartReconcileTrace(ctx context.Context, log Logger, k8sresourcetype, name string) context.Context {
+	ctx = NewTrace(ctx)
+	AddMetadata(ctx, "type", k8sresourcetype)
+	AddMetadata(ctx, "name", name)
```
erikfuller, line 77: Let's add the namespace here too

pkg/utils/gwlog/gwlog.go
```diff
@@ -1,14 +1,120 @@
...
+	"fmt"
 	"log"
 	"os"
 
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 )
 
-type Logger = *zap.SugaredLogger
+type TracedLogger struct {
```
erikfuller, line 13: So I was thinking it might be useful to see the log lines not linked to any traces (so trace_id=nil). We could use this to drive down log lines without the additional tracking. This would also simplify the conditionals below, so we would basically always append trace_id and whatever the value for it, even if it's missing. What do you think?

Alternatively, let's save the value from `GetTrace()` each time so we don't have to calcluate it twice.

pkg/utils/gwlog/metadata.go
```diff
@@ -0,0 +1,86 @@
...
+
+	log.Infow(ctx, ReconcileStart, GetMetadata(ctx)...)
```
erikfuller, line 79: Rather than a special method to exclude the trace id from metadata, what about trying to cast to a `TracedLogger`, then either calling a method that logs all metadata, including the trace id, or logging directly against the `InnerLogger`. If it's not a `TracedLogger`, then we can just log all metadata.

erikfuller (changes requested):
Thanks for the PR @shulin-sq, being able to better correlate logs will really help with debugging. I have a few comments and requests, please just let me know if anything is unclear and feel free to push back if you think I've got something wrong.

Note there are also a few conflicts after merging in the tls-passthrough PR, please resolve those as well. There shouldn't be any further PRs coming through before this one so this should be a one-time conflict resolution.

pkg/utils/gwlog/actions.go
```diff
@@ -0,0 +1,4 @@
+package gwlog
+
+const ReconcileStart = "[ACTION_RECONCILE_START]"
+const ReconcileEnd = "[ACTION_RECONCILE_END]"
```
erikfuller, line 4: Just thinking this one through. For "easy" querying, should these just be `"RECONCILE_START"` and `"RECONCILE_END"`? We don't have those strings anywhere else in the code, so should be OK and then wouldn't potentially need escaping for the square brackets.

pkg/utils/gwlog/gwlog.go
shulin-sq: I think it depends on what logging tooling you are using. I think in our case we can do something like `-trace_id` to grab all log lines without the trace_id value

I can add it if you think it's useful to others

pkg/utils/gwlog/metadata.go
```diff
@@ -0,0 +1,86 @@
...
+	newCtx := context.WithValue(ctx, metadata, newMetadata())
+	AddMetadata(newCtx, traceID, currID.String())
+
+	return newCtx
+}
+
+func AddMetadata(ctx context.Context, key, value string) {
+	if ctx.Value(metadata) != nil {
+		ctx.Value(metadata).(*metadataValue).set(key, value)
+	}
```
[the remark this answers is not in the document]
shulin-sq: ack, I think doing nothing is fine but I might want to make it explicit

there's a conversation around if we want to add metadata even when a new trace is not started. I will follow up on that in a separate comment.

pkg/utils/gwlog/metadata.go
shulin-sq: TracedLogger and Innerlogger are not interchangable. They do not implement the same interface since I added a context parameter to the methods of the original logger. That's unfortunately why I had to address every usage of the logger.

I think that would be quite cool but I may be missing how this might work. Maybe you can give an example.

pkg/utils/gwlog/metadata_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+func TestMetadata(t *testing.T) {
+	ctx := NewTrace(context.TODO())
+	AddMetadata(ctx, "foo", "bar")
+
+	md := GetMetadata(ctx)
+	mdMap := map[string]bool{}
+	for _, m := range md {
+		mdMap[fmt.Sprint(m)] = true
+	}
```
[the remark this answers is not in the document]
shulin-sq: md here is a list since that's the input type accepted by the zap logger

test/go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/aws/aws-application-networking-k8s/test
 
-go 1.20
+go 1.21
+
+toolchain go1.22.1
```
shulin-sq, line 6: I'll remove this, this is something that got added by my IDE

pkg/utils/gwlog/metadata.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func GetMetadata(ctx context.Context) []interface{} {
```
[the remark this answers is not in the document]
shulin-sq: I was flip flopping around keeping this. My reasoning to keep it around is to allow the developer to choose to print out the metadata at any time instead of just at the end of the trace. But maybe I can keep that option open for the future and whoever needs it can make this a public method.

I think it still might be a nice helper, but maybe can be private.

pkg/utils/gwlog/gwlog.go
erikfuller: I think it's fine, let's just go with saving the value for now.

pkg/utils/gwlog/metadata_test.go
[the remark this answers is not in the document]
erikfuller: ooooh right. Sorry I missed we were converting from the map to a list. Thanks for highlighting.

pkg/utils/gwlog/actions.go
shulin-sq: adding _MARKER as a suffix selfishly as our internal logging tool is case insensitive. just in case someone wants to log reconcile_start etc.

pkg/utils/gwlog/metadata.go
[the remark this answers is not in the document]
shulin-sq: I looked at this again and I think that the behavior here is correct. I imagine the developer would confirm that their metadata was correctly added in a particular log line.

Alternative is to add a log line or exit here to denote that the developer did something unintended, but seems rather aggressive.

shulin-sq:
@erikfuller I updated the PR, please let me know if there are further requested modifications

pkg/utils/gwlog/metadata.go
```diff
@@ -0,0 +1,85 @@
...
+	return newCtx
+}
+
+func AddMetadata(ctx context.Context, key, value string) {
+	if ctx.Value(metadataKey) != nil {
+		ctx.Value(metadataKey).(*metadata).set(key, value)
+	}
+}
+
+func getMetadata(ctx context.Context) []interface{} {
```
erikfuller, line 45: Just leaving a thought for later consideration. I think rather than computing this dynamically, it might make more sense from a performance perspective to simply keep a copy of this ready whenever a mutation is made to the underlying map.

pkg/utils/gwlog/metadata.go
erikfuller: So here I was thinking of something like
```
traceLogger, ok := log.(TraceLogger)
if ok {
  // here, getMetadata would include everything including trace ID
  traceLogger.InnerLogger.Infow(ReconcileStart, getMetadata(ctx))
}
```
but this was because I missed we were constructing a list rather than a map without the trace id. So I don't think my idea really saves anything other than excluding the trace id from the metadata list.

pkg/utils/gwlog/metadata.go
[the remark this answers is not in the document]
erikfuller: Thanks for having a look. I suppose most of the unit tests would also be sending through a placeholder context so it would get pretty chatty if we warned here.

erikfuller (approved):
Thanks again for putting this together. Look forward to seeing this in action!

PR #652, opened 2024-06-14
Release artifacts for v1.0.6

zijun726911:
Release artifacts for v1.0.6

### DIFF
--- a/files/controller-installation/deploy-v1.0.6.yaml
+++ b/files/controller-installation/deploy-v1.0.6.yaml
@@ -6947,7 +6947,7 @@ spec:
                     type: string
                   protocolVersion:
                     description: The protocol version used when performing health
-                      checks on targets. Defaults to HTTP/1.
+                      checks on targets.
                     enum:
                     - HTTP1
                     - HTTP2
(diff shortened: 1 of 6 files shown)


files/controller-installation/deploy-v1.0.6.yaml
```diff
@@ -7861,9 +8828,55 @@ spec:
...
+          readOnly: true
       securityContext:
         runAsNonRoot: true
       serviceAccountName: gateway-api-controller
       terminationGracePeriodSeconds: 10
+      volumes:
+      - name: webhook-cert
+        secret:
+          defaultMode: 420
+          secretName: webhook-cert
```
erikfuller, line 8845: Why is all the webhook stuff only showing up now? Did we miss it before somehow?

erikfuller: Discussed offline. This is expected since we baseline this diff off the v1.0.1.yaml. In future, we should consider iterating on the prev revision instead to reduce the size of this diff.

erikfuller (approved):

Issue #671, opened 2024-10-25
Add configurable concurrency values


erikfuller:
Today, the different sub-controllers all use the default value for `MaxConcurrentReconciles`, which is 1. For Some controllers this is fine, but because most changes get reconciled as routes, we end up with a backlog of routes whenever larger changes occur in the cluster.

We need to increase the default concurrency for routes and evaluate whether we should add concurrency values for other sub-controllers. Be sure to add the new value to the Helm chart.

PR #672, opened 2024-10-30
Add configurable max workers for route controller instances

erikfuller:
**What type of PR is this?**
Feature/Improvement

#671 

Adds a configurable value for the number of max workers for route controller instances. Note this worker count applies to each route type: HTTP, GRPC, and TLS. This change also updates locking mechanism for target group garbage collection to be more performant and allow parallel execution for route reconciliation.

Created 50 HTTP routes. On start or change, all 50 are reconciled. Previously with max workers = 1 (default), the reconciliation rate was 1/s and total time was roughly 50s. For 5 workers, the rate was closer to 5/s, and total time was around 12s. When set to very large numbers (e.g. 25 or 100), the controller encounters throttling from the Lattice APIs. At the default API rate limits, the controller maxes out between 5-10/s and handles the throttles fine enough. If customers need higher reconciliation rates, they can request API limit increases.

I also checked the Helm install, including a `--dry-run` to verify the environment variable as well as a deployment. When run correctly, the max worker count will show in the logs.

```
# example log line when worker count set to 10
2024-10-29T16:37:47.088-0700	INFO	runtime	controller/controller.go:234	Starting workers	{"controller": "grpcroute", "controllerGroup": "gateway.networking.k8s.io", "controllerKind": "GRPCRoute", "worker count": 10}
```

This is how I was eyeballing the reconciliation rate and found the issue with the target group GC lock.
```
# pull the metrics for the HTTPRoute queue depth
while true; do curl -s http://127.0.0.1:8080/metrics | grep 'workqueue_depth{name="httproute"}' | awk -v d="$(date -j +%H:%M:%S)"  '{ print d, $2 }'; sleep 1; done
```

Also validated the controller errors out when an invalid value is supplied.

**Automation added to e2e**:
n/a

No

No. Default for workers is still 1 if the value unspecified.

Yes, new optional/opt-in feature.

```release-note
Allow configuring concurrency for route reconciliation via ROUTE_MAX_CONCURRENT_RECONCILES environment variable.
```

### DIFF
--- a/pkg/deploy/stack_deployer.go
+++ b/pkg/deploy/stack_deployer.go
@@ -81,7 +81,7 @@ func NewLatticeServiceStackDeploy(
 		tgGcSynth := lattice.NewTargetGroupSynthesizer(log, cloud, k8sClient, tgMgr, tgSvcExpBuilder, svcBuilder, nil)
 		tgGcFn := NewTgGcFn(tgGcSynth)
 		tgGc = &TgGc{
-			lock:    sync.Mutex{},
+			lock:    sync.RWMutex{},
 			log:     log.Named("tg-gc"),
 			ctx:     context.TODO(),
 			isDone:  atomic.Bool{},
@@ -130,7 +130,7 @@ func NewTgGcFn(tgSynth *lattice.TargetGroupSynthesizer) TgGcCycleFn {
 }
 
 type TgGc struct {
-	lock    sync.Mutex
+	lock    sync.RWMutex
 	log     gwlog.Logger
 	ctx     context.Context
 	isDone  atomic.Bool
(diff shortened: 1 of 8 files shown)


pkg/deploy/stack_deployer.go
```diff
@@ -130,7 +130,7 @@ func NewTgGcFn(tgSynth *lattice.TargetGroupSynthesizer) TgGcCycleFn {
 }
 
 type TgGc struct {
-	lock    sync.Mutex
+	lock    sync.RWMutex
```
erikfuller, line 133: The intent with this change is to many reconciliations to process in parallel, so long as target group GC isn't happening and vice versa. Previously, all reconciliations were serialized as a result of this lock. Since every `backendRef` gets a unique target group in Lattice, even when the `backendRefs` reference the same Kubernetes service, parallelized route reconciliation should be fine to run in parallel since they're not mutating any of the same resources.

As the logic grows in complexity, we may need to consider something like locks on individual target groups rather than a single "lock them all" lock.

zijun726911 (approved):

PR #684, opened 2024-12-16
Release artifacts for v1.1.0

erikfuller:
Release artifacts for v1.1.0


files/controller-installation/deploy-v1.0.3.yaml
mikestvz, line 1: Do we need to delete this one?

erikfuller: There's logic in the `gitub-release.sh` script that only keeps the last five of these files. I'm not really sure there's ongoing value in keeping these older files around but there isn't an urgent need to remove them either. 

These files will continue to be available via version tags, for example: https://github.com/aws/aws-application-networking-k8s/tree/v1.0.7/files/controller-installation

erikfuller: The lines in the script that remove old YAMLS: https://github.com/aws/aws-application-networking-k8s/blob/main/scripts/github-release.sh#L73-L74

mikestvz (approved):
/LGTM

Issue #701, opened 2025-02-04
Automate Controller Setup


rlymbur:
Current documentation is outdated and requires several manual steps for new developers onboarding.

PR #702, opened 2025-02-04
Automate Controller Setup

rlymbur:
**What type of PR is this?**
Feature & Documentation

#701 

This reduces developer onboarding time and simplifies the setup process. It also includes upgrades to our tooling dependencies.

N/A

`make setup` was ran successfully to setup a new cluster with `make presubmit` passing.

**Automation added to e2e**:

N/A

No

No

```release-note
`make setup` now replaces `make toolchain` to further automate the setup process
```


scripts/setup.sh
```diff
@@ -0,0 +1,197 @@
...
+        export AWS_REGION=$region
+        export CONTROLLER_VERSION=$controller_version
+
+        describe_cluster_output=$( aws eks describe-cluster --name "$CLUSTER_NAME" --output text 2>&1 )
+        if [[ $describe_cluster_output == *"ResourceNotFoundException"* ]]; then
+            echo "Creating cluster with name: $cluster_name"
+
+            eksctl create cluster --name "$CLUSTER_NAME" --region "$AWS_REGION"
+
+            echo "Allowing traffic from VPC Lattice to EKS cluster"
```
mikestvz, line 98: We should add a validation that cluster is created(line 96) before proceeding with any additional steps.

scripts/setup.sh
```diff
@@ -0,0 +1,197 @@
...
+
+        export CLUSTER_NAME=$cluster_name
+        export AWS_REGION=$region
+        export CONTROLLER_VERSION=$controller_version
+
+        describe_cluster_output=$( aws eks describe-cluster --name "$CLUSTER_NAME" --output text 2>&1 )
+        if [[ $describe_cluster_output == *"ResourceNotFoundException"* ]]; then
+            echo "Creating cluster with name: $cluster_name"
+
+            eksctl create cluster --name "$CLUSTER_NAME" --region "$AWS_REGION"
```
mikestvz, line 96: I got a regex error on cluster name , wondering if we advise on naming conventions 
i.e 
`Error: validation for Test-Make_setup-Cluster failed, name must satisfy regular expression pattern: [a-zA-Z][-a-zA-Z0-9]`

scripts/setup.sh
```diff
@@ -0,0 +1,197 @@
...
+
+            echo "EKS cluster created successfully."
+        else
+            echo "Cluster: $cluster_name already exists. Skipping creation."
+        fi
+    fi
+    echo "---------------------------------"
+}
+
+crds() {
```
mikestvz, line 183: would there be a dependency on the create clusted to add the CRDs?

mikestvz (changes requested):

scripts/setup.sh
```diff
@@ -0,0 +1,197 @@
...
+cluster() {
+    read -p "Do you want to create an EKS cluster? (Y/N): " create_cluster
+    if [[ $create_cluster == 'Y' || $create_cluster == 'y' ]]; then
+        read -p "Enter Cluster Name: " cluster_name
+        read -p "Enter AWS Region: " region
+        read -p "Enter Controller Version, the current is $CURRENT_CONTROLLER_VERSION: " controller_version
+
+        export CLUSTER_NAME=$cluster_name
+        export AWS_REGION=$region
+        export CONTROLLER_VERSION=$controller_version
```
mikestvz, line 90: We can default to v.1.1.0 if the user does not provide it.

mikestvz (changes requested):

scripts/setup.sh
rlymbur: It's possible the user may be using a different cluster, not created here.

scripts/setup.sh
rlymbur: Yes, good plan.

scripts/setup.sh
rlymbur: Yes that's a good idea, added!

scripts/setup.sh
rlymbur: I have added this validation now.

mikestvz (approved):
LGTM

Issue #673, opened 2024-11-01
Fix flaky e2e test for Access Log Subscriptions


erikfuller:
This e2e tests appear to be regularly failing on:
```
Access Log Policy creation produces Access Log Subscriptions with Bucket, Log Group, and Delivery Stream destinations on the same targetRef
```

https://github.com/aws/aws-application-networking-k8s/actions/workflows/e2e-test.yaml

Error:
```
Access Log Policy creation produces Access Log Subscriptions with Bucket, Log Group, and Delivery Stream destinations on the same targetRef
/home/runner/work/aws-application-networking-k8s/aws-application-networking-k8s/test/suites/integration/access_log_policy_test.go:398
{"level":"info","ts":"2024-11-01T18:14:08.365Z","caller":"test/framework.go:186","msg":"Creating objects: *v1alpha1.AccessLogPolicy/test-access-log-policy-bucket"}
{"level":"info","ts":"2024-11-01T18:14:08.469Z","caller":"test/framework.go:186","msg":"Creating objects: *v1alpha1.AccessLogPolicy/test-access-log-policy-log-group"}
{"level":"info","ts":"2024-11-01T18:14:08.561Z","caller":"test/framework.go:186","msg":"Creating objects: *v1alpha1.AccessLogPolicy/test-access-log-policy-delivery-stream"}
  [FAILED] in [It] - /home/runner/work/aws-application-networking-k8s/aws-application-networking-k8s/test/suites/integration/access_log_policy_test.go:468 @ 11/01/24 18:17:08.654
{"level":"info","ts":"2024-11-01T18:17:09.729Z","caller":"test/framework.go:264","msg":"Deleting objects: *v1alpha1.AccessLogPolicy/test-access-log-policy-bucket"}
{"level":"info","ts":"2024-11-01T18:17:09.834Z","caller":"test/framework.go:282","msg":"Waiting for NotFound, objects: *v1alpha1.AccessLogPolicy/test-access-log-policy-bucket"}
{"level":"info","ts":"2024-11-01T18:17:11.018Z","caller":"test/framework.go:264","msg":"Deleting objects: *v1alpha1.AccessLogPolicy/test-access-log-policy-delivery-stream"}
{"level":"info","ts":"2024-11-01T18:17:11.118Z","caller":"test/framework.go:282","msg":"Waiting for NotFound, objects: *v1alpha1.AccessLogPolicy/test-access-log-policy-delivery-stream"}
[...]

PR #703, opened 2025-02-05
Correct Failing E2E Tests

rlymbur:
**What type of PR is this?**
Bug

#673

Increases timeout on E2E tests and adds a custom retryer to address [these failures](https://github.com/aws/aws-application-networking-k8s/actions/runs/13082442826/job/36508547103).

```
make e2e-test

...

Ran 69 of 69 Specs in 3036.131 seconds
SUCCESS! -- 69 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestIntegration (3036.14s)
PASS
ok      github.com/aws/aws-application-networking-k8s/test/suites/integration   3036.770s
```

**Automation added to e2e**:

N/A

No.

No.

No.


test/suites/integration/grpcroute_test.go
```diff
@@ -70,7 +70,7 @@ var _ = Describe("GRPCRoute test", Ordered, func() {
 				g.Expect(len(rules)).To(Equal(1))
 				g.Expect(*rules[0].Match.HttpMatch.Method).To(Equal("POST"))
 				g.Expect(*rules[0].Match.HttpMatch.PathMatch.Match.Prefix).To(Equal("/"))
-			}).Within(30 * time.Second).Should(Succeed())
+			}).Within(1 * time.Minute).Should(Succeed())
```
mikestvz, line 73: Any particular reason to increase the time ?

test/suites/integration/access_log_policy_test.go
```diff
@@ -86,6 +86,9 @@ var _ = Describe("Access Log Policy", Ordered, func() {
 	)
 
 	BeforeAll(func() {
+		SetDefaultEventuallyTimeout(5 * time.Minute)
```
mikestvz, line 89: What is the reasoning behind the 5 min timeout?

rlymbur: The default 3 minutes is on the edge now and is not enough to reliably pass each time. Until we make controller optimizations, I'd rather we create a baseline with this passing, and we can decrease it in the future.

test/suites/integration/grpcroute_test.go
rlymbur: Similar to the other timeout, the latency in our reconciliation is too slow for the existing 30 seconds. Ideally we reduce these after making future optimizations, but the tests themselves are still useful now to test overall logic.

test/suites/integration/ram_share_test.go
```diff
@@ -50,7 +51,11 @@ var _ = Describe("RAM Share", Ordered, func() {
 	BeforeAll(func() {
 		secondaryTestRoleArn = os.Getenv("SECONDARY_ACCOUNT_TEST_ROLE_ARN")
 
-		primarySess := session.Must(session.NewSession(&aws.Config{Region: aws.String(config.Region)}))
+		primarySess := session.Must(session.NewSession(&aws.Config{
+			Region:  aws.String(config.Region),
+			Retryer: retry.WithMaxRetries(3),
```
erikfuller, line 56: Do we really need our own retryer ("retrier"? whatever) logic? I see a [DefaultRetryer](https://docs.aws.amazon.com/sdk-for-go/api/aws/client/#DefaultRetryer) that looks to be a superset of our custom retryer logic. Should be close enough in any case that we then wouldn't have to maintain our own.

erikfuller (dismissed):
Thanks for digging into this. Really just my one question around the retryer code to address before moving ahead.

test/suites/integration/ram_share_test.go
erikfuller: I'll take "good enough" over "exactly meeting our needs" if it means less code to maintain here.

rlymbur: Looks like the default one will work with these latest changes; executing the failing RAM ones from before.

```
Ran 3 of 69 Specs in 109.419 seconds
SUCCESS! -- 3 Passed | 0 Failed | 0 Pending | 66 Skipped
```

I'm going to create another PR to add [Merge Queue](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue) support as well to prevent us from getting into a failure state again.

Our current E2E action only runs on a push and not any PR. This is by design since we don't want to run any arbitrary code against our test environment. The problem though, is the PR could introduce test failures that would be merged and it would be too late. The merge queue can create an additional step in the middle that allows the E2E tests to execute after approval, but before merge to catch any regression.

test/suites/integration/ram_share_test.go
```diff
@@ -50,7 +51,19 @@ var _ = Describe("RAM Share", Ordered, func() {
 	BeforeAll(func() {
 		secondaryTestRoleArn = os.Getenv("SECONDARY_ACCOUNT_TEST_ROLE_ARN")
 
-		primarySess := session.Must(session.NewSession(&aws.Config{Region: aws.String(config.Region)}))
+		retryer := awsClient.DefaultRetryer{
+			MinThrottleDelay: 1 * time.Second,
+			MinRetryDelay:    1 * time.Second,
+			MaxThrottleDelay: 5 * time.Second,
+			MaxRetryDelay:    5 * time.Second,
+			NumMaxRetries:    3,
```
iph, line 59: What's the default retries we are overriding? 

I get that these are integration tests -- but 3 is "quite a bit".

iph (approved):
LGTM, had a clarifying question around amount of retries but that's about it.

test/suites/integration/ram_share_test.go
iph: There is a middle ground of running integ tests automatically for maintainers, while manually approving for PR intake. Merge queue may be a bit "late in the process".

test/suites/integration/ram_share_test.go
rlymbur: The default retries is actually already three, probably best I remove it here to avoid any confusion. For some more context:
```
const (
	// DefaultRetryerMaxNumRetries sets maximum number of retries
	DefaultRetryerMaxNumRetries = 3

	// DefaultRetryerMinRetryDelay sets minimum retry delay
	DefaultRetryerMinRetryDelay = 30 * time.Millisecond

	// DefaultRetryerMinThrottleDelay sets minimum delay when throttled
	DefaultRetryerMinThrottleDelay = 500 * time.Millisecond

	// DefaultRetryerMaxRetryDelay sets maximum retry delay
	DefaultRetryerMaxRetryDelay = 300 * time.Second

	// DefaultRetryerMaxThrottleDelay sets maximum delay when throttled
	DefaultRetryerMaxThrottleDelay = 300 * time.Second
)
```

mikestvz (approved):
LGTM  🚀

PR #790, opened 2025-08-11
Correct additional target group creation when using the exportedPorts with HTTPRoute

rlymbur:
**What type of PR is this?**
Bug

Fixes issue where multiple target groups are continually created for HTTPRoute with ServiceExport using `exportedPorts`

This PR fixes a bug in the ServiceExport target group creation logic where the `K8SProtocolVersion` field was being inconsistently overwritten with the route type string (e.g., "HTTP") instead of maintaining the actual protocol version (e.g., "HTTP1"). This inconsistency caused target group lookups by tags to fail during reconciliation, leading to duplicate target groups being created instead of reusing existing ones.

The fix ensures that `K8SProtocolVersion` remains consistent with the actual VPC Lattice protocol version, which is used for target group identification and lookup via AWS resource tags.

Repro steps:
1. Create a ServiceExport with exportedPorts specifying RouteType: "HTTP"
2. Create an HTTPRoute that references a ServiceImport backed by the ServiceExport
3. Observe multiple target groups being created continuously in AWS VPC Lattice console
4. Check controller logs for repeated target group creation events

The issue occurs because the target group spec hash changes between reconciliations due to inconsistent `K8SProtocolVersion` values, causing the controller to think it needs to create new target groups.

- Added integration test `httproute_serviceexport_test.go` that mirrors the existing gRPC equivalent
- Verified that only one target group is created for HTTP routes with ServiceExport using exportedPorts
- Confirmed that target group lookup by tags works correctly with consistent protocol version
- Tested that the fix doesn't affect existing functionality for legacy ServiceExport behavior

**Automation added to e2e**:
Yes - Added `test/suites/integration/httproute_serviceexport_test.go` which tests HTTPRoute with ServiceExport using exportedPorts, mirroring the existing gRPC test pattern.

No

No breaking changes. This fix only affects the internal target group spec generation and doesn't change any external APIs or behavior. Existing target groups will continue to work, and the fix prevents creation of duplicate target groups going forward.

No user-facing changes. This is an internal bug fix that resolves duplicate target group creation.

```release-note
Fixed issue where HTTPRoute with ServiceExport using exportedPorts would create multiple target groups continuously
```

Yes, all tests pass including the new HTTPRoute ServiceExport integration test.

### DIFF
--- a/pkg/gateway/model_build_targetgroup.go
+++ b/pkg/gateway/model_build_targetgroup.go
@@ -225,12 +225,6 @@ func (t *svcExportTargetGroupModelBuildTask) buildTargetGroupForExportedPort(ctx
 	spec.K8SServiceNamespace = t.serviceExport.Namespace
 	spec.K8SProtocolVersion = protocolVersion
 
-	// Add a tag for the route type to help with identification
-	// This is not used by the controller but can be helpful for debugging
-	if exportedPort.RouteType != "" {
-		spec.K8SProtocolVersion = exportedPort.RouteType
-	}
-
 	stackTG, err := model.NewTargetGroup(t.stack, spec)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


mikestvz (approved):
LGTM

PR #878, opened 2026-01-22
Remove port configuration from health checks

aristosvo:
Removed the port configuration from health check settings to make the example work again. We can either remove it or keep it around, but it should be port 8090 in that case

**What type of PR is this?**
documentation

The health check doesn't work

Update the health check by removing the port to make it target the workload port (which works)

Follow the multi-cluster guide

Screenshot 2026-01-22 at 11 48 07

Screenshot 2026-01-22 at 11 48 22

**Automation added to e2e**:
n/a

no

n/a

no

```release-note

```

n/a

### DIFF
--- a/docs/guides/getstarted.md
+++ b/docs/guides/getstarted.md
@@ -329,6 +329,7 @@ spec:
     healthyThresholdCount: 2
     unhealthyThresholdCount: 3
     path: "/health"
+    port: 8090
     protocol: HTTP
     protocolVersion: HTTP1
     statusMatch: "200-299"


docs/guides/getstarted.md
```diff
@@ -329,7 +328,6 @@ spec:
     healthyThresholdCount: 2
     unhealthyThresholdCount: 3
     path: "/health"
```
aristosvo:
```suggestion
    path: "/health"
    port: 8090
```

aristosvo:
Consider it done ✅

PR #884, opened 2026-02-09
fix: set ResolvedRefs condition on Gateway listener status

abdallahmahran10:
## Problem

Gateway API v1.0+ conformance tests require a `ResolvedRefs` condition on each listener status. The controller only sets the `Accepted` condition, causing multiple conformance test failures:

- GatewayWithAttachedRoutes (2/3 subtests)
- HTTPRouteSimpleSameNamespace
- HTTPRouteCrossNamespace
- HTTPRouteExactPathMatching (6/6 subtests)

Error: `expected condition 'ResolvedRefs' to be 'True' but got: (condition not found)`

## Solution

For valid listeners, add a `ResolvedRefs: True` condition alongside the existing `Accepted` condition. Setting `True` is correct because:
- **TLS CertificateRefs**: VPC Lattice uses ACM for TLS, not K8s Secrets — not applicable
- **AllowedRoutes.Kinds**: Already validated by `listenerRouteGroupKindSupported()`
- **Cross-namespace refs**: No cross-namespace TLS refs — not applicable

For invalid listeners (unsupported route kinds), `ResolvedRefs: False` was already set.

## Changes

- **`pkg/controllers/gateway_controller.go`**: Added `ResolvedRefs` condition for valid listeners in `UpdateGWListenerStatus()`
- **`pkg/controllers/gateway_controller_test.go`**: Added table-driven test verifying both conditions are present with correct values

## Testing

```
go test ./... # All 24 packages pass
```

## Dependencies

This PR depends on #883 (stale listener cleanup).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
-# AWS Gateway API Controller for VPC Lattice
+# Amazon VPC Lattice Gateway API Controller
 
 <p align="center">
     <img src="docs/images/kubernetes_icon.svg" alt="Kubernetes logo" width="100" /> 
     <img src="docs/images/controller.png" alt="AWS Load Balancer logo" width="100" />
 </p>
 
-AWS Application Networking is an implementation of the Kubernetes [Gateway API](https://gateway-api.sigs.k8s.io/). This project is designed to run in a Kubernetes cluster and orchestrates AWS VPC Lattice resources using Kubernetes Custom Resource Definitions like Gateway and HTTPRoute.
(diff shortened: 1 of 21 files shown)


SinghVikram97 (commented):
Changes lgtm. Do we also need to update route and gateway controller to call `UpdateGWListenerStatus` function during both Upsert and Delete? Or is that a bigger change and not in scope of this PR?

abdallahmahran10:
> Changes lgtm. Do we also need to update route and gateway controller to call `UpdateGWListenerStatus` function during both Upsert and Delete? Or is that a bigger change and not in scope of this PR?

Yes we should, I was planning to do that as part of another PR.

Issue #888, opened 2026-02-11
Allow creation of ServiceNetworks in code


starlightromero:
**Is your feature request related to a problem? Please describe.**
The Helm chart and deployment allow for setting a default service network. However, if you want to create multiple gateways (service networks and gateways have a 1:1 relationship), you need to use some other tool to manage them (Terraform, Crossplane, CDK) or create them in a non-IaC way via the console or cli.

**Describe the solution you'd like**
There should be a ServiceNetwork CRD that allows the creation of ServiceNetworks via custom resources. 

Alternatively, a service network should automatically be created when a gateway is created. Then the Gateway resource would manage the service network in true GitOps fashion.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

PR #936, opened 2026-04-10
feat: Add ServiceNetwork CRD for VPC Lattice service network lifecycle management

SinghVikram97:
**What type of PR is this?**

feature

https://github.com/aws/aws-application-networking-k8s/issues/888

Adds a cluster scoped ServiceNetwork CRD that provides Kubernetes native lifecycle management for VPC Lattice service networks. Currently, service networks can only be managed via the DEFAULT_SERVICE_NETWORK env var or externally via AWS CLI/console. This CRD allows users to create and delete service networks declaratively through Kubernetes resources.

- `make presubmit` passes 
- 15 unit tests added 
- 4 e2e integration tests added

**Automation added to e2e**:
Yes, added `test/suites/integration/servicenetwork_test.go` with 4 test cases covering create, adopt, and delete blocked scenarios. Added ServiceNetwork to TestObjects in test framework for cleanup tracking.

No. Uses existing VPC Lattice SDK calls (CreateServiceNetwork, DeleteServiceNetwork, FindServiceNetwork) already available in the codebase.

No breaking changes. The CRD is additive. Gateway continues to work with or without a ServiceNetwork CR. DEFAULT_SERVICE_NETWORK continues to work unchanged. The controller gracefully handles the CRD not being installed (optional CRD registration).

Yes, a new CRD is available for users to manage service network lifecycle.

**release-note**
```
Added ServiceNetwork CRD for Kubernetes native management of VPC Lattice service
networks. Users can now create, adopt, and delete service networks declaratively
via custom resources as an alternative to `DEFAULT_SERVICE_NETWORK`.
```

```
FOCUS="ServiceNetwork CRD" make e2e-test

Ran 4 of 148 Specs in 128.226 seconds
SUCCESS! -- 4 Passed | 0 Failed | 0 Pending | 144 Skipped
--- PASS: TestIntegration (128.23s)
```


pkg/deploy/lattice/service_network_manager.go
```diff
@@ -265,6 +272,78 @@ func (m *defaultServiceNetworkManager) CreateOrUpdate(ctx context.Context, servi
...
+		}
+		return err
+	}
+
+	snArn := aws.StringValue(sn.SvcNetwork.Arn)
+	owned, err := m.cloud.IsArnManaged(ctx, snArn)
+	if err != nil {
+		m.log.Warnf(ctx, "Cannot check ownership of ServiceNetwork %s: %s, skipping deletion", snName, err)
+		return nil
+	}
```
liucalvin, line 334: in this case should we retry? otherwise if it turns out to be managed, the custom resource will be deleted but not deleted from lattice

pkg/deploy/lattice/service_network_manager.go
```diff
@@ -265,6 +272,78 @@ func (m *defaultServiceNetworkManager) CreateOrUpdate(ctx context.Context, servi
 	return model.ServiceNetworkStatus{ServiceNetworkARN: serviceNetworkArn, ServiceNetworkID: serviceNetworkId}, nil
 }
 
+func (m *defaultServiceNetworkManager) Upsert(ctx context.Context, name string, tags services.Tags) (model.ServiceNetworkStatus, error) {
```
liucalvin: do we handle updating tags?

pkg/deploy/lattice/service_network_manager_test.go
```diff
@@ -822,3 +822,234 @@ func Test_UpsertVpcAssociation_WithAdditionalTags_NoExistingAssociation(t *testi
...
+	defer c.Finish()
+	ctx := context.TODO()
+	mockLattice := mocks.NewMockLattice(c)
+	cloud := pkg_aws.NewDefaultCloud(mockLattice, TestCloudConfig)
+
+	name := "test-sn"
+	arn := "arn:aws:vpc-lattice:us-west-2:123456789:servicenetwork/sn-123"
+	id := "sn-123"
+
+	mockLattice.EXPECT().FindServiceNetwork(gomock.Any(), name).Return(nil, nil)
```
liucalvin: should FindServiceNetwork return `notFoundError`?

liucalvin (commented):
left a few comments, also some questions:
- with this implementation do we guard against a user creating a ServiceNetwork CR with the same name as the DEFAULT_SERVICE_NETWORK?
- do we need to add permissions for this CRD to the ClusterRole? 
- does config/crds/kustomization.yaml need to include the new CRD?

SinghVikram97:
> left a few comments, also some questions:
> 
> * with this implementation do we guard against a user creating a ServiceNetwork CR with the same name as the DEFAULT_SERVICE_NETWORK?

No guard, but if user creates a service network CR with same name as `DEFAULT_SERVICE_NETWORK` then it will adopt / own it as ManagedBy tag will match.

> * do we need to add permissions for this CRD to the ClusterRole?
> * does config/crds/kustomization.yaml need to include the new CRD?

Good catch, will update them.

pkg/deploy/lattice/service_network_manager.go
SinghVikram97: Yes, makes sense, will update

pkg/deploy/lattice/service_network_manager.go
SinghVikram97: we should I will add handling for it

pkg/controllers/servicenetwork_controller.go
```diff
@@ -0,0 +1,190 @@
...
+func (r *serviceNetworkReconciler) reconcileUpsert(ctx context.Context, sn *anv1alpha1.ServiceNetwork) error {
+	if err := r.finalizerManager.AddFinalizers(ctx, sn, serviceNetworkFinalizer); err != nil {
+		return err
+	}
+
+	additionalTags := k8s.GetAdditionalTagsFromAnnotations(ctx, sn)
+
+	status, err := r.snManager.Upsert(ctx, sn.Name, additionalTags)
+	if err != nil {
+		r.updateStatus(ctx, sn, metav1.ConditionFalse, "ReconcileError", err.Error(), "", "")
```
liucalvin: should we pass the `sn.Status.ServiceNetworkARN` and `sn.Status.ServiceNetworkID` to the updateStatus call like we do for Delete? so the status is still visible

liucalvin (approved):
one small comment, otherwise lgtm!

pkg/controllers/servicenetwork_controller.go
SinghVikram97: Yes, good point, fixed

PR #938, opened 2026-04-14
feat: add periodic reconciliation for drift detection

pscheri:
Introduce `RECONCILE_DEFAULT_RESYNC_SECONDS` env var that enables periodic re-reconciliation across all controllers. When set, successful reconciliations schedule a requeue after the configured interval, allowing the controller to detect and recover from out-of-band changes to VPC Lattice resources.

Default is 0 (disabled), preserving current behavior. Minimum value is 60 seconds to avoid excessive Lattice API calls.

**What type of PR is this?**
feature

N/A — new feature. Addresses the lack of drift detection for VPC Lattice resources modified or deleted out-of-band.

Currently, reconciliation only triggers on K8s object events. If a Lattice resource is deleted or modified via the AWS Console/CLI, the controller never notices because:
- `SyncPeriod` is not set (defaults to 10h)
- Even when resync fires, predicates compare old vs new K8s state — since nothing changed in K8s, they return false

This PR adds a `RECONCILE_DEFAULT_RESYNC_SECONDS` env var. When set, `HandleReconcileError` returns `RequeueAfter` on success, bypassing predicates entirely. All controllers using `HandleReconcileError` benefit from drift detection (routes, gateways, services, service exports, access log policies).

Note: `iamauthpolicy` and `vpcassociationpolicy` controllers don't use `HandleReconcileError`, so they are not covered. This can be addressed in a follow-up.

Changes:
- `pkg/config/controller_config.go`: Add `RECONCILE_DEFAULT_RESYNC_SECONDS` env var, parse to `time.Duration`, validate minimum 60s
- `pkg/runtime/reconcile.go`: Return `RequeueAfter` on successful reconciliation when interval > 0
- `pkg/runtime/reconcile_test.go`: Unit tests for the requeue behavior
- `docs/guides/environment.md`: Document the new env var
- `helm/values.yaml`: Add `reconcileDefaultResyncSeconds` value
- `helm/templates/deployment.yaml`: Wire env var into container spec
- `config/manager/manager.yaml`: Add env var for non-Helm deployments
- `test/suites/integration/drift_detection_test.go`: 4 e2e test cases

- `make presubmit` passes (vet, manifest, lint, test)
- e2e drift detection tests run against EKS cluster in us-east-2 with `RECONCILE_DEFAULT_RESYNC_SECONDS=60`
- Manual drift test: created HTTPRoute, deleted Lattice service out-of-band via console, controller detected and recreated it within 60 seconds

**Automation added to e2e**:
Added `test/suites/integration/drift_detection_test.go` with 4 test cases:
1. Service deletion → verifies recreation with new resource ID
2. Listener deletion → verifies recreation with new resource ID
3. Rule deletion → verifies recreation with new resource ID
4. Rule action modification → verifies revert from FixedResponse back to original Forward config

Tests require `RECONCILE_DEFAULT_RESYNC_SECONDS` to be set on both the controller and test runner. When not set (default), the tests fail fast with a clear error message rather than timing out.

Run with:
RECONCILE_DEFAULT_RESYNC_SECONDS=60 FOCUS="Drift detection" REGION=us-east-2 make e2e-test

No. No new API calls — the reconciler already calls all Lattice APIs on every run. This just controls how often that run happens.

No. Default is 0 (disabled), preserving current behavior exactly. Existing deployments are unaffected unless the env var is explicitly set.

Add `RECONCILE_DEFAULT_RESYNC_SECONDS` environment variable to enable periodic re-reconciliation for drift detection. When set to a positive integer (minimum 60 seconds), the controller will periodically re-check VPC Lattice resources and restore any that were modified or deleted out-of-band. Default is 0 (disabled).

Resource creation/reconciliation tests pass. Traffic verification tests time out due to running the controller locally (Lattice data plane not reachable from test pods) — this is a pre-existing limitation of the local e2e setup, not related to this change.
Drift detection tests (4/4 passed):
Ran 4 of 148 Specs in 228.106 seconds
SUCCESS! -- 4 Passed | 0 Failed | 0 Pending | 144 Skipped
--- PASS: TestIntegration (228.11s)
PASS


pkg/config/controller_config.go
```diff
@@ -107,6 +109,15 @@ func configInit(sess *session.Session, metadata EC2Metadata) error {
 		RouteMaxConcurrentReconciles = routeMaxConcurrentReconcilesInt
 	}
 
+	routeReconcileInterval := os.Getenv(ROUTE_RECONCILE_INTERVAL)
+	if routeReconcileInterval != "" {
+		routeReconcileIntervalInt, err := strconv.Atoi(routeReconcileInterval)
+		if err != nil {
+			return fmt.Errorf("invalid value for ROUTE_RECONCILE_INTERVAL: %s", err)
+		}
+		RouteReconcileInterval = routeReconcileIntervalInt
```
SinghVikram97: We can convert to time.Duration here and store it in first pass

SinghVikram97: We can add validation here for it's minimum possible value

pkg/config/controller_config.go
```diff
@@ -43,6 +44,7 @@ var WebhookEnabled = ""
 var DisableTaggingServiceAPI = false
 var ServiceNetworkOverrideMode = false
 var RouteMaxConcurrentReconciles = 1
+var RouteReconcileInterval = 0 // seconds; 0 = disabled (current behavior)
```
SinghVikram97: we can remove the default from here

pkg/runtime/reconcile.go
```diff
@@ -2,13 +2,18 @@ package runtime
...
 import (
 	"errors"
+	"time"
 
+	"github.com/aws/aws-application-networking-k8s/pkg/config"
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
 // HandleReconcileError will handle errors from reconcile handlers, which respects runtime errors.
 func HandleReconcileError(err error) (ctrl.Result, error) {
```
SinghVikram97, line 11: `HandleReconcileError` is used by other controllers so this will affect all those controllers too

docs/guides/environment.md
```diff
@@ -112,3 +112,15 @@ The Helm chart sets this value to "false" by default.
 **Default:** 1
 
 Maximum number of concurrently running reconcile loops per route type (HTTP, GRPC, TLS)
+
+---
+
+#### `ROUTE_RECONCILE_INTERVAL`
```
SinghVikram97: - Should we add this for all controllers or just RouteController? Something like `RECONCILE_DEFAULT_RESYNC_SECONDS`  if we are adding for all controllers
- Should we also add a minimum possible value for it 1 minute maybe, to avoid hammering lattice APIs

SinghVikram97 (changes requested):
Thank you for contributing. Left a few review comments. 

We also need to add this env variable to a few other places 
- For Helm users
   - Add in [values.yaml](https://github.com/aws/aws-application-networking-k8s/blob/main/helm/values.yaml)
   - Add in [deployment.yaml](https://github.com/aws/aws-application-networking-k8s/blob/main/helm/templates/deployment.yaml)
-  For Non Helm users
    - Add in [manager.yaml](https://github.com/aws/aws-application-networking-k8s/blob/main/config/manager/manager.yaml#L56)

iamauthpolicy and vpcassociationpolicy controllers don't call HandleReconcileError, so they won't get drift detection

pkg/config/controller_config.go
```diff
@@ -107,6 +110,21 @@ func configInit(sess *session.Session, metadata EC2Metadata) error {
...
+	reconcileDefaultResyncSeconds := os.Getenv(RECONCILE_DEFAULT_RESYNC_SECONDS)
+	if reconcileDefaultResyncSeconds != "" {
+		reconcileDefaultResyncSecondsInt, err := strconv.Atoi(reconcileDefaultResyncSeconds)
+		if err != nil {
+			return fmt.Errorf("invalid value for RECONCILE_DEFAULT_RESYNC_SECONDS: %s", err)
+		}
+		// Minimum 60 seconds to avoid excessive Lattice API calls.
+		// Each reconciliation makes multiple API calls (FindService, ListListeners,
+		// ListRules, ListTargetGroups, etc.), so lower values risk throttling.
+		if reconcileDefaultResyncSecondsInt < 60 {
```
liucalvin, line 122: should we also check for `reconcileDefaultResyncSecondsInt != 0`?

pscheri: agree, 0 is a valid option for disabled drift detection and self-healing. Pushing a small change on that

test/suites/integration/drift_detection_test.go
```diff
@@ -0,0 +1,294 @@
...
+		Expect(err).ToNot(HaveOccurred())
+
+		for _, assoc := range assocResp.Items {
+			_, err := testFramework.LatticeClient.DeleteServiceNetworkServiceAssociation(
+				&vpclattice.DeleteServiceNetworkServiceAssociationInput{
+					ServiceNetworkServiceAssociationIdentifier: assoc.Id,
+				})
+			Expect(err).ToNot(HaveOccurred())
+		}
+		if len(assocResp.Items) > 0 {
```
SinghVikram97, line 64: Can we use eventually block here which lists service network service associations and expect them to empty

test/suites/integration/drift_detection_test.go
```diff
@@ -0,0 +1,294 @@
...
+			ServiceIdentifier: svc.Id,
+		})
+		Expect(err).ToNot(HaveOccurred())
+
+		// Wait for drift detection to recreate it with a new ID
+		Eventually(func(g Gomega) {
+			recreated := testFramework.GetVpcLatticeService(ctx, route)
+			g.Expect(recreated).ToNot(BeNil())
+			g.Expect(*recreated.Status).To(Equal(vpclattice.ServiceStatusActive))
+			g.Expect(aws.StringValue(recreated.Id)).ToNot(Equal(originalId))
```
SinghVikram97, line 78: The lattice generated domain name for the service will change when service is re-created, wondering if we should document that

test/suites/integration/drift_detection_test.go
```diff
@@ -0,0 +1,294 @@
...
+			g.Expect(err).ToNot(HaveOccurred())
+			var newNonDefault []*vpclattice.RuleSummary
+			for _, rule := range resp.Items {
+				if !aws.BoolValue(rule.IsDefault) {
+					newNonDefault = append(newNonDefault, rule)
+				}
+			}
+			g.Expect(newNonDefault).To(HaveLen(len(nonDefaultRules)))
+			// The recreated rule should have a different ID
+			for _, rule := range newNonDefault {
```
SinghVikram97, line 190: Can we do
```
g.Expect(newNonDefault).ToNot(BeEmpty())
ids := lo.Map(newNonDefault, func(r *vpclattice.RuleSummary, _ int) string {
    return aws.StringValue(r.Id)
})
g.Expect(ids).ToNot(ContainElement(originalRuleId))
```

test/go.mod
```diff
@@ -12,13 +12,16 @@ require (
 	github.com/onsi/ginkgo/v2 v2.27.1
 	github.com/onsi/gomega v1.39.0
 	github.com/samber/lo v1.52.0
+	github.com/stretchr/testify v1.11.1
```
SinghVikram97, line 15: Why do we need these 3 new dependencies, can u check once:
- github.com/stretchr/testify v1.11.1
- k8s.io/apiextensions-apiserver v0.34.1
- sigs.k8s.io/yaml v1.6.0

test/suites/integration/drift_detection_test.go
```diff
@@ -0,0 +1,294 @@
+package integration
```
SinghVikram97, line 1: The Github workflow which runs e2e-test installs controller [here](https://github.com/aws/aws-application-networking-k8s/blob/main/.github/workflows/validate-merge-queue-e2e-test.yaml#L99) which doesn't have `RECONCILE_DEFAULT_RESYNC_SECONDS`.  My thinking was that we modify the workflow so that it runs drift detection test at the end

So the steps would be 
1. Install the controller with default ie. no resync
2. Run both the make e2e-test and make webhook-e2e-test
3. Upgrade the controller for drift detection, a step after `Run Test` [here](https://github.com/aws/aws-application-networking-k8s/blob/main/.github/workflows/validate-merge-queue-e2e-test.yaml#L127) which does 
     ``` 
      helm upgrade gateway-api-controller ./helm \
       --namespace aws-application-networking-system \
       --set=reconcileDefaultResyncSeconds=60 \
       --reuse-values
      ```
4. We add another step `Run drift detection tests` which runs `make drift-e2e-test`

We add another makefile target [here](https://github.com/aws/aws-application-networking-k8s/blob/main/Makefile#L230) which runs these

5. In order for `make e2e-test` to skip these tests we can modify `Fail("RECONCILE_DEFAULT_RESYNC_SECONDS must be set t` to be Skip so that they don't get run

Let me know if that makes sense

pkg/runtime/reconcile.go
```diff
@@ -3,12 +3,16 @@ package runtime
...
 	if err == nil {
+		if config.ReconcileDefaultResyncSeconds > 0 {
+			return ctrl.Result{RequeueAfter: config.ReconcileDefaultResyncSeconds}, nil
```
SinghVikram97, line 14: Should we add jitter here of 0-20% so that it prevents thundering herd?

SinghVikram97 (changes requested):

test/go.mod
pscheri: My understanding is that those were added automatically when running `go mod tidy`, from `make vet`, and were already imported in `test/conformance/conformance_test.go`. Happy to remove if needed

pkg/runtime/reconcile.go
pscheri: on normal operations there should be some organic spreading on implicit delays. I guess this could potentially be an issue during controller startup. Will still add it for those scenarios

test/suites/integration/drift_detection_test.go
pscheri: Make sense, I think that at some point in the future, if this features get enabled by default, this should change again to be considered across all tests.

docs/guides/environment.md
```diff
@@ -112,3 +112,17 @@ The Helm chart sets this value to "false" by default.
...
+
+---
+
+#### `RECONCILE_DEFAULT_RESYNC_SECONDS`
+
+**Type:** *int (seconds)*
+
+**Default:** 0 (disabled)
+
+When set to a positive value (minimum 60), all controllers will periodically re-reconcile resources at the specified interval in seconds. This enables drift detection: if a VPC Lattice resource (service, listener, rule, target group, service network, etc.) is modified or deleted out-of-band (e.g. via the AWS Console or CLI), the controller will detect the discrepancy and restore the resource to match the desired state defined in Kubernetes.
```
SinghVikram97: > if a VPC Lattice resource (service, listener, rule, target group, service network, etc.)

nit: Service network won't be reconciled as correctly mentioned in `drift-detection.md`.

pkg/config/controller_config.go
```diff
@@ -107,6 +110,22 @@ func configInit(sess *session.Session, metadata EC2Metadata) error {
...
+			return fmt.Errorf("invalid value for RECONCILE_DEFAULT_RESYNC_SECONDS: %s", err)
+		}
+		// 0 explicitly disables periodic reconciliation.
+		// Otherwise, minimum 60 seconds to avoid excessive Lattice API calls.
+		// Each reconciliation makes multiple API calls (FindService, ListListeners,
+		// ListRules, ListTargetGroups, etc.), so lower values risk throttling.
+		if reconcileDefaultResyncSecondsInt != 0 && reconcileDefaultResyncSecondsInt < 60 {
+			return fmt.Errorf("RECONCILE_DEFAULT_RESYNC_SECONDS must be 0 (disabled) or at least 60 seconds, got %d", reconcileDefaultResyncSecondsInt)
+		}
+		ReconcileDefaultResyncSeconds = time.Duration(reconcileDefaultResyncSecondsInt) * time.Second
```
SinghVikram97: nit: Can we log the value after config parsing [here](https://github.com/aws/aws-application-networking-k8s/blob/main/cmd/aws-application-networking-k8s/main.go#L172)

SinghVikram97 (approved):

PR #949, opened 2026-04-30
add user agent for all sdk calls

SinghVikram97:
**What type of PR is this?**
feature

N/A

Adds a custom user agent suffic (`amazon-vpc-lattice-gateway-api-controller`) to all AWS API calls made by the controller. This allows identifying traffic originating from this controller for debugging and metrics purposes.

N/A

Unit test added and passing (`TestAddUserAgentHandler`)

**Automation added to e2e**:
N/A

No

No breaking changes.

No.

release-note
NONE

N/A, no e2e tests impacted.

### DIFF
--- a/pkg/aws/cloud.go
+++ b/pkg/aws/cloud.go
@@ -6,10 +6,11 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 
+	"maps"
+
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/vpclattice"
-	"maps"
 
 	"github.com/aws/aws-application-networking-k8s/pkg/aws/metrics"
 	"github.com/aws/aws-application-networking-k8s/pkg/aws/services"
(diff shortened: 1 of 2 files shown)


pkg/aws/cloud.go
```diff
@@ -237,3 +240,8 @@ func (c *defaultCloud) isOwner(managedBy string) bool {
 func getManagedByTag(cfg CloudConfig) string {
 	return fmt.Sprintf("%s/%s/%s", cfg.AccountId, cfg.ClusterName, cfg.VpcId)
 }
+
+// addUserAgentHandler appends the controller identifier to the User-Agent header on all AWS API calls.
+func addUserAgentHandler(sess *session.Session) {
+	sess.Handlers.Build.PushBack(request.MakeAddToUserAgentFreeFormHandler("amazon-vpc-lattice-gateway-api-controller"))
```
liucalvin, line 246: what do you think about adding the controller version here? Would be nice to also have metrics on which version customers are using

liucalvin (approved):

pkg/aws/cloud.go
SinghVikram97: Yes, that would be nice to have. Need to make controller version available as env variable tho. Will make a separate PR for that.

PR #943, opened 2026-04-16
Migrate AWS SDK for Go to v2 

liucalvin:
**What type of PR is this?**

enhancement

**Which issue does this PR fix**:
closes #595

**What does this PR do / Why do we need it**:
AWS go SDK was deprecated last year: https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/

This PR upgrades the sdk to v2

**If an issue # is not available please add repro steps and logs from aws-gateway-controller showing the issue**:
n/a

**Testing done on this change**:
ran unit tests, e2e tests

- verified metrics endpoint still works

**Automation added to e2e**:
n/a

**Will this PR introduce any new dependencies?**:
no

**Will this break upgrades or downgrades. Has updating a running cluster been tested?**:

**Does this PR introduce any user-facing change?**:
no

**Do all end-to-end tests successfully pass when running `make e2e-test`?**:
yes

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #957, opened 2026-05-29
fix: address pre-existing drift-detection issues amplified by periodic reconciliation

pscheri:
**What type of PR is this?**

bug

(Three correctness/efficiency fixes plus a test-stability fix, all stemming from
behaviors that became significant once periodic reconciliation / drift detection
was introduced in #938 and extended in #953.)

N/A — follow-up to #938 and #953. No tracking issue; repro steps below.

When `RECONCILE_DEFAULT_RESYNC_SECONDS` is enabled, every controller using the
shared `HandleReconcileError` re-reconciles on a fixed interval. Several
pre-existing behaviors that were harmless when reconcile only fired on
Kubernetes events become problematic when they fire on every interval. This PR
addresses four of them:

1. **IAMAuthPolicy: silent failure when the underlying Lattice resource is
   missing.** `IAMAuthPolicyController.reconcileUpsert` wrapped `pm.Put` errors
   in `services.IgnoreNotFound`, so a deleted underlying VPC Lattice service or
   service network produced a *successful* reconcile, an `Accepted=True` status
   condition, and an emptied resource-id annotation. The user saw a healthy
   policy that was not actually being applied. The `Put` NotFound path now sets
   `Accepted=False`/`Invalid` with a descriptive message and preserves the
   existing annotation. The `IgnoreNotFound` on the delete paths (idempotent
   delete) is intentional and preserved.

2. **IAMAuthPolicy: unconditional rewrite every reconcile.** `putSn`/`putSvc`
   always called `PutAuthPolicy` + `UpdateService`/`UpdateServiceNetwork`
   regardless of drift, i.e. two mutation calls per policy on every periodic
   pass. The manager now calls `GetAuthPolicy` first and skips both writes when
   `AuthPolicyState` is `Active` and the live policy document is semantically
   equal to the desired one (whitespace/key-order-insensitive JSON comparison;
   conservative — any uncertainty falls back to rewriting).

3. **AccessLogPolicy: unconditional `UpdateAccessLogSubscription` every
   reconcile.** When the source matched, the manager always called
   `UpdateAccessLogSubscription` even if the destination ARN was already
   correct. It now skips that call when the destination is in sync; tag drift is
   still reconciled via the (already diff-aware) `Tagging.UpdateTags`.

4. **Route targets: unconditional `RegisterTargets` every reconcile.**
   `TargetsManager.Update` always re-registered the full desired target set.
   It now skips `RegisterTargets` when every desired target is already
   registered in a non-draining state (a desired-but-draining target is still
   re-registered so it is resurrected). `RegisterTargets` is idempotent
   server-side, so this is a pure call-volume reduction with no change to
   converged state.

Also fixes a flaky e2e cleanup: the `VpcAssociationPolicy Drift detection`
`AfterAll` unconditionally recreated the baseline `test-gateway` VPC association
and 409'd ("VPC has already been associated to a service network") on clusters
where the association already existed. The cleanup is now idempotent.

**If an issue # is not available please add repro steps and logs**:

With `RECONCILE_DEFAULT_RESYNC_SECONDS=60` and an `IAMAuthPolicy` targeting a
Gateway whose VPC Lattice service network is then deleted out-of-band:

- Before: `kubectl get iamauthpolicy` shows `Accepted=True`, the resource-id
  annotation is emptied, and the controller logs a successful reconcile every
  interval while applying nothing.
- After: the policy's `Accepted` condition flips to `False`/`Invalid` with a
  message naming the missing resource, and the annotation is preserved.

For #2/#3/#4, before this change CloudTrail shows a steady stream of
`PutAuthPolicy`/`UpdateService`/`UpdateAccessLogSubscription`/`RegisterTargets`
write events every interval even with no drift; after, those calls only occur
when the live state actually differs.

- `go vet ./...`, `gofmt`, and `go test ./pkg/...` all pass.
- `golangci-lint run` (CI-pinned v2.11.3) passes on all changed files.
- Unit tests added (see below).
- Controller run locally against an EKS cluster (us-east-2) with
  `RECONCILE_DEFAULT_RESYNC_SECONDS=60`; controller logs confirm the new
  diff-before-write behavior live: IAMAuthPolicy reconciles issue `GetAuthPolicy`
  and then skip `PutAuthPolicy`/`UpdateService` when in sync; rule manager logs
  "rule unchanged, no updates required"; VpcAssociationPolicy reconciles without
  redundant `UpdateServiceNetworkVpcAssociation`.
- Focused drift e2e against the live cluster (`--ginkgo.focus="Drift detection"`,
  `RECONCILE_DEFAULT_RESYNC_SECONDS=60`):
  - IAMAuthPolicy Drift detection — auth policy deletion recovery ✅, auth-type
    revert recovery ✅
  - VpcAssociationPolicy Drift detection — SNVA deletion recovery ✅,
    security-group drift recovery ✅ (including the fixed `AfterAll`; focused
    re-run: 2 Passed, 0 Failed)

**Automation added to e2e**:

No new e2e specs. The new manager-level behavior is covered by unit tests, and
the existing drift e2e suite exercises the IAMAuthPolicy and VpcAssociationPolicy
drift-recovery paths these changes sit under. New unit tests:

- `Test_IAMAuthPolicy_LatticeResourceNotFound` — asserts `Accepted=False`/`Invalid`
  and annotation preservation when the Lattice resource is missing.
- `pkg/deploy/lattice/iamauthpolicy_manager_test.go` (new) — `policyDocsEqual`
  cases (identity, whitespace, key order, malformed JSON, empty,
  list-order-significance) and `Put` behavior across in-sync (Service and
  ServiceNetwork), policy-doc drift, auth-type drift, GetAuthPolicy NotFound, and
  GetAuthPolicy other-error.
- Access log subscription manager — in-sync skip, in-sync-with-tag-drift, and
  drifted-destination update cases.
- Targets manager — all-in-sync skip, draining resurrection, and
  extra-live-target deregister while skipping register.

No. No new APIs or imports. One additional read API call per reconcile for
IAMAuthPolicy (`GetAuthPolicy`) and access log subscriptions
(`GetAccessLogSubscription`, already called) in exchange for eliminating
mutation calls when nothing has drifted. Net effect in steady state: fewer
total write calls; reads unchanged or +1.

**Will this break upgrades or downgrades. Has updating a running cluster been
tested?**:

No breaking changes. When `RECONCILE_DEFAULT_RESYNC_SECONDS` is unset (default),
reconcile frequency is unchanged. The IAMAuthPolicy status-condition change is
only observable in the previously-silent failure case (Lattice resource missing),
where the new behavior is strictly more correct. No CRD or API changes.

Yes — an IAMAuthPolicy whose underlying VPC Lattice resource has been deleted
out-of-band now reports `Accepted=False` (reason `Invalid`) instead of silently
reporting `Accepted=True`.

```release-note
IAMAuthPolicy now reports a non-Accepted status condition when its underlying VPC
Lattice resource is missing, instead of reporting Accepted while applying nothing.
Drift-detection reconciles for IAM auth policies, access log subscriptions, and
route targets now skip redundant VPC Lattice mutation calls when the live state
already matches the desired state, reducing API call volume when
RECONCILE_DEFAULT_RESYNC_SECONDS is enabled.
```

Ran the drift-detection suite (the suite relevant to these changes) against a live EKS cluster in us-east-2 with RECONCILE_DEFAULT_RESYNC_SECONDS=60:

Will run 8 of 159 specs
IAMAuthPolicy Drift detection recreates an auth policy deleted out-of-band • [42s]
IAMAuthPolicy Drift detection re-enables AWS_IAM auth type when flipped to NONE • [60s]
VpcAssociationPolicy Drift detection recreates the SNVA deleted out-of-band [Serial] • [129s]
VpcAssociationPolicy Drift detection restores security groups drifted out-of-band [Serial] • [269s]
SUCCESS! (VpcAssociationPolicy focused re-run: 2 Passed | 0 Failed)

### DIFF
--- a/docs/guides/drift-detection.md
+++ b/docs/guides/drift-detection.md
@@ -70,7 +70,7 @@ Cross-account RAM sharing associations tied to the original service will not car
 
 Each periodic reconciliation makes multiple VPC Lattice API calls per resource. For clusters with many routes or policies, shorter intervals will result in higher API call volume. Choose an interval that balances recovery time against API usage for your environment.
 
-Most managers diff the live state against the desired state on each pass and skip mutation calls when nothing has drifted. In the common case where a resource is already in sync, a periodic reconcile makes only read calls (e.g. `GetAuthPolicy`, `GetServiceNetworkVpcAssociation`); the corresponding `Put`/`Update` calls only fire when drift is detected. There are two known exceptions where a mutation fires on every pass regardless of drift: target registration (the route controller calls `RegisterTargets` with the full desired set each time) and access log subscriptions (the access log policy controller calls `UpdateAccessLogSubscription` once the source matches). These are pre-existing patterns and may be tightened in follow-up work.
+Most managers diff the live state against the desired state on each pass and skip mutation calls when nothing has drifted. In the common case where a resource is already in sync, a periodic reconcile makes only read calls (e.g. `GetAuthPolicy`, `GetServiceNetworkVpcAssociation`, `ListTargets`); the corresponding `Put`/`Update`/`Register` calls only fire when drift is detected.
 
 ### Controller coverage
 
--- a/docs/guides/drift-detection.md
+++ b/docs/guides/drift-detection.md
@@ -68,7 +68,9 @@ Cross-account RAM sharing associations tied to the original service will not car
 
 ### API usage
 
-Each periodic reconciliation makes multiple VPC Lattice API calls per resource (FindService, ListListeners, ListRules, ListTargetGroups, etc.). For clusters with many routes, shorter intervals will result in higher API call volume. Choose an interval that balances recovery time against API usage for your environment.
+Each periodic reconciliation makes multiple VPC Lattice API calls per resource. For clusters with many routes or policies, shorter intervals will result in higher API call volume. Choose an interval that balances recovery time against API usage for your environment.
+
+Most managers diff the live state against the desired state on each pass and skip mutation calls when nothing has drifted. In the common case where a resource is already in sync, a periodic reconcile makes only read calls (e.g. `GetAuthPolicy`, `GetServiceNetworkVpcAssociation`); the corresponding `Put`/`Update` calls only fire when drift is detected. There are two known exceptions where a mutation fires on every pass regardless of drift: target registration (the route controller calls `RegisterTargets` with the full desired set each time) and access log subscriptions (the access log policy controller calls `UpdateAccessLogSubscription` once the source matches). These are pre-existing patterns and may be tightened in follow-up work.
 
 ### Controller coverage
 
--- a/pkg/controllers/iamauthpolicy_controller.go
+++ b/pkg/controllers/iamauthpolicy_controller.go
@@ -191,7 +191,21 @@ func (c *IAMAuthPolicyController) reconcileUpsert(ctx context.Context, k8sPolicy
 	}
 	statusPolicy, err := c.pm.Put(ctx, modelPolicy)
 	if err != nil {
-		return services.IgnoreNotFound(err)
+		// The underlying VPC Lattice resource (Service or ServiceNetwork) does
+		// not exist. This can happen if it was deleted out-of-band, or — for
+		// Gateway-targeted policies — if the ServiceNetwork has not yet been
+		// created (the gateway controller does not create ServiceNetworks).
+		// Surface this as a non-Accepted condition rather than silently
+		// succeeding; preserve the existing resource-id annotation so any
+		// previously-tracked resource can still be cleaned up if it returns.
+		if services.IsNotFoundError(err) {
+			msg := fmt.Sprintf("VPC Lattice %s %q not found", modelPolicy.Type, resourceName)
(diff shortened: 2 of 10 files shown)


pscheri:
Rebased onto `main` to pick up #943 (AWS SDK for Go v2 migration), which had landed
since this branch was opened and caused conflicts.

What changed in the rebase:
- All four commits re-ported from SDK v1 to v2 semantics (`*WithContext` → ctx-first
  methods, `[]*vpclattice.X` → `[]types.X`, `aws.StringValue`/`Int64Value` →
  `aws.ToString`/`aws.ToInt32`, string enums → `types.*` enum values, `*int64` ports →
  `*int32`, error type assertions → `errors.As` with smithy/`types` exceptions).
- No functional changes from the pre-rebase version; the v1→v2 port is mechanical.
- Verified `go build ./...` passes at every commit, full unit suite passes, and the
  drift-detection e2e suite passes 8/8 against a live EKS cluster on v2.

Force-pushed (history rewritten by the rebase), so the diff now applies cleanly on top
of current `main`.

PR #968, opened 2026-06-18
feat: extend ServiceImport to reference external target groups by ARN

SinghVikram97:
**What type of PR is this?**
feature

Adds a new `application-networking.k8s.aws/target-group-arn` annotation on`ServiceImport` that references an existing, customer-managed VPC Lattice target group by ARN. Combined with the existing`application-networking.k8s.aws/service-name-override` annotation, a route can weight-split traffic between EKS pods and the external target group, enabling gradual ECS / Lambda / EC2 / ALB to EKS migration using Lattice native weighted routing.

Key behaviors:
  - The external target group is reference-only: the controller never tags, mutates, registers targets on, or deletes it.
  - Malformed-ARN, target-group-not-found, and annotation-conflict failures surface as `ResolvedRefs=False` conditions at build time, before any claimed service is mutated.
  - The claim teardown path is hardened: `TryOwn` / `TryOwnFromTags` take an `isDeleting` flag so an untagged customer service is not auto-adopted and deleted on a route delete.

Added integration tests

No

No

Yes
```release-note 
Add the `application-networking.k8s.aws/target-group-arn` annotation on ServiceImport to reference an existing VPC Lattice target group by ARN. Combined with `service-name-override`, this enables weighted ECS/Lambda/EC2/ALB to EKS migration.
```

**Do all end-to-end tests successfully pass when running make e2e-test?**
Yes

### DIFF
--- a/pkg/gateway/model_build_rule.go
+++ b/pkg/gateway/model_build_rule.go
@@ -5,15 +5,20 @@ import (
 	"errors"
 	"fmt"
 	"strconv"
+	"strings"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apitypes "k8s.io/apimachinery/pkg/types"
 
 	anv1alpha1 "github.com/aws/aws-application-networking-k8s/pkg/apis/applicationnetworking/v1alpha1"
+	"github.com/aws/aws-application-networking-k8s/pkg/aws/services"
+	"github.com/aws/aws-application-networking-k8s/pkg/config"
 	"github.com/aws/aws-application-networking-k8s/pkg/k8s"
 	"github.com/aws/aws-application-networking-k8s/pkg/model/core"
 	"github.com/aws/aws-application-networking-k8s/pkg/utils"
 
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/vpclattice"
 	"github.com/aws/aws-sdk-go-v2/service/vpclattice/types"
 	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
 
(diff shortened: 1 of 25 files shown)


pkg/gateway/model_build_rule.go
```diff
@@ -284,31 +289,68 @@ func (t *latticeServiceModelBuildTask) getTargetGroupsForRuleAction(ctx context.
...
+						return nil, k8s.NewConflictingServiceImportAnnotationsError(
+							fmt.Sprintf("ServiceImport %s/%s sets %s together with a tag-discovery annotation "+
+								"(%s / %s / %s); set exactly one",
+								svcImportName.Namespace, svcImportName.Name, k8s.TargetGroupArnAnnotation,
+								k8s.ExportNameAnnotation, k8s.AwsVpcAnnotation, k8s.AwsEksClusterNameAnnotation))
+					}
+					if _, err := k8s.TargetGroupIdFromArn(tgArn); err != nil {
+						return nil, k8s.NewInvalidExternalTargetGroupError(tgArn, err)
+					}
+					out, rerr := t.lattice.GetTargetGroup(ctx, &vpclattice.GetTargetGroupInput{
```
liucalvin, line 317: should we also check for protocol mismatch between the target group and listener?

pkg/k8s/utils.go
```diff
@@ -531,3 +540,80 @@ func GetExportNameFromServiceImport(serviceImport client.Object) string {
...
+func ServiceImportHasTagDiscoveryAnnotation(annotations map[string]string) bool {
+	for _, key := range []string{ExportNameAnnotation, AwsVpcAnnotation, AwsEksClusterNameAnnotation} {
+		if _, ok := annotations[key]; ok {
+			return true
+		}
+	}
+	return false
+}
+
+type InvalidExternalTargetGroupError struct {
```
liucalvin, line 567: nit: might be worth moving these (as well as existing) error types/functions into a separate `pkg/k8s.errors.go` file for readability

liucalvin (approved):
looks great! just some minor comments

pkg/gateway/model_build_rule.go
SinghVikram97: Since the service and target group both already exist, the listener protocol on the existing service matches the target group. The mismatch can happen if the Gateway listener isn't aligned with the listener the existing service has, and in claim mode the controller would change that listener before Lattice gets a chance to validate. It's documented as things to validate [here](https://github.com/SinghVikram97/aws-application-networking-k8s/blob/67273052100d56fe5c555b6c10785fe112a9f837/docs/guides/migrate-to-eks.md?plain=1#L32).

pkg/k8s/utils.go
SinghVikram97: good point, can do in a followup.

PR #971, opened 2026-06-24
conformance test fixes

### LINKED ISSUE #494: Improved conformance for 500 response error conditions
There are a number of locations in the spec where the implementation should return 500s. In some of those locations, the controller is configuring lattice to return a 404 instead (for example when there are no valid backendRefs). Once the Lattice API supports 500 as a fixed response, the controller should be updated to improve conformance with the spec.

liucalvin:
**What type of PR is this?**
bug

 Fixes 3 failing Gateway API conformance tests (GATEWAY-HTTP Core profile): HTTPRouteCrossNamespace, HTTPRouteExactPathMatching, HTTPRouteInvalidNonExistentBackendRef
  
Closes #494

  Two controller bugs prevented Core conformance from passing:
  
  1. Gateway address never populated (route_controller.go): After creating a Lattice service, updateRouteStatusWithServiceInfo calls FindService which returns nil due to AWS API eventual consistency.
  Previously, this returned nil (success) without setting the DNS annotation, so the gateway address was never populated and no retry was triggered. Fix: populate the fields using the service ARN and DNS from the CreateService response 
  2. Unknown-kind backendRef returns 404 instead of 500 (model_build_rule.go): When a backendRef has an unknown kind (not "Service" or "ServiceImport"), neither code branch executed, leaving the rule
  with no target group ID. Traffic hit the listener's default 404 action instead of the spec-required 500. Fix: route unknown kinds through InvalidBackendRefTgId which triggers the existing 500
  FixedResponse path.  

  Additionally: timeout multiplier increased from 2× to 5× for Lattice propagation timing, and Makefile conformance-test target now passes cluster config to Helm.
  

  Run make conformance-test with Gateway API v1.5.0 — the 3 tests above fail. HTTPRouteCrossNamespace/HTTPRouteExactPathMatching fail because the gateway never gets an address.
  HTTPRouteInvalidNonExistentBackendRef fails because the response is 404 not 500.
  

```
  --- PASS: TestConformance (170.76s)
  Core result: Passed 11, Failed 0, Skipped 22
```

**Automation added to e2e**:

no

no

no

yes

### DIFF
--- a/pkg/controllers/route_controller.go
+++ b/pkg/controllers/route_controller.go
@@ -444,7 +444,12 @@ func (r *routeReconciler) updateRouteStatusWithServiceInfo(ctx context.Context, 
 	}
 
 	if svc == nil {
-		r.log.Infof(ctx, "Service not found for route %s-%s", route.Name(), route.Namespace())
+		// Only requeue if the route has accepted parents (a service should have been created).
+		// Routes with all parents rejected won't have a Lattice service.
+		if !core.HasAllParentRefsRejected(route) {
+			r.log.Infof(ctx, "Service not found for route %s-%s, will retry", route.Name(), route.Namespace())
+			return lattice_runtime.NewRequeueNeededAfter("service DNS not yet available", 5*time.Second)
+		}
 		return nil
 	}
 
--- a/pkg/controllers/route_controller.go
+++ b/pkg/controllers/route_controller.go
@@ -51,6 +51,7 @@ import (
 	"github.com/aws/aws-application-networking-k8s/pkg/gateway"
 	"github.com/aws/aws-application-networking-k8s/pkg/k8s"
 	"github.com/aws/aws-application-networking-k8s/pkg/model/core"
+	latticemodel "github.com/aws/aws-application-networking-k8s/pkg/model/lattice"
 	lattice_runtime "github.com/aws/aws-application-networking-k8s/pkg/runtime"
 	k8sutils "github.com/aws/aws-application-networking-k8s/pkg/utils"
 	"github.com/aws/aws-application-networking-k8s/pkg/utils/gwlog"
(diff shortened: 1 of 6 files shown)


pkg/controllers/route_controller.go
```diff
@@ -413,7 +413,12 @@ func (r *routeReconciler) updateRouteStatusWithServiceInfo(ctx context.Context,
 	}
 
 	if svc == nil {
-		r.log.Infof(ctx, "Service not found for route %s-%s", route.Name(), route.Namespace())
+		// Only requeue if the route has accepted parents (a service should have been created).
+		// Routes with all parents rejected won't have a Lattice service.
+		if !core.HasAllParentRefsRejected(route) {
+			r.log.Infof(ctx, "Service not found for route %s-%s, will retry", route.Name(), route.Namespace())
+			return lattice_runtime.NewRequeueNeededAfter("service DNS not yet available", 5*time.Second)
```
SinghVikram97: Do we have access to DNS from create service response: https://github.com/aws/aws-application-networking-k8s/blob/main/pkg/deploy/lattice/service_manager.go#L78

pkg/gateway/model_build_rule.go
```diff
@@ -367,6 +366,10 @@ func (t *latticeServiceModelBuildTask) getTargetGroupsForRuleAction(ctx context.
 			} else {
 				ruleTG.StackTargetGroupId = tg.ID()
 			}
+		} else {
+			// Unknown kind - mark as invalid so rule returns 500 per Gateway API spec
+			t.log.Infof(ctx, "Unknown backendRef kind %s on route %s, marking as invalid", string(*backendRef.Kind()), t.route.Name())
```
SinghVikram97, line 371: Can we remove this from the list now https://github.com/aws/aws-application-networking-k8s/blob/main/test/conformance/skip-tests.yaml#L93 ?

pkg/controllers/route_controller.go
```diff
@@ -283,6 +284,19 @@ func (r *routeReconciler) buildAndDeployModel(
 	return stack, err
 }
 
+func serviceStatusFromStack(stack core.Stack) *latticemodel.ServiceStatus {
+	var resServices []*latticemodel.Service
+	if err := stack.ListResources(&resServices); err != nil {
+		return nil
+	}
+	for _, resSvc := range resServices {
```
SinghVikram97, line 294: can we add comment here to remind us ->  at most one Service per stack or something

SinghVikram97 (approved):

pkg/gateway/model_build_rule.go
liucalvin: this test still fails because it uses the shared Gateway address instead of the per-route address, which can point to a different valid route on the gateway instead of the invalid route in this test, so the test fails because it's still resolving a different lattice service domain which returns 404. 

can investigate more if this can be supported as a follow up

https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/tests/httproute-invalid-nonexistent-backendref.go#L49

PR #973, opened 2026-06-29
fix(helm): expose service network override value

### LINKED ISSUE #625: Helm chart doesn't support ENABLE_SERVICE_NETWORK_OVERRIDE option
Documentation mention use of [ENABLE_SERVICE_NETWORK_OVERRIDE](https://www.gateway-api-controller.eks.aws.dev/guides/environment/#enable_service_network_override), however there is no way to enable option in Helm chart.

Helm chart does provide provide to set `DEFAULT_SERVICE_NETWORK` using `defaultServiceNetwork` though.

Saithej2k:
**What type of PR is this?**

bug

Fixes #625

The Helm deployment template already maps `ENABLE_SERVICE_NETWORK_OVERRIDE` from `.Values.enableServiceNetworkOverride`, but the value is not exposed in `values.yaml` or the chart schema.

This adds the chart value with a default of `false`, adds it to `values.schema.json`, and documents the Helm value name in the environment guide.

N/A

```bash
git diff --check
jq empty helm/values.schema.json
```

**Automation added to e2e**:

No. This is a Helm values/schema/docs change only.

No.

No breaking change expected. The new Helm value defaults to `false`, preserving the existing rendered environment value unless users opt in.

Updating a running cluster was not tested locally.

Yes. Users can now discover and set the service network override option through the Helm chart values.

```release-note
Expose the `enableServiceNetworkOverride` Helm value for configuring `ENABLE_SERVICE_NETWORK_OVERRIDE`.
```

```bash
Not run locally. This change only updates Helm chart values, schema, and documentation.
```

### DIFF
--- a/docs/guides/environment.md
+++ b/docs/guides/environment.md
@@ -75,6 +75,8 @@ The created service network will be also associated with cluster VPC.
 
 When set as "true", the controller will run in "single service network" mode that will override all gateways to point to default service network, instead of searching for service network with the same name. Can be used for small setups and conformance tests.
 
+The Helm chart exposes this setting as `enableServiceNetworkOverride`.
+
 ---
 
 #### `WEBHOOK_ENABLED`
--- a/helm/values.schema.json
+++ b/helm/values.schema.json
@@ -166,6 +166,9 @@
       "type": "string",
       "enum": ["cluster", "namespace"]
     },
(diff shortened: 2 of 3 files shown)


liucalvin (approved):
thanks for the contribution!

PR #983, opened 2026-07-30
fix certificate resolution for routes with omitted sectionName

### LINKED ISSUE #979: Mismatched behavior for listener matching and cert resolution when sectionName is omitted.
**Describe the bug**

When an HTTPRoute (or GRPCRoute/TLSRoute) attaches to a Gateway via a parentRef that omits sectionName, the controller correctly auto-matches the route to listeners by hostname intersection, but it ignores the TLS certificate configuration of the auto-matched listener. The resulting VPC Lattice Service is created with an empty CustomerCertARN.

The root cause is that listener attachment and certificate resolution use two different matching strategies:
  - Attachment (pkg/gateway/model_build_listener.go, buildListeners) uses core.IsRouteAllowedByListener, which matches by hostname intersection and treats sectionName as optional. So a sectionName-less parentRef fans out to every matching listener.
  - Certificate resolution (pkg/gateway/model_build_lattice_service.go, getACMCertArn) matches by exact sectionName equality and skips any parentRef where sectionName == nil

 Because the loop bails on sectionName == nil, two things go wrong for auto-matched routes:
  1. The explicit cert on the matched listener (section.TLS.Options[awsCustomCertARN]) is never read.
  2. The local hasTLSTerminateListener flag never flips to true. That flag also gates the automatic ACM certificate-discovery fallback (~lines 288–305), so auto-discovery is skipped too.

Result: getACMCertArn returns "", spec.CustomerCertARN is empty, and the Lattice Service ends up with no certificate — even though the listener the route attached to has a valid one.

**To Reproduce**
1. Create a Gateway with an HTTPS/TLS listener that has tls.mode: Terminate and a cert configured, e.g. tls.options."application-networking.k8s.aws/certificate-arn": <acm-arn> (and a hostname, e.g. *.example.com).
2. Create an HTTPRoute whose hostnames intersect that listener (e.g. foo.example.com), with a parentRef that points to the Gateway without a sectionName.
3. Apply the route. Observe status.parents[].conditions shows Accepted: True — attachment succeeded.
4. Inspect the created VPC Lattice Service (console or aws vpc-lattice get-service): certificateArn / custom-domain cert is empty.
5. Add an explicit sectionName: <listener-name> to the same parentRef and re-apply → the certificate now appears on the Lattice Service. This confirms the sectionName == nil skip is the
cause.

**Expected behavior**
Certificate resolution should use the same listener-matching logic as attachment. When a parentRef omits sectionName, the controller should resolve the certificate (and the hasTLSTerminateListener state that gates ACM auto-discovery) from the listener(s) the route actually attached to — not skip resolution entirely. A route that auto-matches a TLS-terminate listener should receive that listener's certificate without requiring a manually specified sectionName.

**Screenshots**

**Additional context**

**Possible Solution**
If you have any ideas on how to solve the issue, please describe them here.

liucalvin:
**What type of PR is this?**

bug

fixes #979

This PR fixes a bug where VPC Lattice Services are created without a TLS certificate when an HTTPRoute/GRPCRoute/TLSRoute attaches to a Gateway without specifying sectionName in its parentRef.
 
Certificate resolution now iterates all Gateway listeners and applies the same matching logic as listener attachment when sectionName is nil. If a matched listener has TLS mode Terminate, its cert annotation is read. If no annotation is present, the auto-discovery fallback is enabled — same behaviour as when sectionName is explicitly specified.

added unit tests and integ tests covering the exact issue

**Automation added to e2e**:
yes, new integ tests

no

no

```release-note
Fixed TLS certificates not being applied to VPC Lattice Services when a route's parentRef omits sectionName.
```

```
Ran 162 of 170 Specs in 3031.611 seconds
SUCCESS! -- 162 Passed | 0 Failed | 0 Pending | 8 Skipped
```

### DIFF
--- a/pkg/gateway/model_build_lattice_service.go
+++ b/pkg/gateway/model_build_lattice_service.go
@@ -265,21 +265,19 @@ func (t *latticeServiceModelBuildTask) getACMCertArn(ctx context.Context) (strin
 			continue
 		}
 
-		if parentRef.SectionName == nil {
-			continue
+		matched, err := t.matchedListeners(ctx, gw, parentRef)
+		if err != nil {
+			return "", err
 		}
 
-		for _, section := range gw.Spec.Listeners {
-			if section.Name == *parentRef.SectionName && section.TLS != nil {
-				if section.TLS.Mode != nil && *section.TLS.Mode == gwv1.TLSModeTerminate {
-					hasTLSTerminateListener = true
-					curCertARN, ok := section.TLS.Options[awsCustomCertARN]
-					if ok {
-						t.log.Debugf(ctx, "Found certification %s under section %s", curCertARN, section.Name)
-						return string(curCertARN), nil
-					}
+		for _, section := range matched {
+			if section.TLS != nil && section.TLS.Mode != nil && *section.TLS.Mode == gwv1.TLSModeTerminate {
+				hasTLSTerminateListener = true
+				curCertARN, ok := section.TLS.Options[awsCustomCertARN]
+				if ok {
+					t.log.Debugf(ctx, "Found certification %s under section %s", curCertARN, section.Name)
+					return string(curCertARN), nil
 				}
-				break
 			}
 		}
 	}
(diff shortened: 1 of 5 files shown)


Grabes (approved):
LGTM

